PR #22, opened 2015-05-25
Copy userinfo from base to path relative URL

tomalec:
as discussed in w3c/web-platform-tests#1852


annevk:
This patch is somewhat wrong unfortunately. We need to copy the username/password. Userinfo is just a syntax construct, not an actual part of a URL.

PR #32, opened 2015-07-01
Link "ASCII hex digits" in IPv4 number parser to its definiton.

SimonSapin:
(I think the `<style>` change is Bikeshed now doing syntax highlighting in python with Pygments rather than in JS with Prism.)


annevk:
Yeah, I needed to update bikeshed. Thanks!

PR #72, opened 2015-09-11
Define a host's "public suffix" and "registrable domain"

mikewest:
This patch attempts to formalize the way in which the Public Suffix List
relates to URLs, thereby making it possible to more easily explain the
limitations of features like `document.domain`, and cookies somewhat
absurd-seeming behavior across diverse hosts.

This patch does not expose an API to query this concept, as requested in
[1](https://www.w3.org/Bugs/Public/show_bug.cgi?id=25865). Doing so in a future patch should be quite straightforward,
however, if we decide that that's a reasonable thing to do.

***
[Preview](https://s3.amazonaws.com/pr-preview/mikewest/url/public.html) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/url/21171d3..mikewest:public:7a4feda.html)

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -255,6 +255,92 @@ point <a for=url>URLs</a> from <var>A</var> can come from untrusted sources.
 <a for=host>host</a> serves as a network address, but it is sometimes (ab)used as opaque
 identifier in <a for=url>URLs</a> where a network address is not necessary.
 
+<p>A <a>host</a>'s <dfn export for=host id=concept-host-public-suffix>public suffix</dfn> is
+the portion of a <a>host</a> which is controlled by a registrar, public or otherwise.
+To obtain a <var>host</var>'s <a>public suffix</a>, run these steps:
+
+1.  If <var>host</var> is an <a for=host>IPv4 address</a> or a <a for=host>IPv6 address</a>
+    return the empty string.
+
+2.  Let <var>host labels</var> be the result of <a lt="strictly split">strictly splitting</a>
+    <var>host</var> on the U+002E FULL STOP ('<code>.</code>') character.
+
+3.  Apply the <a>domain to ASCII</a> algorithm to each token in <var>host labels</var>.
+
+4.  Let <var>host</var> be the result of concatenating the items in <var>host labels</var>, in
+    the same order, separated by the U+002E FULL STOP ('<code>.</code>') character.
+
+5.  While <var>host</var> is not the empty string:
+
+    1.  If <var>host</var> matches the end of a suffix in the Public Suffix List,
+        return <var>host</var>. [PSL]
+
+    2.  <a>Strictly split</a> <var>host</var> on the U+002E FULL STOP ('<code>.</code>')
+        character, remove the first item from the resulting list, and let <var>host</var>
+        be the result of concatenating the remaining results, in the same order,
+        separated by the U+002E FULL STOP ('<code>.</code>') character.
+
+4.  Return the last item in <var>host labels</var>.
+
+<div class="example nobackref">
+  1. <code>example.com</code>'s public suffix is "<code>com</code>".
+  2. <code>example.co.uk</code>'s public suffix is "<code>co.uk</code>".
+  3. <code>example.appspot.com</code>'s public suffix is "<code>appspot.com</code>".
+  4. <code>subdomain.example.com</code>'s public suffix is "<code>com</code>".
+  5. <code>subdomain.example.co.uk</code>'s public suffix is "<code>co.uk</code>".
+  6. <code>subdomain.example.appspot.com</code>'s public suffix is "<code>appspot.com</code>".
+  7. <code>1.2.3.4</code>'s public suffix is "".
+  8. <code>2001:0db8:85a3:0000:0000:8a2e:0370:7334</code>'s public suffix is "".
+  
+  Assuming that <code>not-in-the-public-suffix-list</code> is not in the public suffix list:
+
+  1. <code>example.not-in-the-public-suffix-list</code>'s public suffix is "<code>not-in-the-public-suffix-list</code>".
+
+</div>
+
+<p>A <a>host</a>'s <dfn export for=host id=concept-host-registrable-domain>registrable
+domain</dfn> is the portion of a <a>host</a> which is assigned by a registrar, public
+or otherwise. To obtain a <var>host</var>'s <a>registrable domain</a>, run these steps:
+
+1.  If <var>host</var> is an <a for=host>IPv4 address</a> or a <a for=host>IPv6 address</a>
+    return <var>host</var>.
+
+2.  <a>Strictly split</a> <var>host</var> on the U+002E FULL STOP ('<code>.</code>')
+    character, apply the <a>domain to ASCII</a> algorithm to each returned token,
+    and let <var>host</var> be the result of concatenating the results, in the same
+    order, separated by the U+002E FULL STOP ('<code>.</code>') character.
+
+3.  Let <var>public suffix</var> be <var>host</var>'s <a>public suffix</a>.
+
+4.  If <var>public suffix</var> is <var>host</var>, return the empty string.
+
+5.  Let <var>registrable domain</var> be <var>public suffix</var>, <var>host labels</var>
+    and <var>public labels</var> be the result of <a lt="strictly split">strictly
+    splitting</a> <var>host</var> and <var>public suffix</var> respectively on the
+    U+002E FULL STOP ('<code>.</code>') character.
+
+6.  While the last item in <var>host labels</var> is an exact match with the last item
+    in <var>public labels</var>, pop the last item off both lists.
+
+7.  Prepend the last item in <var>host labels</var> to <var>registrable domain</var>.
+
+8.  Return <var>registrable domain</var>.
+
+<div class="example nobackref">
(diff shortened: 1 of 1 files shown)


mikewest:
Not sure how well this is explained, but it's a starting point.

mikewest:
@sleevi and other folks interested in publicsuffix/list#27 should have a look. Hi Ryan! You love the PSL SO MUCH, right? Let's get it into every spec.

url.bs
```diff
@@ -255,6 +255,92 @@ point <a for=url>URLs</a> from <var>A</var> can come from untrusted sources.
 <a for=host>host</a> serves as a network address, but it is sometimes (ab)used as opaque
 identifier in <a for=url>URLs</a> where a network address is not necessary.
 
+<p>A <a>host</a>'s <dfn export for=host id=concept-host-public-suffix>public suffix</dfn> is
+the portion of a <a>host</a> which is controlled by a registrar, public or otherwise.
+To obtain a <var>host</var>'s <a>public suffix</a>, run these steps:
+
+1.  If <var>host</var> is an <a for=host>IPv4 address</a> or a <a for=host>IPv6 address</a>
+    return the empty string.
```
sleevi: "If host is not a valid domain, return the empty string" seems to be more accurate (and reflects Chrome's implementation, at least)

If our GURL canonicalizer spits it out, then we spit out the empty string.

url.bs
```diff
@@ -255,6 +255,92 @@ point <a for=url>URLs</a> from <var>A</var> can come from untrusted sources.
...
+
+2.  Let <var>host labels</var> be the result of <a lt="strictly split">strictly splitting</a>
+    <var>host</var> on the U+002E FULL STOP ('<code>.</code>') character.
+
+3.  Apply the <a>domain to ASCII</a> algorithm to each token in <var>host labels</var>.
```
sleevi: This is not required.

url.bs
```diff
@@ -255,6 +255,92 @@ point <a for=url>URLs</a> from <var>A</var> can come from untrusted sources.
...
+
+4.  Let <var>host</var> be the result of concatenating the items in <var>host labels</var>, in
+    the same order, separated by the U+002E FULL STOP ('<code>.</code>') character.
+
+5.  While <var>host</var> is not the empty string:
+
+    1.  If <var>host</var> matches the end of a suffix in the Public Suffix List,
+        return <var>host</var>. [PSL]
```
sleevi: This doesn't seem correct.

https://publicsuffix.org/list/ is the canonical description of the algorithm, which explains the list format and a formal algorithm and terminology definition. See "Algorithm"

url.bs
```diff
@@ -255,6 +255,92 @@ point <a for=url>URLs</a> from <var>A</var> can come from untrusted sources.
...
+  1. <code>example.not-in-the-public-suffix-list</code>'s public suffix is "<code>not-in-the-public-suffix-list</code>".
+
+</div>
+
+<p>A <a>host</a>'s <dfn export for=host id=concept-host-registrable-domain>registrable
+domain</dfn> is the portion of a <a>host</a> which is assigned by a registrar, public
+or otherwise. To obtain a <var>host</var>'s <a>registrable domain</a>, run these steps:
+
+1.  If <var>host</var> is an <a for=host>IPv4 address</a> or a <a for=host>IPv6 address</a>
+    return <var>host</var>.
```
sleevi: Valid domain

url.bs
```diff
@@ -255,6 +255,92 @@ point <a for=url>URLs</a> from <var>A</var> can come from untrusted sources.
...
+
+2.  <a>Strictly split</a> <var>host</var> on the U+002E FULL STOP ('<code>.</code>')
+    character, apply the <a>domain to ASCII</a> algorithm to each returned token,
+    and let <var>host</var> be the result of concatenating the results, in the same
+    order, separated by the U+002E FULL STOP ('<code>.</code>') character.
```
sleevi: Not necessary

url.bs
```diff
@@ -255,6 +255,92 @@ point <a for=url>URLs</a> from <var>A</var> can come from untrusted sources.
...
+
+5.  Let <var>registrable domain</var> be <var>public suffix</var>, <var>host labels</var>
+    and <var>public labels</var> be the result of <a lt="strictly split">strictly
+    splitting</a> <var>host</var> and <var>public suffix</var> respectively on the
+    U+002E FULL STOP ('<code>.</code>') character.
+
+6.  While the last item in <var>host labels</var> is an exact match with the last item
+    in <var>public labels</var>, pop the last item off both lists.
+
+7.  Prepend the last item in <var>host labels</var> to <var>registrable domain</var>.
```
sleevi: ?

sleevi:
I'm not terribly keen for this, as explained before :)

Does the "Formal Algorithm" for the PSL, as described at https://publicsuffix.org/list/ , not meet the needs? Why not?

@annevk 's remarks on https://github.com/publicsuffix/list/issues/27 suggest the reason it wouldn't is that implementations of the PSL Algorithm, with respect to parsing rules, are able to use whatever internal storage format they want (either as punycode - as Chrome does - or as full unicode).

The above algorithm doesn't treat how rules delegated as ICANN or not are handled (see "Divisions")

url.bs
annevk: It should be "is not a domain" then. "Valid domain" is a special term we tried to create to restrict the syntax of a domain further for the purposes of conformance. But it's not entirely baked yet or something that would be implemented by user agents.

url.bs
annevk: Why not? Although the splitting isn't necessary, that part is presumably copied from HTML, but HTML is wrong. Better to copy from the URL Standard itself.

annevk:
@sleevi they cannot just use whatever internal storage format, they can also use whatever conversion algorithm they want. Both seem problematic.

sleevi:
@annevk If that's all it would take to just reference the "Formal Algorithm" of the PSL, why not simply incorporate that as a profile here?

That is, something like "Let X be the result of evaluating the formal algorithm of the PSL as if both the rules and the hostname were canonicalized as (IDNA2003, IDNA2008, UTF-8, w/e)".

That is, you're not normatively requiring an implementation that exactly follows the spec's prosaic description, only normatively requiring something that yields the same inputs and outputs, with an added requirement on what the externally observable affects are regarding conversions.

For example, if my storage was UTF-8, but the spec described it "as if IDNA2008", then prior to sending it to my PSL implementation (which is UTF-8), I'd validate whatever strings were convertible to/from IDNA2008. Doesn't that resolve it?

annevk:
I think that might work, though it's not exactly pretty. That is, if that is the intended setup ideally those bits would be pluggable from the calling site of the algorithm.

sleevi:
@annevk We can certainly fix the PSL side to make it, if that would help.

My main concern is that we really want there to be "one" PSL algorithm, and since the PSL lives in many more cases than just browsers, and because the algorithm is tightly coupled to the data format (with respect to rules and sections and such), it seems that algorithm should live at the PSL.

We should make it possible for those to plug in the necessary bits they need to ensure consistency between implementations of their use case (e.g. for browsers), but I think we would naturally want the rule processing algorithm to live with the set of rules :)

annevk:
Okay, that sounds like a decent compromise. (Ideally of course there be no difference on this between user agents implementing the URL Standard and other user agents just dealing with domains, but if it has to be that way, this is as close as we can get.)

sleevi:
@annevk I agree that would be ideal, but I think it also matters the contextual nature of the protocol. For example, if you're processing an e-mail address, you're certainly working on localname@hostname, but what 'hostname' means is not necessarily the same as the URL standard (for reasons such as encoding/escaping for the framing SMTP or for the type of name resolution). That's why I was concerned with putting a normative requirement on the PSL to expect external inputs in a particular form (or to be convertible to a particular internal storage form). Ideally, that would be left to the implementer to decide.

But we can certainly add hooks to the PSL to ensure that however a program implements the PSL, it needs to ensure the input form and storage form match, to add a step to validate that condition, and add a parameter to indicate that whatever that particular form is, that form is at least specified in the 'overall' specification (URL or otherwise)

annevk:
@sleevi how would one go about patching/PR'ing the PSL algorithm to take these hooks? I would love to get this defined better.

sleevi:
One does not simply patch the PSL algorithm - it documents the state of what is shipping in hundreds of unique implementations. The normal spec challenges - the world we have vs what we want

annevk:
@sleevi earlier you said we could add hooks.

sleevi:
Right, and in that time we've been working through issues with implementations, in particular, wild cards, that have exposed the PSL much more fragile than anticipated. Re-reading these hooks (... It has been five months), we can add documentation to the effect of the spec by submitting an Issue to https://github.com/publicsuffix/list

annevk:
I don't think much changed to that algorithm over the past five months. https://github.com/publicsuffix/list/issues/27 is still open, I guess I can open a new issue and close that one if you think that helps?

I still think the Formal Algorithm is hugely confusing by stating "The domain and all rules must be canonicalized in the normal way for hostnames - lower-case, Punycode (RFC 3492)." which is not at all the normal way. E.g., that does not acknowledge browsers recognize (and normalize away) four types of dots.

annevk:
We'll leave this to HTML for now and require everyone to reuse the algorithm defined there: https://github.com/whatwg/html/pull/2365.

PR #78, opened 2016-01-19
Add a note in the URL parser.

SimonSapin:
When reading this part of the parser I first thought that it was possible to enter the relative state with a non-relative base URL, which would lead to instructions that don’t make much sense. I’m now convinced that is not possible, but the reason is subtle and worth noting.

I did not include regenerating `url.html` as the diff would be noisy, I seem to have a different version of Bikeshed.


annevk:
Thank you, ended up fixing this in a slightly different way.

PR #82, opened 2016-02-05
move myself to being a contributor

rubys:
$ git push
ERROR: Permission to whatwg/url.git denied to rubys.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -2888,6 +2888,7 @@ Rodney Rehm,
 Roy Fielding,
 Ryan Sleevi,
 Santiago M. Mola,
+Sam Ruby,
 Sebastian Mayr,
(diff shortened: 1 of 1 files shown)


annevk:
If you still want to contribute I can reinstate you. This was a side effect of cleaning up teams. However, you haven't committed anything in a while so maybe it's for the best.

rubys:
In the event that I resume contributing, we can re-evaluate it at that time.

My primary interest is in interop. Should there be a group of people interested in comparative test results and working to reduce differences, I would be interested.

annevk:
I had to change your patch slightly since you put your name in the wrong spot.

PR #92, opened 2016-02-15
Have new URLSearchParams() remove a leading "?"

### LINKED ISSUE #81: Let's strip leading '?' if present in `URLSearchParams` init string param
(not sure if this is the best place to put this)

Regarding `URLSearchParams` constructor string param:
IMO it would be good to skip leading `?` if present, before starting to do the real parsing.

The reason is that `window.location.search` returns `?a=1&b=2` with leading `?`.
For consistency the URLSearchParams constructor should allow the leading `?` thus.

Looking at the spec and test they don't explicitly mention anything about leading `?`

https://url.spec.whatwg.org/#interface-urlsearchparams
https://github.com/w3c/web-platform-tests/blob/0ae07c6dec055b416e91a8e0e431eec8c2b2552b/url/urlsearchparams-get.html (master as of today)

In the current Firefox stable 44 and Chrome Canary 50 implementations this happens:

`new URLSearchParams('a=1&b=2').get('a') == 1`
`new URLSearchParams('?a=1&b=2').get('a') == null`
`new URLSearchParams('?a=1&b=2').get('?a') == 1`

what if instead we skipped leading `?` if present and return this:

`new URLSearchParams('?a=1&b=2').get('a') == 1`
`new URLSearchParams('?a=1&b=2').get('?a') == null`

What do you think?

annevk:
Fixes #81.

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -2068,11 +2068,11 @@ then runs these steps:
 <h3 id=origin>Origin</h3>
 <!-- Still need to watch the final bits -->
 
-<p class=note>See <a spec=html for=/>origin</a>'s definition in HTML for the necessary
+<p class=note>See <a spec=html for=origin>origin</a>'s definition in HTML for the necessary
 background information. [[!HTML]]
 
 <p>A <a for=url>URL</a>'s <dfn export for=url id=concept-url-origin>origin</dfn> is the
-<a spec=html for=/>origin</a> returned by running these steps, switching on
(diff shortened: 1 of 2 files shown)


annevk:
This could not be landed because the reference system from bikeshed suddenly decided that "concept-request-origin" is a thing HTML defines, despite it being annotated to the contrary. Paging @tabatkins for assistance.

PR #105, opened 2016-03-23
Rename non-relative to cannot-be-a-base-URL.

### LINKED ISSUE #89: Rename non-relative-flag
* not-a-base-url flag
* cannot-be-a-base-url flag

Better suggestions welcome.

When we change this at least HTML needs to be updated too.

SimonSapin:
Fix #89.

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -844,8 +844,9 @@ resource the <a for=url>URL</a>'s other components identify. It is initially nul
 
 <p class="note no-backref">This is not an <a>ASCII string</a> on purpose.
 
-<p>A <a for=url>URL</a> also has an associated <dfn>non-relative flag</dfn>. It is
-initially unset.
+<p>A <a for=url>URL</a> also has an associated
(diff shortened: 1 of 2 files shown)


url.bs
```diff
@@ -844,8 +844,9 @@ resource the <a for=url>URL</a>'s other components identify. It is initially nul
 
 <p class="note no-backref">This is not an <a>ASCII string</a> on purpose.
 
-<p>A <a for=url>URL</a> also has an associated <dfn>non-relative flag</dfn>. It is
-initially unset.
+<p>A <a for=url>URL</a> also has an associated
+<dfn><span id=non-relative-flag>cannot-be-a-base-URL flag</span></dfn>.
```
SimonSapin: The `<span>` is to keep external links working. Bikshed has an `oldids` attribute for this, but (for now) it only applies to headings.

annevk:
Thanks, landed as https://github.com/whatwg/url/commit/2f4161d9f82b1752d9627ee7f34886e94464b054.

PR #113, opened 2016-04-13
`url.port = ""` sets the port number to null.

SimonSapin:
It was previously (per spec) a no-op.

This matches Firefox. (Chromium sets the port to zero.)
http://software.hixie.ch/utilities/js/live-dom-viewer/saved/4054

(`url.html` not regenerated since the latest version of Bikeshed does various things differently, generating unrelated diff churn.)

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -1632,9 +1632,6 @@ optionally with an <a>encoding</a>
          <li><p>Set <var>buffer</var> to the empty string.
         </ol>
 
-       <li><p>If <var>state override</var> is given and <var>buffer</var> is the empty string,
-       Set <var>url</var>'s <a for=url>port</a> to null.
-
        <li><p>If <var>state override</var> is given, terminate this algorithm.
 
        <li><p>Set <var>state</var> to <a>path start state</a>, and decrease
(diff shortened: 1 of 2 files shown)


SimonSapin:
This also affects `url.host = "example.net:"`

annevk:
The example.net does not seem to match Firefox (although Chrome does set it to zero there).

annevk:
We could also define this as part of https://url.spec.whatwg.org/#dom-url-port or check that state override is "port state"...

annevk:
Probably defining it in the port setter would be clearer.

annevk:
@achristensen07 @cdumez what do you think about changing the behavior of `url.port = ""`?

Presumably we should also change HTML if we change this here. Did you check `location`?

achristensen07:
I agree with this change.  If someone wants port 0, they would set url.port to 0, not the empty string. I was planning on changing the behavior of port 0 vs. no port anyways.

annevk:
It seems Firefox and Safari use the default port (not 0) for `location`. Safari does still use 0 elsewhere, but it seems Alex is willing to change that, yay. I guess I'll change this PR to be specifically for the port setter and create a corresponding patch for HTML.

annevk:
Thanks for the review @zcorpan and thanks @SimonSapin for the original patch (although do note I went with a different fix).

PR #114, opened 2016-04-13
Stop after setting an url’s query to null.

SimonSapin:
Don’t go on to setting it to the empty string in a following step, which makes setting to null pointless

This is probably what was intended, as it is similar to what happens in the `hash` setter.


annevk:
Thank you, landed as https://github.com/whatwg/url/commit/4f1c2ddbdb866b1150819622ec04a86813294059.

PR #115, opened 2016-04-22
Improve the comment about blob URL origins.

Ms2ger:


annevk:
Thank you, landed as https://github.com/whatwg/url/commit/787a4e5fb4b1798ea810252c8c2c5eefd589280b.

PR #123, opened 2016-05-25
Clarify the informative nature of referenced RFCs

### LINKED ISSUE #122: Section 3 (Hosts) should be non-normative
We noticed today that this section:

https://url.spec.whatwg.org/#hosts-%28domains-and-ip-addresses%29

Contradicts the URL spec parsing a bit.  The RFC for domains does not allow characters like `&`, but the URL spec parsing does allow it.

So section 3 should probably be marked non-normative.

annevk:
Fixes #122.

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -254,6 +254,10 @@ point <a for=url lt=URL>URLs</a> from <var>A</var> can come from untrusted sourc
 <a for=host>host</a> serves as a network address, but it is sometimes (ab)used as opaque
 identifier in <a for=url lt=URL>URLs</a> where a network address is not necessary.
 
+<p class=note>The RFCs referenced in the paragraphs below are for informative purposes only. They
+have no influence on <a for=host>host</a> syntax, parsing, and serialization. Unless stated
+otherwise in the sections that follow.
+
 <p>A <dfn export for=host id=concept-domain>domain</dfn> identifies a realm within a
 network.
 [[RFC1034]]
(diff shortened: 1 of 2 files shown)


annevk:
@wanderview, does this look okay?

domenic:
It seems to me there are still two contradictory definitions of host. One that says "go look at the RFC", and one that is a component of URL records.

annevk:
@domenic I don't understand? A URL's host is either null or a host. A host is a domain, IPv4 address, or IPv6 address. I don't think it's defined twice, where are you seeing that?

domenic:
The concepts are defined twice: in 3 and in 3.2.

annevk:
3.2 doesn't define a concept. It defines syntax.

domenic:
Well, both define "host", and the URL Standard doesn't seem to use the syntax defined in the RFC referenced in 3. I think if you want to make this clear you should change 3.2 to "host syntax" and should point out that the RFC defines the concept but contains syntax that must be ignored in favor of the "host syntax" definition.

annevk:
They define a different "host". Just like URL record's host is different. I mean this note says exactly that. Ignore the referenced RFCs for the purposes of syntax, parsing, and serialization...

domenic:
My point is that having two dfns with the same text is extremely confusing especially when it's easy to read both as defining syntax.

annevk:
That is a more general problem. I guess we could introduce the disambiguation terms "host record" and "host string" here too, and similar for host's sub parts, but that's a bigger issue than this.

domenic:
Well, all I can say is this PR doesn't make things clearer for me (see also #118) but maybe it helps @wanderview.

annevk:
I think that is worth explaining separately from the informative nature of the RFCs.

annevk:
I still don't have a plan for clarifying out the various definitions of host/query/port/etc. but this seemed like a good clarification either way.

PR #146, opened 2016-08-23
Update origin links now that HTML defines it with for=''.

jyasskin:
I've filed https://github.com/w3c/csswg-drafts/issues/422 about a conflicting definition of "origin".

However, we shouldn't need even as much markup as I'm using (`<a spec=html for=/>origin</a>`), to link to HTML's origin. @tabatkins, adding a `link-defaults` of `spec:html; type:dfn; text:origin; for:/` should cause `<a>origin</a>` to link to HTML, right? It doesn't.

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -2093,11 +2093,11 @@ then runs these steps:
 <h3 id=origin>Origin</h3>
 <!-- Still need to watch the final bits -->
 
-<p class=note>See <a spec=html for=origin>origin</a>'s definition in HTML for the necessary
+<p class=note>See <a for=/>origin</a>'s definition in HTML for the necessary
 background information. [[!HTML]]
 
 <p>A <a for=url>URL</a>'s <dfn export for=url id=concept-url-origin>origin</dfn> is the
-<a spec=html for=origin>origin</a> returned by running these steps, switching on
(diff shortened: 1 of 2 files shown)


annevk:
I suspect we can leave out spec=html now?

jyasskin:
Sadly, not until [css-cascade-3] is fixed to avoid defining its own "origin" with no `for`.

tabatkins:
Hmm, I wonder what "for" value we would use for it?  (In general, these kinds of total clashes are unfortunate but expected in rare cases; it's one of the reasons `spec` is part of the autolinking discriminator. But "origin" is common enough that it's worth making this easy, and I think that HTML's origin is much more popular than CSS's.)

Maybe `for=cascade`?

annevk:
for="style rule" (or CSSStyleRule, depending on how you plan to define CSS' model) would make sense when reading the definition. Just like CSS probably has for=viewport for origin too.

jyasskin:
[css-cascade-3] is fixed, so I've removed the `spec=html` here.

annevk:
Thank you, landed as https://github.com/whatwg/url/commit/995362a9eab93d1b3e2aebfbaf992699c9a995f8.

PR #153, opened 2016-10-19
URL's href attribute needs to reset URLSearchParams too

### LINKED ISSUE #117: The href setter needs to reset the query object
The `href` setter of `URL` objects is currently:

> Let parsedURL be the result of running the basic URL parser on the given value.
> If parsedURL is failure, throw a TypeError exception.
> Set context object’s url to parsedURL.

This looks like it’s missing additional steps to propagate the change to `searchParams`, as tested in https://github.com/w3c/web-platform-tests/blob/0a518aaff/url/url-constructor.html#L121-L123. Something like:

> Let query be parsedURL’s query, if that is non-null, and the empty string otherwise.
> Set context object’s query object to a new URLSearchParams object using query, and then set that query object’s url object to context object.

(Similar to what the `URL` constructor does.)

annevk:
Fixes #117.


annevk:
@SimonSapin @nox interested in reviewing?

nox:
SGTM.

PR #154, opened 2016-10-19
Editorial: add a note and example for query encoding

### LINKED ISSUE #95: "A query must be zero or more URL units.  "
https://url.spec.whatwg.org/#url-syntax

> A query must be zero or more URL units\.
> 

This should probably say something about encodings. Maybe make unescaped non-ASCII characters invalid in the query when the encoding isn't utf-8 (or utf-16?). And add a note/example explaining this.

Related https://github.com/htacg/tidy-html5/issues/378

zcorpan:
Fixes #95.

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -1101,15 +1101,14 @@ U+F0000 to U+FFFFD,
 U+100000 to U+10FFFD.
 
 <p class=note>Code points higher than U+007F will be converted to
-<a lt="percent-encoded byte">percent-encoded bytes</a> by the
-<a>URL parser</a>, except for code points appearing in
-<a for=url lt=fragment>fragments</a>.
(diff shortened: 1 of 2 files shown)


url.bs
```diff
@@ -1105,6 +1105,28 @@ U+100000 to U+10FFFD.
 <a>URL parser</a>, except for code points appearing in
 <a for=url lt=fragment>fragments</a>.
 
+<p class=note>In HTML, when the document encoding is a legacy encoding, code points in the
+<a for=urlsyntax>query</a> that are higher than U+009F will be converted to
```
annevk: 7F?

zcorpan: Copied from the note above, but I suppose it should be 7F...

PR #156, opened 2016-10-28
Stop decoding all %2e's in paths

### LINKED ISSUE #87: Percent-decode more stuff?
In https://github.com/whatwg/url/issues/77, I was surprised that `.` was such a special case. It is indeed in Firefox 44, but in Chromium 48 other characters are also percent-decoded: `-`, `_`, `~`, and ASCII alpha-numerics. (I don’t have other browsers to test right now.)

http://software.hixie.ch/utilities/js/live-dom-viewer/?saved=3878

It may be worth testing other URL components too.

annevk:
This basically reverts bee5ad8041adfe6fc676c527bbc3f3cf4562ef67 as it
did not end up shipping in Firefox and the tentative conclusion is that
percent-decoding in URLs while parsing creates more trouble than it’s
worth.

We’ll still handle %2e when it’s part of a single-dot or double-dot
path segment.

Fixes #87.

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -1900,12 +1900,8 @@ string <var>input</var>, optionally with a <a>base URL</a> <var>base</var>, opti
        <li><p>If <a>c</a> is "<code>%</code>" and <a>remaining</a> does
        not start with two <a>ASCII hex digits</a>, <a>syntax violation</a>.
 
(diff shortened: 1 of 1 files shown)


annevk:
Tests: https://github.com/w3c/web-platform-tests/pull/4104.

PR #165, opened 2016-11-18
Editorial: start using Infra Standard

annevk:


annevk:
Currently DOM, HTML, and Infra all define "ASCII case-insensitive" so this cannot land just yet.

HTML should remove it first, then DOM can drop it, then we can land this.

PR #166, opened 2016-11-26
fix typo in IPv4 number parser algorithm

cameronsteele:
I hope that I am correct in my understanding, here. As ASCII hex digits
are the letters A to F and are used to represent digits with values 10
to 15, it will confused people newer to these concepts if they're asked
to also replace the values 0-9 with ASCII hex digits; even though from a
philosophical standpoint you could argue all of the numbers 0 to 15 are
technically in such a format.


cameronsteele:
also, apologies for adding my name to acknowledgements for such an innane thing-- just wanted to follow README instructions 100% but in retrospect it doesn't really belong there...

cameronsteele:
actually, my understanding was incorrect.

PR #169, opened 2016-12-08
Percent encode fragments too

### LINKED ISSUE #150: unicode fragments
Right now there is a note: "Unfortunately not using percent-encoding is intentional as implementations with majority market share exhibit this behavior."

I think we should change "Append c to url’s fragment." to at least "If url’s scheme is a special scheme, append c to url’s fragment.  Otherwise, append the result of running UTF-8 percent encode with the simple encode set on c" or something similar.  

Right now Firefox and Safari percent-encode non-ASCII characters in the fragment.  Chrome percent-encodes non-ASCII characters in the fragment if the scheme is not a special scheme.  Edge does not percent-encode any non-ASCII characters in the fragment.

This change is needed to continue to support percent-encoding non-ascii characters after a '#' in a data URL, which works in Safari, Chrome, and Firefox.

annevk:
Fixes #150.

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -824,10 +824,8 @@ hierarchical form. It is initially the empty list.
 null or an <a>ASCII string</a> holding data. It is initially null.
 
 <p>A <a for=/>URL</a>'s <dfn export for=url id=concept-url-fragment>fragment</dfn> is
-either null or a string holding data that can be used for further processing on the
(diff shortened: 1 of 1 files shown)


annevk:
https://github.com/w3c/web-platform-tests/issues/4298 has test changes. Make sure to mention those when committing.

PR #174, opened 2016-12-19
Editorial: clarify the "shorten a path" algorithm

domenic:
Sometimes "single" is used to mean "any" which causes a different, incorrect interpretation of this algorithm (see https://github.com/jsdom/whatwg-url/pull/60). Clarify by using a more precise algorithm.


annevk:
This looks good, although the usage of closing tags is inconsistent. As is the lack of newlines between `<li>`s. The other minor problem is that the rest of the document still needs to be upgraded to make more use of the Infra Standard, but I don't think that needs to block.

annevk:
LGTM. You're okay with my added nit commit?

domenic:
Yep, looks good

PR #175, opened 2016-12-19
Simplify constructing URLSearchParams

### LINKED ISSUE #27: Creation of URLSearchParams from Object/Map
Currently `URLSearchParams` cannot be directly created from an `Object` holding keys and values.

Combination of `Object.keys`, `Array.reduce` and `append` feels like boilderplate given that creation of query strings from objects is a common pattern in popular JS libraries, e.g. jQuery, superagent, node's url.format.

I suggest accepting an `Object`, `Map` (or anything that is iterable and has string keys) in the constructor and the `set` method.

i.e.

``` js
URLSearchParams({foo:"bar", baz:"quz"});
```

or

``` js
const tmp = new URLSearchParams();
tmp.set({foo:"bar", baz:"quz"});
```

could be equivalent to:

``` js
const tmp = new URLSearchParams();
tmp.append("baz", "quz");
tmp.append("foo", "bar");
```

I suggest always sorting the keys, since Object and Map don't guarantee any particular order, but a consistent order is very desirable for improved cacheability of resources.

annevk:
Fixes #27.

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -2704,7 +2704,7 @@ steps:
 <h3 id=interface-urlsearchparams>Interface {{URLSearchParams}}</h3>
 
 <pre class=idl>
-[Constructor(optional (USVString or URLSearchParams) init = ""),
+[Constructor(optional (sequence&lt;sequence&lt;USVString>> or record&lt;USVString, USVString> or USVString) init = ""),
  Exposed=(Window,Worker)]
 interface URLSearchParams {
   void append(USVString name, USVString value);
@@ -2717,6 +2717,14 @@ interface URLSearchParams {
   stringifier;
 };
 </pre>
+
+<div class=example id=example-constructing-urlsearchparams>
+ <p>Constructing and stringifying a {{URLSearchParams}} object is fairly straightforward:
+
+ <pre><code class="lang-javascript">
+let params = new URLSearchParams({key: 730d67})
+params.toString() // key=730d67</code></pre>
+</div>
 
 <p>A {{URLSearchParams}} object has an associated
 <dfn export for=URLSearchParams id=concept-urlsearchparams-list>list</dfn> of name-value pairs,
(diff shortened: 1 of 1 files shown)


jasnell:
This would certainly be interesting and I would implement this in the Node.js impl if adopted in the spec. I'd like to clarify: the `init` param could be either a string, Iterable, or object. If Iterable, the items must be pairs of strings (USVString to be exact), and for objects, only own properties whose keys/values are strings would be considered. Is that correct?

annevk:
Yeah. Though the USVString-ness happens through coercion, so if you have numbers there that would technically work through the magic of IDL. What's blocking this change is tests and asking browsers if they'd ship this.

jasnell:
Understood. I saw the "needs implementer interest" label. Wanted to express that we would impl and ship in Node.js if added and I think it's a worthwhile change so +1 from this side. I've opened an issue to track the progress on this and will keep an eye on it.

domenic:
The behavior here is governed by [the definiton for how to convert a JS value into a union type](https://heycam.github.io/webidl/#es-union). I think it ends up looking basically like this:

```js
if (arg === null || arg === undefined) {
  // record<USVString, USVString>
  
  // add zero key/value pairs to the URLSearchParams
} else if (typeof arg === "object") {
  const iterator = arg[Symbol.iterator];
  if (iterator === undefined || iterator === null) {
    // record<USVString, USVString>
    const map = new Map();
    for (const key of Object.getOwnPropertyNames(arg)) {
      const desc = Object.getOwnPropertyDescriptor(arg, key);
      if (desc !== undefined && desc.enumerable) {
        const typedKey = toUSVString(key);
        const typedValue = toUSVString(arg[key]);
        map.set(typedKey, typedValue);
      }
    }
    
    // Use `map` to add key/value pairs to the URLSearchParams
  } else {
    // sequence<sequence<USVString>>
    const pairs = [];
    
    // Note: per spec we have to first exhaust the lists (convert
    // from JS value to sequence<sequence<USVString>>) then process them
    for (const x of iterator) {
      const gotten = [];
      pairs.push(gotten);
      for (const y of x) {
        gotten.push(toUSVString(y));
      }
    }
    
    for (const pair of pairs) {
      if (pair.length !== 2) {
        throw new TypeError();
      }
      
      // Add (`pair[0]`, `pair[1]`) key/value pair to URLSearchParams
    }
  }
} else {
  // USVString
  const string = toUSVString(arg);
  // parse and process the string to add key/value pairs to the URLSearchParams
}
```

annevk:
(To be clear, it's certainly very cool that Node.js ships a compatible URL parser and API these days and your support hopefully makes for a more compelling case across the board. I didn't mean to sound dismissive.)

annevk:
Mozilla supports this change. @mikewest, I take it Chrome would implement this too in due course?

annevk:
Once the tests are reviewed and landed I'll file browser issues and land this change.

mikewest:
Yeah. This seems totally reasonable to me. Added some flags to the bug you filed against Chromium. Thanks!

PR #177, opened 2016-12-19
Editorial: add some URL parsing examples

### LINKED ISSUE #119: Make it clearer that just about anything is parseable, probably via some examples
Not everyone realizes that pretty much everything that contains a `:` is parseable as a URL (and is in fact parsed as such by browsers). People don't realize this just looking at the parser algorithm, and people coming from the RFCs have very different expectations.

A series of examples illustrating how absurd inputs get parsed as URLs would be great. Maybe draw a few from https://github.com/w3c/web-platform-tests/blob/master/url/urltestdata.json and then link to that file for more details.

Also, I believe there are some examples where you can contain a `:` but parsing will still fail. I'd like to have those included as well, as I always fail to remember them (when trying to qualify my "anything with a `:`" statements).

Potentially related to #118.

annevk:
Fixes #119.

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -1130,6 +1130,37 @@ different document encoding. Using the <a>UTF-8</a> encoding everywhere solves t
  <li><p>Return <var>url</var>.
 </ol>
 
+<div class=example id=example-url-parsing>
+ <table>
+  <tr>
+   <th>Input
+   <th>Base
+   <th>Output
+  <tr>
+   <td><code>example</code>
+   <td><code>https://example.com/demo</code>
+   <td><code>https://example.com/example</code>
+  <tr>
+   <td><code>hello:world</code>
+   <td><code>https://example.com</code>
+   <td><code>hello:world</code>
+  <tr>
+   <td><code>https://example.com:demo</code>
+   <td>
+   <td>Failure
+  <tr>
+   <td><code>https:example.invalid</code>
+   <td><code>https://example.com</code>
+   <td><code>https://example.com/example.invalid</code>
+  <tr>
+   <td><code>\example\..\demo/.\</code>
+   <td><code>https://example.com</code>
+   <td><code>https://example.com/demo/</code>
+ </table>
+
+ <p>The base and output URL are represented in serialized form for brevity.</p>
+</div>
+
 <hr>
 
 <p>The <dfn export id=concept-basic-url-parser lt='basic URL parser'>basic URL parser</dfn> takes a


annevk:
@domenic feel free to add more examples in new commits.

url.bs
```diff
@@ -1130,6 +1130,37 @@ different document encoding. Using the <a>UTF-8</a> encoding everywhere solves t
...
+  <tr>
+   <td><code>hello:world</code>
+   <td><code>https://example.com</code>
+   <td><code>hello:world</code>
+  <tr>
+   <td><code>https://example.com:demo</code>
+   <td>
+   <td>Failure
+  <tr>
+   <td><code>https:example.invalid</code>
```
domenic: Reminder to self: we should add this example but against about:blank since I am pretty sure (will check) that it results in `https://example.invalid/`.

annevk:
I wonder if we should move this elsewhere and add an additional column to this table with respect to validity. Or perhaps `<mark>` the invalid portions of input.

Then we could also address #118.

Maybe after "shorten a url’s path"?

domenic:
OK, added a number of my favorite examples. There were certainly more (e.g. demonstrating some of the host normalization like IDNA and IPv4/v6 parsing would have been cool; demonstrating file URLs would probably help) but this I think hammers home the main points about how a large variety of weird URLs cam parse, but some things can still cause failure.

Regarding validity. I think the current location is pretty good. Maybe what we need is a paragraph spelling out the fact that URL syntax is all about validity, and what validity means in practice. That paragraph could link to this example, perhaps.

Adding a column for validity seems OK but we might need some more valid examples in that case, as I emphasized the invalid ones. Also I'm not sure it applies that well since we talk about validity on an absolute URL, not on an input to the URL parser, right? So e.g. the row that lists `example` + `https://example.com/demo` would not work out great, would it?

url.bs
```diff
@@ -1130,6 +1130,58 @@ different document encoding. Using the <a>UTF-8</a> encoding everywhere solves t
...
+   <td><code>https://example.com</code>
+   <td><code>https://example.com/example.org</code>
+  <tr>
+   <td><code>\example\..\demo/.\</code>
+   <td><code>https://example.com</code>
+   <td><code>https://example.com/demo/</code>
+  <tr>
+   <td><code>example</code>
+   <td><code>https://example.com/demo</code>
+   <td><code>https://example.com/example</code>
```
annevk: This one was upfront because it demonstrates a rather simple thing that everyone takes for granted.

annevk: Hmm okay.

annevk:
Validity would be about the input column. Though it could be a column on its own describing the issue.

domenic:
I think it would be kind of confusing to say that `example` is invalid input in the `example` + `https://example.com/demo` demo. (That is, `example` by itself does not correspond to the URL syntax definitions.) So maybe a separate table of valid and invalid URLs would be good.

annevk:
`example` by itself matches https://url.spec.whatwg.org/#syntax-url-path-relative which certainly is a URL string.

domenic:
Oh, I didn't realize; very interesting. I think that's not usually what people mean by "valid URL" but it does have a rigorous definition, so good enough for me.

annevk:
There is some handwaving around base URL which at some point should be resolved. Since to be a valid relative-URL string, we do need a base URL.

annevk:
I decided to go ahead and merge this since it's an improvement over the status quo and the validity question deserves a bit more thought and study.

PR #178, opened 2016-12-19
Make application/x-www-form-urlencoded encode filenames

### LINKED ISSUE #109: File names in application/x-www-form-urlencoded serializer
b174f8f2170a4ce6fd4031206e0490eff2b0aa02 added

> Otherwise, if tuple has a type, and tuple’s type is "file", set outputPair’s value to tuple’s value’s filename.

Contrary to the general case, a filename here does not go through the "byte serializer" (or character encoding). This means that any `&` or `=` character in the name will be serialized as-is, incorrectly creating new name/value pairs when the serialized string is parsed again. Is this really intended?

It’s also not clear from https://html.spec.whatwg.org/#concept-input-type-file-selected what the type of a file name is supposed to be in the first place. (Bytes vs Unicode vs WTF-16/DOMString.)

annevk:
We’ll leave converting an OS filename to a JavaScript string to someone
else for now.

Fixes #109.

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -2311,28 +2311,31 @@ takes a list of name-value or name-value-type tuples <var>tuples</var>, optional
   <p>For each <var>tuple</var> in <var>tuples</var>, run these substeps:
 
   <ol>
-   <li><p>Let <var>outputPair</var> be a new name-value pair.
+   <li><p>Let <var>name</var> be the result of <a lt="urlencoded byte serializer">serializing</a>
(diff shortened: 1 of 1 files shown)


annevk:
I don't think this can be tested automatically. I tested this in Live DOM Viewer using
```html
<form action=http://software.hixie.ch/utilities/cgi/test-tools/echo accept-charset=windows-1252><input type=file name=x><input type=submit></form>```
as markup and a locally created `™.txt` resource as input.

Issue #173, opened 2016-12-18
Editorial: extra "and" in IPv4 parser


domenic:
https://url.spec.whatwg.org/commit-snapshots/373dbedbbf0596f723ce8a195923da98b698aeb0/#host-parsing

> If input contains a code point that is not a radix\-R digit, and return failure\. 

The "and" is extra here.

PR #180, opened 2016-12-20
Editorial: removing extra "and" in IPv4 parser

italoacasas:
Fixes: https://github.com/whatwg/url/issues/173


annevk:
Hi @italoacasas, thanks! Could you make it "then return failure"? Also, would you like to add yourself to the acknowledgments section?

annevk:
Thanks again! In the future please add a comment after you've made a change since GitHub notification system doesn't pick up on changes solely to the commit.

PR #185, opened 2016-12-28
Add opaque hosts

### LINKED ISSUE #148: Change host parser for non-special schemes
`new URL("unknown://†/")` does not result in "Punycode" (results in `%E2%80%A0`) and `new URL("notspecial://H%4fSt/path")` does not remove the URL encoding (results in `H%4fSt` not `host`). It seems all implementations are in agreement here so I'm not sure it's a change worth pursuing. Barring any objections I plan on changing the standard.

Acknowledge: Alex Christensen.

annevk:
For URLs without a special scheme we cannot use the host parser
directly due to compatibility issues and we need to serialize slightly
differently too.

Fixes #148.

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -231,9 +231,9 @@ point <a for=/>URLs</a> from <var>A</var> can come from untrusted sources.
      https://mothereff.in/punycode -->
 
 <p>A <dfn export id=concept-host>host</dfn> is a <a>domain</a>, an
-<a>IPv4 address</a>, or an <a>IPv6 address</a>. Typically a
-<a for=/>host</a> serves as a network address, but it is sometimes (ab)used as opaque
-identifier in <a for=/>URLs</a> where a network address is not necessary.
+<a>IPv4 address</a>, an <a>IPv6 address</a>, or an <a>opaque host</a>. Typically a <a for=/>host</a>
+serves as a network address, but it is sometimes used as opaque identifier in <a for=/>URLs</a>
+where a network address is not necessary.
 
 <p class=note>The RFCs referenced in the paragraphs below are for informative purposes only. They
 have no influence on <a for=/>host</a> syntax, parsing, and serialization. Unless stated
@@ -256,6 +256,13 @@ eight <dfn id=concept-ipv6-piece lt='IPv6 piece'>16-bit pieces</dfn>.
 
 <p class="note">Support for <code>&lt;zone_id></code> is
 <a href="https://www.w3.org/Bugs/Public/show_bug.cgi?id=27234#c2">intentionally omitted</a>.
+
+<p>An <dfn export>opaque host</dfn> is an <a>ASCII string</a> holding data that can be used for
+further processing.
+
+<p class="note no-backref">An <a>opaque host</a> is only used by <a lt="is special">non-special</a>
+<a for=/>URLs</a>.
+
 
 
 <h3 id=idna>IDNA</h3>
@@ -292,8 +299,8 @@ eight <dfn id=concept-ipv6-piece lt='IPv6 piece'>16-bit pieces</dfn>.
 <h3 id=host-syntax>Host syntax</h3>
 
 <p>A <dfn export id=syntax-host>host string</dfn> must be a <a>domain string</a>, an
-<a>IPv4 address string</a>, or "<code>[</code>", followed by an <a>IPv6 address string</a>, followed
-by "<code>]</code>".
+<a>IPv4 address string</a>, or: "<code>[</code>", followed by an <a>IPv6 address string</a>,
+followed by "<code>]</code>".
 
 <p>A <var>domain</var> is a <dfn>valid domain</dfn> if these steps return success:
 
@@ -334,6 +341,11 @@ separated from each other by "<code>.</code>".
      far as what developers can do we should be liberal
 
      XXX should we define the format inline instead just like STD 66? -->
+
+<p>An <dfn export>opaque-host string</dfn> must be zero or more <a>URL units</a>.
+
+<p class="note no-backref">This is not part of the definition of <a>host string</a> as it requires
+context to be distinguished.
 
 
 <h3 id=host-parsing>Host parsing</h3>
@@ -700,7 +712,7 @@ They serve no purpose other than being a location the algorithm can jump to.
  <a>IPv6 serializer</a> on <var>host</var>,
  followed by "<code>]</code>".
 
- <li><p>Otherwise, <var>host</var> is a <a>domain</a>, return <var>host</var>.
+ <li><p>Otherwise, <var>host</var> is a <a>domain</a> or <a>opaque host</a>, return <var>host</var>.
 </ol>
 
 The <dfn id=concept-ipv4-serializer>IPv4 serializer</dfn> takes an
@@ -813,15 +825,15 @@ an <a>ASCII string</a> identifying a user. It is initially the empty string.
 either null or an <a>ASCII string</a> identifying a user's credentials. It is initially
 null.
 
-<p>A  <a for=/>URL</a>'s <dfn export for=url id=concept-url-host>host</dfn> is either
-null or a <a for=/>host</a>. It is initially null.
+<p>A <a for=/>URL</a>'s <dfn export for=url id=concept-url-host>host</dfn> is null or a
+<a for=/>host</a>. It is initially null.
 
 <p>A  <a for=/>URL</a>'s <dfn export for=url id=concept-url-port>port</dfn> is either
 null or a 16-bit unsigned integer that identifies a networking port. It is initially null.
 
-<p>A  <a for=/>URL</a>'s <dfn export for=url id=concept-url-path>path</dfn> is a list of
-zero or more <a>ASCII string</a> holding data, usually identifying a location in
-hierarchical form. It is initially the empty list.
+<p>A <a for=/>URL</a>'s <dfn export for=url id=concept-url-path>path</dfn> is a list of zero or more
+<a>ASCII strings</a> holding data, usually identifying a location in hierarchical form. It is
+initially the empty list.
 
 <p>A  <a for=/>URL</a>'s <dfn export for=url id=concept-url-query>query</dfn> is either
 null or an <a>ASCII string</a> holding data. It is initially null.
@@ -939,7 +951,7 @@ input might be a <a>relative-URL string</a>.
 <ul class=brief>
  <li><p>a <a>URL-scheme string</a> that is an <a>ASCII case-insensitive</a> match for a
  <a>special scheme</a> and not an <a>ASCII case-insensitive</a> match for "<code>file</code>",
- followed by "<code>:</code>" and a <a>scheme-relative-URL string</a>
+ followed by "<code>:</code>" and a <a>scheme-relative-special-URL string</a>
  <li><p>a <a>URL-scheme string</a> that is <em>not</em> an <a>ASCII case-insensitive</a> match for a
  <a>special scheme</a>, followed by "<code>:</code>" and a <a>relative-URL string</a>
  <li><p>a <a>URL-scheme string</a> that is an <a>ASCII case-insensitive</a> match for
@@ -963,8 +975,8 @@ must be a <a>relative-URL string</a>, optionally followed by "<code>#</code>" an
 switching on <a>base URL</a>'s <a for=url>scheme</a>:
 
 <dl class=switch>
- <dt>Not "<code>file</code>"
- <dd><p>a <a>scheme-relative-URL string</a>
+ <dt>A <a>special scheme</a> that is not "<code>file</code>"
+ <dd><p>a <a>scheme-relative-special-URL string</a>
  <dd><p>a <a>path-absolute-URL string</a>
  <dd><p>a <a>path-relative-scheme-less-URL string</a>
  <dt>"<code>file</code>"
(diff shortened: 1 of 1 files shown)


annevk:
@achristensen07 what do you think of this approach?

achristensen07:
Seems good.  Make sure to add a test verifying something like "new URL('path', 'asdf://host')" serializes to 'asdf://host/path' adding the slash.  I broke something by forgetting that in my initial implementation.  I don't add a slash for query-only or fragment-only relative URLs, but that might be worth more discussion.

annevk:
@achristensen07 running https://quuz.org/url/liveview.html#notspecial://test?x in Safari TP does yield a slash for me.

achristensen07:
Yes, but "new URL('?query', 'asdf://host')" does not add a slash right now in STP.  Maybe this is inconsistent.

I think this is a good approach, but we might need another issue for these query-only and fragment-only URLs with non-special schemes.

url.bs
```diff
@@ -846,6 +846,14 @@ resource the <a for=/>URL</a>'s other components identify. It is initially null.
 
 <hr>
 
+<p>A <dfn export>non-special-URL host</dfn> is an <a>ASCII string</a> holding data that can be used
+for further processing.
+
+<p class="note no-backref">A <a>non-special-URL host</a> is emphatically not a <a for=/>host</a> and
+only used by <a lt="is special">non-special URLs</a>.
```
domenic: _is_ only used by...?

url.bs
```diff
@@ -2096,9 +2116,21 @@ then runs these steps:
  <li><p>If <var>url</var>'s <a for=url>cannot-be-a-base-URL flag</a> is set, append the first string
  in <var>url</var>'s <a for=url>path</a> to <var>output</var>.
 
- <li><p>Otherwise, append "<code>/</code>", followed by the strings in <var>url</var>'s
- <a for=url>path</a> (including empty strings), separated from each other by
- "<code>/</code>", to <var>output</var>.
+ <li>
+  <p>Otherwise, then:
```
domenic: Remove ", then" here

url.bs
annevk: "X is A and B" seems fine?

domenic: When you add "not"s it's confusing. "X is not A and B" could be either "X is not A and not B" or "X is not A and is B".

url.bs
```diff
@@ -1195,6 +1203,27 @@ different document encoding. Using the <a>UTF-8</a> encoding everywhere solves t
...
+  <ol>
+   <li><p>If <var>input</var> is the empty string, then return failure.
+   <!-- Otherwise parsing URLs would not be idempotent:
+
+        https://@/example.org/ -> https:///example.org/ -> https://example.org/ -->
+
+   <li><p>Return the result of <a lt="host parser">host parsing</a> <var>input</var>.
+  </ol>
+
+ <li><p>Return <var>input</var>.
```
annevk: I either need to percent-encode input here or in the state machine.

annevk:
The pathname getter should probably also be adjusted in a way similar to the serializer.

annevk:
> Yes, but "new URL('?query', 'asdf://host')" does not add a slash right now in STP. Maybe this is inconsistent.

Yeah I think that would be quite odd and required a slash in the tests (and standard) for now. I guess I should also test `asdf://host?query`.

url.bs
```diff
@@ -2081,9 +2104,11 @@ then runs these steps:
      <li><p>Append "<code>@</code>" to <var>output</var>.
     </ol>
 
-   <li><p>Append <var>url</var>'s <a for=url>host</a>,
+   <li><p>If <var>url</var> <a>is special</a>, then append <var>url</var>'s <a for=url>host</a>,
    <a lt="host serializer">serialized</a>, to <var>output</var>.
 
+   <li><p>Otherwise, append <var>url</var>'s <a for=url>host</a>, to <var>output</var>.
```
domenic: s/, to output/ to output/

domenic:
This doesn't seem to work either, testing in whatwg-url. In particular `sc://ñ` does not parse as having an empty path, but instead a path containing a single empty string item. Again could be a bug in my spec translation. I'll post a whatwg-url PR for you to review.

url.bs
annevk: I think that's currently somewhat consistently "wrong"...

annevk:
Yeah, you are right. Even with EOF you get a path that is non-empty.

Another problem here is not forbidding certain code points in the host setter.

I guess I should rethink this.

The other question of course is whether all this additional complexity is worth it, versus treating non-special schemes as just having a path and not being able to be base URLs for each other. (Which is what most browsers implement at the moment.)

annevk:
One thought I had is that it's perhaps worth splitting the non-special-URL host change and the change to not serialize a path if none was in the input.

But it might be easy to do the no path in input change. It only really matters if the URL is not special and you hit EOF during the host/hostname state, right? It should be easy enough to account for that.

annevk:
I'm going to pursue making this work, since it seems worthwhile to support relative URLs for non-special schemes, even if it comes at a cost somewhat greater than initially anticipated. I don't think this is the reason to move to the Chrome/Firefox model of effectively setting the cannot-be-base-URL flag for all non-special URLs.

I landed a commit that addresses the issue @domenic found, but I also need to look through the tests again since I think I missed updating a couple tests (in particular setter tests).

annevk:
I also updated the tests. I didn't add any new host/hostname setter tests however.

annevk:
I forgot to address "Another problem here is not forbidding certain code points in the host setter" so this is still not ready.

annevk:
The other thing I noticed that while testing is that currently we allow userinfo and a port to exist while host is empty. That shouldn't be allowed (and isn't in Safari).

annevk:
Self-review keeps finding issues. I also need to update the URL syntax section to account for these new hosts (and allow them to be empty).

annevk:
I think this is ready for review once more. I have also updated the tests. I hope we can get this to work. (I have to say, the alternative of simply not having hosts and relative URLs for non-special URLs is rather appealing at this point.)

annevk:
Sigh, I'd still appreciate early review, but I think we might want to allow for the possibility that path is empty but either query or fragment is not, as @achristensen07 suggested. That would more closely match the intent of the original RFC.

annevk:
This should be once again ready for review. Proposed commit message:
```
Change non-special URLs

This makes two major changes to non-special URLs:

* Parse host differently. Rather than IDNA processing, percent-encode.
* Allow path to be empty. Rather than enforce a path as we do for "https" and other special schemes, allow for the possibility of it being empty for non-special URLs, which more closely aligns with the obsoleted IETF RFC.

Tests: https://github.com/w3c/web-platform-tests/pull/4406.

Fixes #148.
```

annevk:
@domenic would really appreciate review from you. Or maybe I can interest @jasnell? Does Node.js run web-platform-tests?

url.bs
```diff
@@ -1195,6 +1223,38 @@ different document encoding. Using the <a>UTF-8</a> encoding everywhere solves t
...
 <hr>
 
+<p>The <dfn export id=concept-url-host-parser>URL-host parser</dfn> takes a string <var>input</var>
+and a boolean <var>isSpecial</var>, and then runs these steps:</p>
+
+<ol>
+ <li><p>If <var>isSpecial</var> is true, then return the result of
+ <a lt="host parser">host parsing</a> <var>input</var>.
+
+ <li><p>Let <var>output</var> be the empty string.
```
domenic: This step can move below the next one.

url.bs
```diff
@@ -2096,9 +2173,10 @@ then runs these steps:
  <li><p>If <var>url</var>'s <a for=url>cannot-be-a-base-URL flag</a> is set, append the first string
  in <var>url</var>'s <a for=url>path</a> to <var>output</var>.
 
- <li><p>Otherwise, append "<code>/</code>", followed by the strings in <var>url</var>'s
- <a for=url>path</a> (including empty strings), separated from each other by
- "<code>/</code>", to <var>output</var>.
+ <li><p>Otherwise, if <var>url</var>'s <a for=url>path</a>
+ <a for=list lt="is empty">is not empty</a>, then return append "<code>/</code>", followed by the
```
domenic: extra "return"

url.bs
```diff
@@ -1875,7 +1943,16 @@ string <var>input</var>, optionally with a <a>base URL</a> <var>base</var>, opti
 
           <p class=note>This is a (platform-independent) Windows drive letter quirk.
 
-         <li><p>Append <var>buffer</var> to <var>url</var>'s <a for=url>path</a>.
+         <li>
+          <p>If one of the following is true
+
+          <ul class=brief>
+           <li><p><var>url</var>'s <a for=url>path</a> is <a for=list lt="is empty">is not empty</a>
```
domenic: Double "is"

domenic:
@annevk you may want to try hacking on https://github.com/jsdom/whatwg-url/pull/68 for faster turnaround time. It should be pretty straightforward: clone, switch to that branch, npm install, then npm test.

TimothyGu:
> Does Node.js run web-platform-tests?

No, we don't run WPT for URL parsing at this moment. It is something I would like to change.

However, I have indeed imported all the `searchParams` WPT tests for a spec-compliant `URLSearchParams` interface (see nodejs/node#9484).

jasnell:
Well.. we run the suite of URL parser and setter tests adopted from the web-platform-tests. We do not run the complete set. (https://github.com/nodejs/node/blob/master/test/fixtures/url-tests.json and https://github.com/nodejs/node/blob/master/test/fixtures/url-setter-tests.json)

annevk:
Sigh. I believe the current design is fundamentally broken.

By the time you get to the "path state", any knowledge of an empty path or a path solely consisting of "/" is gone. We should probably redo that and instead make special URLs append the empty string to path if they have no path somehow.

I should probably play around with whatwg-url some more to determine how much work that is and whether it's worth splitting the host and path changes into separate PRs.

annevk:
I opened #212 for working through the path changes. Restricting this to changing hosts.

annevk:
@domenic how do I run (a custom) whatwg-url against a particular branch of web-platform-tests?

domenic:
@annevk change the hash in https://github.com/jsdom/whatwg-url/blob/0f228168cb3aa31366c182f4114ec4d821fb351d/scripts/get-latest-platform-tests.js#L17 to the latest commit on that branch, then run `npm test`. It will automatically pull down the .json files from that branch.

annevk:
@achristensen07 I would appreciate if you could take another look at the changes proposed here and the test changes at https://github.com/w3c/web-platform-tests/pull/4406. I verified that they are in sync with each other using jsdom/whatwg-url per super helpful instructions from @domenic above.

This does not change any of the path handling, that is now #212. It does introduce opaque hosts (non-special-URL host felt a little verbose) and fixes parsing for when userinfo is present and the host is not and when port is present and the host is not. It poses fairly minimal requirements on the syntax of an opaque host.

achristensen07:
Here we introduce yet another seemingly arbitrary unique set of code points we don't like for some reason.  Where did this list come from?  Why don't we like them?
U+0000, U+0009, U+000A, U+000D, U+0020, U+0023, U+002F, U+003A, U+003F, and U+0040
I think this part should be left out unless there's a reason why these and only these code points should be forbidden only in non-specially-schemed URL hosts.  It would make more sense to me to use the same set of code points as the hosts of specially-schemed URLs.

I like the test for http://? and http://# and it's something I need to fix in WebKit but it seems unrelated to this change.  Also unrelatedly, I have found three compatibility issues with people using "https://" as a base url and expecting it to parse successfully as it used to do in WebKit but doesn't any more because of compliance with this spec.  Do we want to accommodate such uses, or continue to tell them to just update their content?

annevk:
@achristensen07 it seems if those code points would become part of the host you would get serialization issues. I don't think they can become part of the host due to the URL tokenizer so maybe it's fine to remove them there, assuming nobody invokes the host parser directly. Although I suppose the code point that does not apply to is U+0020.

I don't understand what `https://` as a base URL would do. Are they using `//host/path` scheme-relative URL strings and expect them to be HTTPS once parsed?

achristensen07:
Yep, that's exactly what they are doing.

annevk:
I don't know. Chrome does not support that either. Seems like it should be a separate issue if we need to add that.

annevk:
Also, come to think of it, it's basically the same code points that are restricted. I just removed the code points that are not problematic for non-special URLs, such as [ and ] (restricted for IPv6).

annevk:
We could keep the list of restricted code points the same though if you think that is better. I don't care strongly.

achristensen07:
"https://" is now invalid in WebKit.  I don't think it's worth allowing it in the spec.  It's a relatively small compatibility issue.

annevk:
@domenic this PR and the corresponding tests at https://github.com/w3c/web-platform-tests/pull/4406 can be reviewed again. I have verified it with whatwg-url, but haven't updated my patch for it. Since it's a rather trivial change it seems good to use that as a  test to ensure you get the same results.

url.bs
```diff
@@ -1541,8 +1587,19 @@ string <var>input</var>, optionally with a <a>base URL</a> <var>base</var>, opti
...
+      <p>then run these substeps:
+
+      <ol>
+       <li><p>If <var>@ flag</var> is set and <var>buffer</var> is the empty string,
+       <a>syntax violation</a>, return failure.
+       <!-- No URLs with userinfo, but without host. For special URLs it would also not be
+            idempotent:
+            https://@/example.org/ -> https:///example.org/ -> https://example.org/ -->
+
+       <li><p>Decrease <var>pointer</var> by the number of code points in <var>buffer</var> plus
```
domenic: These could be separate substeps

annevk: I left this as-is since the parser generally groups steps that can reasonably be grouped.

PR #186, opened 2016-12-28
No need for null passwords

### LINKED ISSUE #181: empty auth vs no auth -- same?
`http://:@domain/` (empty username and password)
vs
`http://domain/`

Are they the same?

annevk:
It turns out that preserving the ability to serialize an empty string
password is not something any implementation supports.

Fixes #181.

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -894,9 +894,10 @@ Referrer Policy. [[FETCH]] [[REFERRER-POLICY]]
 <p class="note no-backref"><a>HTTP(S) scheme</a>, <a>network scheme</a>, and <a>fetch scheme</a> are
 used by HTML. [[HTML]]
 
-<p>A <a for=/>URL</a> <dfn export lt="include credentials">includes credentials</dfn> if its
+<p>A <a for=/>URL</a>
+<dfn export lt="include credentials|includes credentials">includes credentials</dfn> if its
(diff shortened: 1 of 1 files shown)


domenic:
To be fair, I haven't tested Safari TP yet, and they went to a lot of trouble to become spec-compliant, so we should check that first... I'm heading into work (and thus to my testing Mac) so I should be able to get back to you soon.

annevk:
Safari TP normalizes it away. I checked.

annevk:
The difference between Firefox and others is with `https://test:@test/`. Then `:` does not disappear. I'll file a bug on that.

annevk:
https://bugzilla.mozilla.org/show_bug.cgi?id=1326175

url.bs
```diff
@@ -2067,16 +2056,15 @@ then runs these steps:
    <li><p>Append "<code>//</code>" to <var>output</var>.
 
    <li>
-    <p>If <var>url</var>'s <a for=url>username</a> is not the empty string
-    or <var>url</var>'s <a for=url>password</a> is non-null, run these substeps:
+    <p>If <var>url</var>'s <a for=url>username</a> or <a for=url>password</a> is not
+    the empty string, then:
```
domenic: This could probably use the "includes credentials" concept?

PR #189, opened 2016-12-29
IPv4 parser: handle the empty string

### LINKED ISSUE #79: Parsing an empty host
If I’m reading the parsing algorithm correctly, parsing `some-non-special-scheme:///` causes the host parser to be called with an empty string. Since _domain to ASCII_ sets _VerifyDnsLength_ to false, _asciiDomain_ is also empty (not failure) and IPv4 parsing is called with an empty string. There, _parts_ is first a list of one empty string, then an empty list:

> Let parts be input split on ".".
> If the last item in parts is the empty string, set syntaxViolationFlag and remove the last item from parts.

Later, _numbers_ stays empty and some of the steps are nonsensical when they talk about the last item of _numbers_.

So:
- The IPv4 parser should probably return failure for the empty string as a first step
- This would make the host of this example URL an empty domain. Should it be null instead?

annevk:
This is a theoretical issue once #185 lands, but as discussed it seems
good to address this anyway. Fixes #79.

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -450,8 +450,17 @@ runs these steps:
 
  <li><p>Let <var>parts</var> be <var>input</var> split on "<code>.</code>".
 
- <li><p>If the last item in <var>parts</var> is the empty string, set
- <var>syntaxViolationFlag</var> and remove the last item from <var>parts</var>.
+ <li>
+  <p>If the last item in <var>parts</var> is the empty string, then:
+
+  <ol>
+   <li><p>Set <var>syntaxViolationFlag</var>.
+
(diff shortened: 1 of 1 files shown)


annevk:
This doesn't require test changes as you cannot hit this scenario currently.

annevk:
This should land after 185, otherwise the added comment is a little off.

domenic:
Can't really test until we figure out https://github.com/whatwg/url/pull/185; please ping me again when we finish that, if I forget.

annevk:
@domenic consider yourself pinged.

Issue #61, opened 2015-08-14
URLUtils is wrong for Location object


annevk:
The `Location` object probably requires special treatment since setting its setters do not actually change the underlying URL directly. They just cause navigation. And the last navigation seems to win, too:

``` js
location.pathname = "x"
location.protocol = "https"
```

Navigates to a URL with a new scheme, not a new path.

annevk:
`Location.prototype.protocol` is also special in that it throws for invalid schemes. But it also ignores schemes that are not compatible. Effectively only `https`, `http`, and `ftp` seem to function. See this older bug report https://www.w3.org/Bugs/Public/show_bug.cgi?id=23523 by @arv which I'll close in favor of this issue.

domenic:
This has been fixed in HTML.

annevk:
I think the invalid protocols still needs to be fixed unfortunately. Requires some further study.

annevk:
Looking at the setter in https://html.spec.whatwg.org/multipage/browsers.html#dom-location-protocol it seems to anticipate a change in the URL Standard where we sometimes return failure for bogus schemes. That will then result in an exception.

I've also noticed that browsers will change the scheme of data URLs and attempt to navigate, which would go against the HTML Standard. They also appear to use DOMException "SyntaxError" rather than TypeError. The exception we should probably change in the HTML Standard, but browsers might want to update to change the scheme less overall.

annevk:
Once those tests lands this can be closed once bugs have been filed against browsers.

annevk:
Bugs filed thus far:

https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/10422816/
https://bugzilla.mozilla.org/show_bug.cgi?id=1328843

annevk:
https://bugzilla.mozilla.org/show_bug.cgi?id=1328894
https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/10426237/
https://bugs.webkit.org/show_bug.cgi?id=166729
https://bugs.chromium.org/p/chromium/issues/detail?id=678633

PR #194, opened 2017-01-03
Return failure in state override scheme parsing

annevk:
This is necessary to define the behavior of the Location object’s
protocol attribute, which throws on syntax errors rather than silently
ignoring them as other protocol setters do.

Tests: https://github.com/w3c/web-platform-tests/pull/4412.

This is the cleanup work mentioned in
https://github.com/whatwg/html/commit/f0a73659b4046cc35a28855f3544dead66345689
that is happening way late.


annevk:
When committing this, mention #61. Closing that properly warrants a couple more tests I think for non-broken schemes.

PR #196, opened 2017-01-06
Disallow invalid IPv4 in IPv6 parser

### LINKED ISSUE #195: IPv6 parser's bug in IPv4 part
URL parser's algorithm correctly reports as valid:
`http://[::1.2.3.4]`

But slightly modified and invalid URLs are reported as valid:
`http://[::1.2.3.]`
`http://[::1.2.]`
`http://[::1.]`

In the 10 step's 7 substep last '.' is skipped (there no failure at 4 substep) and, as EOF code point is reached, after 9 supstep the 10 step ends.

Similarly also isn't reported as invalid:
`http://[::1.2.3.4x]`

In the 7 substep 'x' is skipped (this character is not checked, so there can be any non ASCII digit), EOF code point is reached so no failure at 8 substep and after 9 substep the 10 step ends.

These URLs I also checked with [Live URL Viewer](https://quuz.org/url/liveview.html) and [rust-url](https://docs.rs/crate/url/): the same problem.

annevk:
Fixes #195.

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -607,7 +607,7 @@ then runs these steps:
  <li><p><dfn id=concept-ipv6-parser-ipv4 lt='IPv6 parser IPv4'>IPv4</dfn>:
  If <var>piece pointer</var> is greater than six, <a>syntax violation</a>, return failure.
 
- <li><p>Let <var>dots seen</var> be 0.
+ <li><p>Let <var>numbersSeen</var> be 0.
 
  <li>
   <p>While <a>c</a> is not the <a>EOF code point</a>, run
@@ -615,6 +615,16 @@ then runs these steps:
 
   <ol>
    <li><p>Let <var>value</var> be null.
+
+   <li>
+    <p>If <var>numbersSeen</var> is greater than 0, then:
+
+    <ol>
+     <li><p>If <a>c</a> is a "<code>.</code>" and <var>numbersSeen</var> is less than 4, then
+     increase <var>pointer</var> by one.
+
+     <li>Otherwise, <a>syntax violation</a>, return failure.
+    </ol>
 
    <li><p>If <a>c</a> is not an <a>ASCII digit</a>, <a>syntax violation</a>,
    return failure. <!-- prevent the empty string -->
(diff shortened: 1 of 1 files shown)


annevk:
The reporter still needs to be acknowledged. Tests coming up.

annevk:
Tests: https://github.com/w3c/web-platform-tests/pull/4513.

domenic:
This does not appear to give the correct results.

Given input `::1.`, there should be a failure, but following the new algorithm, there is none:

step 5: pointer = 2, piece pointer = compress pointer = 1

step 6 (main):
 - step 4: value = 1, pointer = 3, length = 1
 - step 6: c = "." => pointer = 2, ipv4 = true

step 10:
 - first round, c = "1"
   - step 3: value = 1, pointer = 3
   - step 4: no failure (c is "." now because pointer incremented in step 3)
   - new step "if dots seen is 3 and c is not the EOF code point, failure": no failure (c is ".", not EOF)
   - new step "if c is "." increase pointer and dots seen by one": pointer is now 4, so c = EOF, so loop exits

annevk:
Thanks @domenic for the detailed run through. I pushed something that should definitely fail for EOF after ".".

rmisev:
Now for invalid IPv4 addresses with more than 3 dots, parser does unnecessary loop and calculations until reaches EOF code point.

I think additional check is required. I created two versions of fix here:

https://github.com/rmisev/url/commits/ipv4-in-ipv6-v1
https://github.com/rmisev/url/commits/ipv4-in-ipv6-v2

rmisev:
I think the point of failure in your algorithm can be [Integer overflow](https://en.wikipedia.org/wiki/Integer_overflow) in step "4. Set <i>piece</i> to <i>piece</i> × 0x100 + <i>value</i>" (when invalid address has many dots). In some programming languages this can cause a run-time error.

So I preffer <i>numbers seen</i> variant.

annevk:
That's a good point. @domenic mind reviewing this approach? Sorry about the churn.

annevk:
Thanks a lot for the patch @rmisev!

PR #197, opened 2017-01-06
Only support UTF-8 for the urlencoded parser

### LINKED ISSUE #84: Restrict application/x-www-form-urlencoded parser to UTF-8
I guess we need to introduce decode without BOM for this special use case in Encoding. The idea here is that Encoding only exposes high-level hooks to other specifications, so its internal algorithms can be reshuffled if we wanted to.

annevk:
Fixes #84.


annevk:
Should I also copy all of http://searchfox.org/mozilla-central/source/dom/url/tests/test_urlSearchParams.html#67 here? @bakulf?

Issue #90, opened 2016-02-15
URLSearchParams: calling .set() and .append() should return itself


marcoscaceres:
Given a list of search params (strings) to be inserted into a URLSearchParams object, it would be useful if calling .set() and .append() would return the instance of URLSearchParams. Two reasons: 
1. URLSearchParams, although not "map-like", is extremely close to an ES Map in look and feel. ES `Map()` returns itself when a developer calls `set()`. Similarly, so do the calls on ES Set(). It is surprising that URLSearchParams does not work the same.  
2. Using `listOfURL.reduce()` (or other functions that operate on lists that expect an object in return) would benefit from this.  

Consider, this would allow the following code: 

``` JS
function buildQuery(items) {
  const params = new URLSearchParams();
  items
    .map(item => item.url)
    // This is not great, requires unnecessary "{ }", ";" x2, and "return"  :(
    .reduce((params, url) => {
       params.append("foo", url);
       return params;
    }, params)
  return params.toString();
}
```

Could become much cleaner:

``` JS
function buildQuery(listOfThings) {
  return items
    .map(listOfThings => listOfThings.url)
    .reduce((params, url) => params.append("foo", url), new URLSearchParams())
    .toString();
}
```

//cc @domenic

marcoscaceres:
(fixed up first code example above)

annevk:
This pattern of returning this is not something TC39 is still advocating I believe. That's why we didn't copy it. It makes it impossible to introduce a useful return value in the future.

marcoscaceres:
Ok, sent to public-script-coord, to hopefully get a little more ocntext. Closing for now. 
https://lists.w3.org/Archives/Public/public-script-coord/2016JanMar/0096.html

marcoscaceres:
Reopening, as there was support on public-script-coord for adding this.

annevk:
The reason I haven't done this yet by the way is that the support was not overwhelming and there's other differences between `Set`/`Map` and `URLSearchParams`/`FormData`/`Headers` (such as the latter returning null). It'll also require some implementation and testing effort.

marcoscaceres:
It would be good to do it for all of them tho (and because they currently return undefined, it might mean that this can be done in a somewhat backwards compatible way). 

This not returning self might become an issue - in as far that it's another place where the platform doesn't match ES! - as more people start migrating to ES6 and beyond, because all other map/set interfaces return self. This makes these interfaces annoying to use when writing functional code. 

[...]

PR #199, opened 2017-01-12
Add URLSearchParams.prototype.sort()

### LINKED ISSUE #26: Sort URLSearchParams prior to stringification
Inspired by [this tweet](https://twitter.com/igrigorik/status/605766338687016960), I'd argue that the specified behavior for the [`URLSearchParams` stringifier](https://url.spec.whatwg.org/#urlsearchparams) should be changed to either automatically sort by default (unlikely, due to @annevk's [objections](https://twitter.com/annevk/status/605887980570411008)), or alternatively to support some sort of boolean opt-in to enable sorting.

I'd imagine that sorting based on parameter names, falling back to parameter values in the case of repeated parameters, would make the most sense.

annevk:
This method is added to increase cache hits when making requests. It’s
opt-in as the order of code points in a URL’s query is significant by
default. It’s up to applications to decide if name order is not
significant for them.

Tests: https://github.com/w3c/web-platform-tests/pull/4531.

Fixes #26.


annevk:
@tabatkins @plinss https://api.csswg.org/bikeshed/ returns 400s to Travis. Something going on?

annevk:
Apologies for having pinged you. It was my error.

annevk:
Feedback: @domenic requests an example for non-destructive sort.

url.bs
```diff
@@ -2858,10 +2866,19 @@ method, when invoked, must run these steps:
...
-<dfn method for=URLSearchParams><code>has(<var>name</var>)</code></dfn>
-method, when invoked, must return true if there is a name-value pair whose name is <var>name</var>
-in <a for=URLSearchParams>list</a>, and false otherwise.
+<hr>
+
+<p>The <dfn method for=URLSearchParams><code>sort()</code></dfn> method, when invoked, must run
+these steps:
+
+<ol>
+ <li><p>Lexicographically sort all name-value pairs, if any, by name, while preserving the relative
```
domenic: As I mentioned in the tests PR, do we have a definition for "lexicographically"? I assume that means é sorts after z, not before f, but I can't be sure.

annevk: I guess we don't really although we do use it here and there. I wonder if @r12a or @aphillips can help us out. I always assumed it meant code point order.

annevk: We use the term for the `Headers` object too, but that's restricted to bytes (although I guess even there it might be confusing, does "@" come before or after "a").

jasnell: The [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort) docs for Array.prototype.sort explicitly calls out using "Unicode codepoint order". I'd recommend the same wording here.

annevk: Having looked at https://tc39.github.io/ecma262/#sec-abstract-relational-comparison I think we want code unit order, not code point order. So something like:

> Sort all name-value pairs, if any, by name on code unit, while preserving the relative order between duplicate names, if any.

domenic: Sounds good, but breaking that up into a few sentences wouldn't hurt IMO:

> Sort all name-value pairs, if any, by their names. Sorting must be done by comparison of code units. The relative order between name-value pairs with equal names must be preserved.

annevk:
Note to self: file some follow-up issues for other places that use "alphabetical" (HTML), "lexicographical" (Fetch), and look if we have more places that could use clarification.

url.bs
```diff
@@ -2858,10 +2866,36 @@ method, when invoked, must run these steps:
...
+ <pre><code class=lang-javascript>
+const url = new URL("https://example.org/?q=🏳️‍🌈&amp;key=e1f7bc78");
+url.searchParams.sort();
+url.search; // "?key=e1f7bc78&amp;q=%F0%9F%8F%B3%EF%B8%8F%E2%80%8D%F0%9F%8C%88"</code></pre>
+
+ <p>To avoid altering the original input, e.g., for comparison purposes, construct a new
+ {{SearchParams}} object:
+
+ <pre><code class=lang-javascript>
+const sorted = (new URLSearchParams(url.search)).sort();</code></pre>
```
TimothyGu: Currently `sort()` returns void, so we either need to make `sort` return `this` or change this example.

annevk: Thanks, I opted for fixing the example. Might change if we fix #90 as requested I suppose.

url.bs
```diff
@@ -2969,6 +3004,7 @@ Tab Atkins,
 Tantek Çelik,
 Tim Berners-Lee,
 簡冠庭 (Tim Guan-tin Chien),
+Timothy Gu,
```
TimothyGu: Please attribute to me as

> Tiancheng "Timothy" Gu

which contains my legal name.

annevk: Thanks, should have asked.

annevk:
@TimothyGu all good now?

aphillips:
The new text doesn't define 'code units' and neither does the document directly or by reference. In addition, it doesn't specify why sort of code units are being sorted. Sorting into UTF-8 byte order doesn't seem to be the intention. I suspect that UTF-16 code unit order is expected. Make that explicit?

I could argue for code point order, but suspect the overhead isn't worth it.

annevk:
The plan is to make sure it's defined in https://infra.spec.whatwg.org/ in due course. https://github.com/whatwg/infra/issues/1 has various ideas and https://github.com/whatwg/infra/issues/55 tracks sorting in particular. I haven't really come up with a concrete proposal yet since in part I feel that I should study more of the various usage patterns we have first.

PR #203, opened 2017-01-14
Editorial: fix JavaScript syntax

TimothyGu:


annevk:
Thanks, while you're there, could you also add double quotes around the string "key=730d67" below? And add your name to the acknowledgments section. (Or I can do that as part of the sort() change.)

annevk:
Thanks! (It would have been to leave the single quotes in the example itself by the way, but output should be consistent.)

Issue #201, opened 2017-01-13
Avoid using "simply"


annevk:
It's not very friendly to those new to a topic. (Also: add this to the style guide and check other standards.)

annevk:
#197 and #207 address most of this. I still have "simple encode set" and "default encode set" and although harmless, those names are rather meaningless too. But maybe best to leave in place until all percent-encoding issues are fully resolved.

PR #207, opened 2017-01-15
Editorial: use simple/y less

annevk:
Fixes part of #201.


zcorpan:
What about this one?

```
<p>The <dfn export id=concept-urlencoded><code>application/x-www-form-urlencoded</code></dfn> format
is a simple way to encode name-value pairs in a byte sequence where all bytes are
<a>ASCII bytes</a>.
```

annevk:
I have a fix for that in #197. And I'm leaving the issue open to think of better terms for simple and default encode set once all the parser issues are settled.

PR #213, opened 2017-01-21
Change path parsing for non-special URLs

### LINKED ISSUE #212: Redo path parsing for non-special URLs
As part of #148 and #185 we noticed that if we want to keep parsing non-special URLs in a way that preserves the ability for them to be base/relative URLs we need to make some changes, that would also bring them closer to the original RFCs.

I tried to do that as part of #185 but failed multiple times, so I'm going to make it a dedicated issue. What I think needs to happen is that we stop eating "/" early on. I think we need to drop the "path start state" as well.

What we want is that for special URLs we ensure that path has at least one item (potentially the empty string) and that for non-special URLs it can be empty if there is no path (e.g., as in `foo://x#x` having no path, but `http://x#x` becoming `http://x/#x` because special).

Currently "relative slash state" goes down "special authority ignore slashes state" even for non-special URLs. That means `///x` against `foo://y` results in `foo://x` rather than `foo:///x`. That also needs to be fixed.

annevk:
This allows paths to be empty for non-special URLs and also takes that
into account when serializing.

Tests: https://github.com/w3c/web-platform-tests/pull/4586.

Fixes #212.

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -837,6 +837,9 @@ null or a 16-bit unsigned integer that identifies a networking port. It is initi
 <a>ASCII strings</a> holding data, usually identifying a location in hierarchical form. It is
 initially the empty list.
 
+<p class="note no-backref">A <a lt="is special">special</a> <a for=/>URL</a> always has a
+<a for=list lt="is empty">non-empty</a> <a for=url>path</a>.
+
 <p>A  <a for=/>URL</a>'s <dfn export for=url id=concept-url-query>query</dfn> is either
 null or an <a>ASCII string</a> holding data. It is initially null.
 
@@ -1512,14 +1515,17 @@ string <var>input</var>, optionally with a <a>base URL</a> <var>base</var>, opti
    <dd>
     <ol>
      <li>
-      <p>If either <a>c</a> is "<code>/</code>", or <var>url</var> <a>is special</a> and
-      <a>c</a> is "<code>\</code>", run these substeps:
+      <p>If <var>url</var> <a>is special</a> and <a>c</a> is "<code>/</code>" or "<code>\</code>",
+      then:
 
       <ol>
        <li><p>If <a>c</a> is "<code>\</code>", <a>syntax violation</a>.
 
        <li><p>Set <var>state</var> to <a>special authority ignore slashes state</a>.
       </ol>
+
+     <li><p>Otherwise, if <a>c</a> is "<code>/</code>", then set <var>state</var> to
+     <a>authority state</a>.
 
      <li><p>Otherwise, set
      <var>url</var>'s <a for=url>username</a> to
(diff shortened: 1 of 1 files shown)


annevk:
@domenic this is ready for review too, but note that it builds on top of the opaque host change.

url.bs
```diff
@@ -1877,12 +1883,27 @@ string <var>input</var>, optionally with a <a>base URL</a> <var>base</var>, opti
...
+     <li>
+      <p>If <var>url</var> <a>is special</a>, then:
 
-     <li><p>Set <var>state</var> to <a>path state</a>, and if neither <a>c</a> is
-     "<code>/</code>", nor <var>url</var> <a>is special</a> and <a>c</a> is
-     "<code>\</code>", decrease <var>pointer</var> by one.
+      <ol>
+       <li><p>If <a>c</a> is "<code>\</code>", <a>syntax violation</a>.
+
+       <li><p>Set <var>state</var> to <a>path state</a> and if <a>c</a> is neither "<code>/</code>"
```
domenic: This really seems like it'd be better as two steps.

domenic:
This does not quite match the tests yet. I noticed that the spec does not appear to match the parse path start changes in https://github.com/whatwg/url/issues/212#issuecomment-274252010. You can see the spec translated into whatwg-url at https://github.com/jsdom/whatwg-url/pull/70.

The failing tests are:

```
 1) Web platform tests parsing <http:\\foo.com\> against <http://example.org/foo/bar>:

      AssertionError: href
      + expected - actual

      -http://example.org/foo.com/
      +http://foo.com/

  2) Web platform tests parsing <http:\\a\b:c\d@foo.com\> against <http://example.org/foo/bar>:

      AssertionError: href
      + expected - actual

      -http://example.org/a/b:c/d@foo.com/
      +http://a/b:c/d@foo.com/

  3) Web platform tests parsing <..> against <file:///1:/>:

      AssertionError: pathname
      + expected - actual

      -/
      +//

  4) Web platform tests parsing <///> against <sc://x/>:

      AssertionError: href
      + expected - actual

      -sc://x///
      +sc:///

  5) Web platform tests parsing <////> against <sc://x/>:

      AssertionError: href
      + expected - actual

      -sc://x////
      +sc:////

  6) Web platform tests parsing <////x/> against <sc://x/>:

      AssertionError: href
      + expected - actual

      -sc://x////x/
      +sc:////x/
```

annevk:
I rewrote the path start changes to be a little neater, they should be identical. I found a fault in your code elsewhere that I think is more likely responsible for the failures you are seeing.

annevk:
I still get two failures I find a little harder to narrow down. Investigating.

annevk:
It seems those two failures were actually broken tests. Fixed in the web-platform-tests PR. Provided a patch in the whatwg-url PR.

Will address your style nit here and then I think we're back to all good.

annevk:
@achristensen07 I assume you are tracking this in WebKit?

I filed no bugs on other browsers since they're not actively implementing this part of the standard yet.

achristensen07:
Yep, I'm following this.  I actually needed to implement this a few months ago to maintain compatibility.  It was implemented in pieces as I found the compatibility issues.
Started with https://bugs.webkit.org/show_bug.cgi?id=163373 (yuck)
Getting closer in https://bugs.webkit.org/show_bug.cgi?id=163413
https://bugs.webkit.org/show_bug.cgi?id=164290 fixed issue 148 which is related
https://bugs.webkit.org/show_bug.cgi?id=165621 was also necessary for this

PR #216, opened 2017-01-24
Host parser UTF-8 decoding should fail, not U+FFFD

### LINKED ISSUE #215: Host parsing UTF-8 decoding can fail
Host parsing step 2 currently reads like this:
"Let domain be the result of UTF-8 decode without BOM on the percent decoding of UTF-8 encode on input."
If the UTF-8 decoding fails, it should explicitly say to return failure.  This matches Chrome, Firefox, and WebKit after https://bugs.webkit.org/show_bug.cgi?id=167330

annevk:
Tests: …

Fixes #215.

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -357,16 +357,15 @@ steps:
    "<code>]</code>" removed.
   </ol>
 
- <li><p>Let <var>domain</var> be the result of
- <a>UTF-8 decode without BOM</a> on the
- <a lt="percent decode">percent decoding</a> of
- <a>UTF-8 encode</a> on <var>input</var>.
- <!-- https://bugzilla.mozilla.org/show_bug.cgi?id=309671 -->
+ <li><p>Let <var>domain</var> be the result of running <a>UTF-8 decode without BOM or fail</a> on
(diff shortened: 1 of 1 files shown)


annevk:
Tests: https://github.com/w3c/web-platform-tests/pull/4617.

domenic:
This appears to be unnecessary. [ToASCII](http://www.unicode.org/reports/tr46/#ToASCII) applied to something with replacement characters will generate an error result, because [Processing](http://www.unicode.org/reports/tr46/#Processing) step 1 ends up in "disallowed" for the replacement character.

annevk:
@achristensen07 it seems your implementation of ToASCII might be buggy.

@domenic thanks, I'll update this PR to instead make a note that implementations could fail early if they desire.

annevk:
I should probably add an explicit test for https://�/ since Chrome and Firefox (haven't tested Edge) still turn that into something. But in general we need more IDNA tests.

annevk:
Test for U+FFFD: https://github.com/w3c/web-platform-tests/pull/4639. I also updated the PR to make it a note instead.

Issue #214, opened 2017-01-23
Forbidden host code points


annevk:
In https://github.com/whatwg/url/pull/185#issuecomment-274251275 @achristensen07 complained about the host code point restrictions I was adding for opaque hosts. I figured I'd explain the rules for code point restrictions in hosts in general once here and then we can either decide to agree or fiddle with the specifics.

Currently we have the following restrictions for non-opaque hosts. I listed the justification for each on the right:

* U+0000 (generally problematic)
* U+0009 (stripped when parsing URLs, would create reparsing issues)
* U+000A (stripped when parsing URLs, would create reparsing issues)
* U+000D (stripped when parsing URLs, would create reparsing issues)
* U+0020 (creates copy-and-paste issues, see #125)
* "#" (would create reparsing issues)
* "%" (would create reparsing issues due to host percent-decoding)
* "/" (would create reparsing issues)
* ":" (would create reparsing issues)
* "?" (would create reparsing issues)
* "@" (would create reparsing issues)
* "[" (would create reparsing issues)
* "\" (would create reparsing issues)
* "]" (would create reparsing issues)

Now for non-opaque hosts I took this list and removed the code points that were no longer problematic. And those are "%" (opaque hosts have no percent decoding), "[" (no IPv6), "\" (no special backslash handling), and "]" (no IPv6).

The reason to be maximally liberal is given in https://github.com/whatwg/url/issues/159#issuecomment-271674428.

@achristensen07 does this help or do you have the same concerns still? And if so, what would you do?

achristensen07:
This makes sense.  I just didn't see a documented reason for the seemingly-arbitrary list, but now it makes sense.

annevk:
Thank you, that should allow us to close a number of issues this week.

achristensen07:
For simplicity's sake, since this is so similar to the list of invalid domain characters, it might be nice to keep them the same.  That would prevent issues if people switch schemes, etc.

annevk:
Currently the specification disallows scheme switching for this reason (and also because non-special and special parse so differently), but making them the same seems reasonable nonetheless. I'll refactor the PR so that the list of code points is a reference rather than duplicated.

PR #218, opened 2017-01-26
Fix: allow percent encoded code points in the non-special URLs host

rmisev:
After 30362553e9ce9fc706d3492bd61886e399fc94e2 introduced the [URL-host parser](https://url.spec.whatwg.org/commit-snapshots/30362553e9ce9fc706d3492bd61886e399fc94e2/#concept-url-host-parser) rejects percent encoded code points in the non-special URLs host (because "%" is [forbidden host code point](https://url.spec.whatwg.org/#forbidden-host-code-point)).

The situation is paradoxical, because parser himself percent encodes some code points, actually producing invalid serialized URLs according to the current standard. 

URL's from #148 to test issue:
`unknown://%E2%80%A0/`
`notspecial://H%4fSt/path`

**Solution**
I suggest that forbidden host code points be checked on percent decoded code points. The decoded data will be used only for this purpose, so the <var>output</var> will not be affected by this change.

Also this is a closer to [special URLs host validity check](https://url.spec.whatwg.org/#concept-host-parser).

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -282,6 +282,19 @@ U+0020,
 or
 "<code>]</code>".<!-- 5D -->
 
+<p>A <dfn export>forbidden opaque host code point</dfn> is
+U+0000,
+U+0009,
+U+000A,
+U+000D,
+U+0020,
+"<code>#</code>",<!-- 23 -->
+"<code>/</code>",<!-- 2F -->
+"<code>?</code>",<!-- 3F -->
+"<code>@</code>",<!-- 40 -->
+or
+"<code>\</code>".<!-- 5C -->
+
 
 <h3 id=idna>IDNA</h3>
 
@@ -1237,12 +1250,7 @@ and a boolean <var>isSpecial</var>, and then runs these steps:</p>
  <li><p>If <var>isSpecial</var> is true, then return the result of
  <a lt="host parser">host parsing</a> <var>input</var>.
 
- <li><p>Let <var>buffer</var> be the result of
- <a>UTF-8 decode without BOM</a> on the
- <a lt="percent decode">percent decoding</a> of
- <a>UTF-8 encode</a> on <var>input</var>.
-
- <li><p>If <var>buffer</var> contains a <a>forbidden host code point</a>, <a>syntax violation</a>,
+ <li><p>If <var>input</var> contains a <a>forbidden opaque host code point</a>, <a>syntax violation</a>,
  return failure.
 
  <li><p>Let <var>output</var> be the empty string.
--- a/url.bs
+++ b/url.bs
@@ -274,19 +274,6 @@ U+0020,
 or
 "<code>]</code>".<!-- 5D -->
 
-<p>A <dfn export>forbidden opaque host code point</dfn> is
-U+0000,
-U+0009,
-U+000A,
-U+000D,
-U+0020,
-"<code>#</code>",<!-- 23 -->
-"<code>/</code>",<!-- 2F -->
-"<code>?</code>",<!-- 3F -->
-"<code>@</code>",<!-- 40 -->
-or
-"<code>\</code>".<!-- 5C -->
-
 
 <h3 id=idna>IDNA</h3>
 
@@ -373,7 +360,7 @@ context to be distinguished.
 
 <h3 id=host-parsing>Host parsing</h3>
 
-<p>The <dfn id=concept-host-parser>host parser</dfn> takes a string <var>input</var> and
+<p>The <dfn id=concept-host-parser>host parser</dfn> takes a string <var>input</var>, a boolean <var>isSpecial</var> and
 an optional <var>Unicode flag</var> (unset unless stated otherwise), and then runs these
 steps:
 
@@ -391,6 +378,9 @@ steps:
    with its leading "<code>[</code>" and trailing
    "<code>]</code>" removed.
   </ol>
+
+ <li><p>If <var>isSpecial</var> is false, then return the result of
+ <a lt="opaque host parser">opaque host parsing</a> <var>input</var>.
 
  <li>
   <p>Let <var>domain</var> be the result of running <a>UTF-8 decode without BOM</a> on the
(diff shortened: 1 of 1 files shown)


annevk:
@achristensen07, this is a good reason to allow "%" for opaque hosts. Though we could also go this route and attempt to decode first.

annevk:
I think my preferred fix here would be to simply not forbid "%" by special casing it in the URL-host parser.

rmisev:
I agree, it makes sense not to decode. Then more code points can be removed from forbidden list:

- We don't needed to check these because:
U+0009, U+000A and U+000D are removed by [basic URL parser](https://url.spec.whatwg.org/#concept-basic-url-parser) in step 3
"<code>#</code>", "<code>/</code>" and "<code>?</code>" are stop code points in [host state](https://url.spec.whatwg.org/#host-state)

- Maybe enable IPv6:
"<code>[</code>", "<code>:</code>", "<code>]</code>" ([host state](https://url.spec.whatwg.org/#host-state) guarantees that "<code>:</code>" appears only between "<code>[</code>" and "<code>]</code>")

Now forbidden list for opaque hosts can be as short as: U+0000, U+0020, "<code>\\</code>", "<code>@</code>"

annevk:
See also #214 for the previous discussion regarding this.

annevk:
In particular I think we should continue to include the code points you cannot reach to avoid issues down the road. Although potentially we could point out that you cannot reach them and that they are therefore excluded. That would be equally clear.

@achristensen07 any thoughts here?

rmisev:
I think must be excluded the "`%`" and "`[`", "`:`", "`]`" to allow IPv6 (for example `ssh:` scheme can use IPv6).
And yes, there no need to exclude not reachable code points.

annevk:
Yeah, that makes sense. It's unfortunate those schemes have to write their own IPv4/6 parser/validation though. Perhaps we need to expose host parsing separately as well to JavaScript at some point to make that easier.

jasnell:
A host parsing JS API would be welcome and would definitely be something worth exploring.

annevk:
@achristensen07 ping, could you please have a look at the proposed change? Safari is the only browser that supports hosts for non-special URLs so your review would be most welcome.

annevk:
@achristensen07 to clarify, the change basically makes it okay to use "%", "[", "]", and ":" (within "[" and "]") in opaque hosts. That way input such as `non-special://[%30:x]/` would work end up with a host of `[%30:x]`.

rmisev:
I further investigate this problem and I think a very basic IPv6 check can help here (change [URL-host parser](https://url.spec.whatwg.org/#concept-url-host-parser) 2 step to):

2. If <var>input</var> starts with "[" and ends with "]", then if <var>input</var> between "[" and "]" contais a [forbidden host code point](https://url.spec.whatwg.org/#forbidden-host-code-point) excluding ":", syntax violation, return failure.
3. Otherwise, if <var>input</var> contains a [forbidden host code point](https://url.spec.whatwg.org/#forbidden-host-code-point) excluding "%", syntax violation, return failure. 

This disallows such weird URLs: `non-special://[:80` or `non-special://]`

And term <dfn>"forbidden opaque host code point"</dfn> (304f0ae21880634842fff05bbcf9b03975a4344) will not be required anymore.

If you agree, I can prepare a new commit.

annevk:
That alternative seems reasonable too, but I think it's the best use of our time to wait for @achristensen07 to give input with regards to WebKit's preferences.

achristensen07:
We definitely need to allow percent-encoded values in non-special URLs to remain, but we need to do something about invalid percent-encoded values.  notspecial://H%4fSt/path is fine but notspecial://H%4gSt/path is not.  In such a case, should we percent-encode the percent?

annevk:
To me it seems fine to leave those be. We could at some point define an API to decode percent-encoded sequences and that could deal with those in various ways, but I'm not sure we need to make life difficult here necessarily. We don't do it elsewhere (e.g., paths, query, fragment).

What's your opinion on the IPv6 question and various solutions?

achristensen07:
My implementation tries to parse and canonicalize a host as an ipv6 address before going to the different behavior for special/non-special.  That's also what the current spec says.  We could disallow [ and ] if it's not a valid ipv6 address.  No problem.

annevk:
The current specification branches before that point, actually: https://url.spec.whatwg.org/#concept-url-host-parser.

achristensen07:
Sure enough.  My implementation doesn't.  We could make the spec match my implementation :)

annevk:
Okay, so your implementation branches right after step 1 of https://url.spec.whatwg.org/#concept-host-parser if I understand you correctly. That would be okay with me as well.

achristensen07:
Yep, that's exactly what I did.  I think it makes sense to have it there, and that makes all the host parsing in the same place in the spec.

annevk:
I adjusted the formatting and placement a bit. Let me know what you think. Are you interested in looking at test changes/new tests as well?

url.bs
```diff
@@ -407,21 +407,6 @@ steps:
...
-
- <li><p>Let <var>output</var> be the empty string.
-
- <li><p>For each code point in <var>input</var>, <a>UTF-8 percent encode</a> it using the
- <a>simple encode set</a>, and append the result to <var>output</var>.
-
- <li><p>Return <var>output</var>.
-</ol>
-
 The <dfn>IPv4 number parser</dfn> takes a string <var>input</var> and a
```
rmisev: It looks like the `<hr>` is missing before "IPv4 number parser"

rmisev: It looks like the `<hr>` is missing before "IPv4 number parser".

rmisev:
It looks like you missed `<hr>` before "IPv4 number parser".

I can add tests, I think to add these URLs:
`non-special://%E2%80%A0/`
`non-special://H%4fSt/path`
`non-special://[1:2::3]/`
`non-special://[1:2::3]:80/`
`non-special://[:80/`

I found one test in the current test suite, which will fail:
`sc://%/` (becomes valid after this patch)
So maybe remove it?

annevk:
I intentionally grouped the IPv4 parsers together.

Those tests look pretty good. Verifying that IPv6 is normalized would be good as well. And instead of removing the test I would change the pass condition.

domenic:
Verified that updating jsdom/whatwg-url according to this patchset passes all the tests in https://github.com/w3c/web-platform-tests/pull/4761 (and doesn't cause any failures).

annevk:
Thanks @rmisev!

PR #224, opened 2017-02-01
Cleanup API for file and non-special URLs

### LINKED ISSUE #97: file: URL with a port number through the host setter
When the `Url::host` setter is invoked for an URL in the `file` scheme, the parser starts in the "host state" (not the "file host state") so the URL can (per spec) end up with a non-null port number, which otherwise never happens.

In this test case: http://software.hixie.ch/utilities/js/live-dom-viewer/?saved=3927 Firefox doesn’t change the URL at all, Chromium changes the host but not the port number.

annevk:
This:

* Stops the username/password/port APIs from functioning when host is
the empty string.
* Makes the host/hostname APIs work better with file URLs and adjusts
the file host state accordingly.
* Make setting host/hostname to the empty string impossible when they
have a username/password/port.
* Fixes #97.

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -1668,6 +1668,10 @@ string <var>input</var>, optionally with a <a>base URL</a> <var>base</var>, opti
        <!-- http://? -> failure
             test://? -> test://? -->
 
+       <li><p>Otherwise, if <var>state override</var> is given, <var>buffer</var> is the empty
+       string, and either <var>url</var> <a>includes credentials</a> or <var>url</var>'s
+       <a for=url>port</a> is non-null, <a>syntax violation</a>, return.
+
        <li><p>Let <var>host</var> be the result of <a lt="URL-host parser">URL-host parsing</a>
        <var>buffer</var> with <var>url</var> <a>is special</a>.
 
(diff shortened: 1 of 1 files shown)


jasnell:
This is good but the third bullet (`make setting host/hostname to the empty string impossible when they have a username/password/port`) seems a bit unnatural to me. What *might* be better is that setting the host/hostname to the empty string would set the username/password/port to empty strings... but of course, that has side effects that may not be obvious to most users so I'm a bit torn on it.

url.bs
```diff
@@ -1859,28 +1863,42 @@ string <var>input</var>, optionally with a <a>base URL</a> <var>base</var>, opti
...
-       <var>state</var> to <a>path start state</a>.
+        <p class=note>This is a (platform-independent) Windows drive letter quirk. <var>buffer</var>
+        is not reset here and instead used in the <a>path state</a>.
+
+       <li>
+        <p>Otherwise, if <var>buffer</var> is the empty string, then:
+
+        <ol>
+         <li><p>If <var>state override</var> is given and <var>url</var>
+         <a>includes credentials</a>, <a>syntax violation</a>, return.
```
domenic: Is "return" the same as "terminate this algorithm" elsewhere?

domenic:
When updating the setters, unfortunately we need to update HTML's location and HTMLHyperlinkElementUtils as well, usually... :-/

url.bs
annevk: Yeah. I'm slowly switching to Infra terminology.

annevk:
@jasnell yeah, it's mostly an edge case I wanted to take care of to avoid messing up serialization. In particular creating a serialization that could not be parsed into the same components.

url.bs
domenic: I think the inconsistency is pretty confusing; when I implemented in whatwg-url I accidentally translated "return" as "bail out of the entire URL parsing algorithm".

domenic:
It occurs to me this might be cleaner and require fewer changes in the dependent APIs if you still went to the host/hostname state, but as the first step it just switched to the file host state if the scheme is file and state override is given.

annevk:
Would you prefer that? I have to say I have very mixed feelings about the whole state override setup.

domenic:
I think I would. I found the state override setup confusing at first but now that I understand how it works it seems fine. And given that, preserving the idea that we can just do the API stuff by directly moving into the state override makes sense to me.

url.bs
annevk: Isn't that what's happening?

domenic: No; it's bailing out of this particular state and back into the main parser loop. I.e. it goes back to step 11 ("Keep running the following state machine by switching on state. If after a run pointer points to EOF code point, go to the next step. Otherwise, increase pointer by one and continue with the state machine.")

domenic: Err, nevermind, that is what's happening. I guess what I meant is that I translated this as "bail out of this step and go back to step 11". Basically I saw two different types of bail-outs, and since I knew that "abort these steps" meant the one, I assumed that "return" meant the other one. I think it's best to change them all at once or not at all.

annevk:
Done. I also verified that the tests still pass with whatwg-url.

domenic:
LGTM but still would prefer not starting to use "return" without switching over everything all at once.

PR #225, opened 2017-02-01
File state did not correctly deal with lack of base URL

### LINKED ISSUE #217: file state needs clarification
It's not clear what to do in <i><b>file state</b></i>, when `c` is `EOF`, "`?`" or "`#`" and base is null or base scheme isn't "<code>file</code>". Possible variants:

1. Continue main loop (11. Keep running the following state machine by switching on state...)
2. Go to "<b>Otherwise</b>", and because "1.", "2." not meet, go to "3. Set state to path state, and decrease pointer by one"

Consider these URL's::
`file:`
`file:#frag`
`file:?q=v`

(1) variant produces following serialized url's:
`file:///` (with empty path list)
`file:///frag` (fragment converted to the path string)
`file:///q=v` (query converted to the path string)

(2) variant produces these serialized url's (path with one empty string):
`file:///`
`file:///#frag`
`file:///?q=v`

Although (2) looks better, I think standard points to (1), because bolded "<b>Otherwise</b>" is for `c`, and not for "`if`" statements.

So, what do you think?

annevk:
Fixes #217.

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -1742,81 +1742,70 @@ string <var>input</var>, optionally with a <a>base URL</a> <var>base</var>, opti
 
    <dt><dfn>file state</dfn>
    <dd>
-    <p>Set <var>url</var>'s <a for=url>scheme</a> to "<code>file</code>",
-    and then, switching on <a>c</a>:
+    <ol>
+     <li><p>Set <var>url</var>'s <a for=url>scheme</a> to "<code>file</code>".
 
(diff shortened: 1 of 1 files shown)


annevk:
Tests at https://github.com/w3c/web-platform-tests/issues/4700.

annevk:
I filed no bugs since browsers already did this correctly.

PR #226, opened 2017-02-02
Meta: state idempotence clearly as a goal

annevk:
Especially since browsers often violate this, which is rather bad for
security.

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -45,6 +45,8 @@ Boilerplate: omit feedback-header, omit conformance
  enhancements to make it easier to work with. Add a new <code><a interface>URL</a></code>
  object as well for URL manipulation without usage of HTML elements. (Useful
  for JavaScript worker environments.)
+
+ <li><p>Ensure the combination of parser, serializer, and API guarantee idempotence.
 </ul>
 
 <p class=note>As the editors learn more about the subject matter the goals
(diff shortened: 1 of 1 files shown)


jasnell:
hmm... +1 on including this but it may need a bit more explanation around it. Specifically, what does the idempotency guarantee guarantee in this context?

annevk:
What I want is that parse(serialize(parse(X))) "host/URL equals" parse(X). But also parse(X) followed by an API manipulation should still serialize and parse to the same thing as parse(X) followed by the API manipulation.

annevk:
Same for N API manipulations. Happy to get improved wording or formalize it somehow. Stating this formally is not my area of expertise.

annevk:
@domenic thoughts? Or should I just drop this for now?

domenic:
I think this is good and worth including, but I tend to agree it's a bit overly brief. So feel free to merge, but if you're up for expanding on it I think that would help people. Perhaps:

> For example, the result of a parse-then-serialize operation applied to any input string will then not change under further parse-then-serialize transformations. Similarly, manipulating a `URL` object using its API will always guarantee that its different properties stay synchronized.

annevk:
Thanks for the suggestion. I modified your suggested text somewhat and added it. Would appreciate another look @jasnell.

Issue #209, opened 2017-01-16
Is file:///C|/foo/bar meant to be a syntax violation?


dd8:
file:///C:/foo/bar parses ok using the state machine, but file:///C|/foo/bar flags a syntax violation in the path state because | is not a URL code point while it's scanning for the / at end of the C| path segment:

2. Otherwise, run these steps:
    If c is not a URL code point and not "%", syntax violation.

Is that intended? C| is not the normalised window drive letter form, but it's not clear from any of the narrative it's a syntax violation.

jasnell:
Yes, `C|` has historically been used (rather unfortunately) as an alternative for `C:` in some contexts. It is a syntax violation but for the sake of being forgiving, it is still accepted by the parser

dd8:
It's probably be a good idea to add a note to that effect to https://url.spec.whatwg.org/#windows-drive-letter definition (otherwise you need to step through the state machine to find this out)

I was uncertain because there are bunch of test suite files in https://github.com/w3c/web-platform-tests/tree/master/conformance-checkers and https://github.com/validator/tests showing file:///C|/foo/bar as a must-be-valid case.

I'll raise an issue in the web-platform-tests repo

domenic:
This is supposedly covered by https://url.spec.whatwg.org/#url-syntax which is pretty clear that the | form is disallowed. But I guess there's no overall note explaining that that section is meant to provide a conformance guideline.

annevk:
So duplicate of #118 then?

annevk:
I guess one thing that's novel to this issue is where those terms in the Infrastructure section should go. They're mostly there because I couldn't find a better place, but maybe they should go into the top-level section 4.

dd8:
I think one thing that could be clearer in the current spec is the high level division of URLs into: 

- conforming URLs
- non-conforming, but parsable URLs (syntax violations)
- non-parsable URLs (return failure)

Adding some syntax violations to the examples table would make this clearer. Would be good to add file: URLs to the examples table (both conforming and non-conforming) since the long history of interop problems with file: URLs suggests they're a source of much confusion, so need most clarity.

annevk:
I created some PRs that go somewhat towards those goals @dd8. Hope they help.

[...]

Issue #60, opened 2015-08-11
Replace "parse error" with "parse violation"


sideshowbarker:
The spec should replace https://url.spec.whatwg.org/#parse-error with **parse violation**.

**parse error** is an ambiguous and misleading term. See https://github.com/whatwg/url/issues/59#issuecomment-129727101 for background.

hillbrad:
Perhaps just adding "and continue" or some other language to indicate that the algorithm "falls through" such a condition would be a more important clarification.

annevk:
I'm a bit stuck on how to satisfy @hillbrad's comment since we have many instances where we require that something is done, but don't want things to end. If we make it "report a syntax error" would that make it clearer that it doesn't terminate the parse operation? Note that the definition for parse error is clear on this already.

hillbrad:
I see the difficulty here.  I wonder if there is room to introduce a
distinction between a "warning" and an "error"?  I tend to think of a
warning as something that may / should be reported but which doesn't cause
termination, whereas an error has stronger connotations of being unable to
recover or continue an algorithm.  Not all that important at the end of the
day, I should have been a more careful and complete reader.

annevk:
Yeah, I think that's why @sideshowbarker suggested "violation". It's not clear to me that "warning" is strong enough since you are actually in error, at least, to the extent we keep caring about conformance.

sideshowbarker:
I think **syntax violation** is perfect in the context of the URL spec but I think optimally it would be good to have the same term for this “parse error” stuff across all specs, including the HTML spec. But in the HTML spec, since the treebuilder part (not just the tokenizer) also contains “parse error” cases, **syntax violation** may not be accurate for in all those cases.

But I guess we can worry about what to change it to in the HTML spec if/when we ever actually get there.

annevk:
Yeah, I wonder what errors there would not be syntax violations, but I agree that if we find we need a different term we should revisit this.

Issue #59, opened 2015-08-10
"/\\example.com" is treated as an unsafe protocol-relative url, not a path, by most browsers


hillbrad:
Am I missing the place where this is described by the parsing algorithm?  Seems it should be between:

https://url.spec.whatwg.org/#no-scheme-state 

and

https://url.spec.whatwg.org/#path-or-authority-state

domenic:
Can you quantify "many" browsers and provide a repro (e.g. with `<a>` tags)?

Confirmed that the current spec (as implemented by [jsdom/whatwg-url](https://github.com/jsdom/whatwg-url)) will treat such input as an invalid URL and fail to parse it entirely.

hillbrad:
![chromecap](https://cloud.githubusercontent.com/assets/2186096/9182077/6e4fe4f2-3f5d-11e5-833d-7c03dbcfd410.png)

This is Chrome 44 on Yosemite. Same behavior for Safari and FF.

hillbrad:
Hmm.. so this is the behavior when you type such a link directly into the URL bar, but when you load from http, it treats "image.jpg" as a hostname.  So as a protocol-relative URL, ignoring the initial /.
![cap2](https://cloud.githubusercontent.com/assets/2186096/9182224/6b9d0f72-3f5e-11e5-9eae-8c082ab4393a.png)

annevk:
Yeah, this is defined by the parser. What is the problem?

domenic:
The parser currently rejects this input whereas browsers do not, it seems.

annevk:
It does? Step 7 of https://url.spec.whatwg.org/#scheme-state handles this, no?

domenic:
Oh, you're right, I forgot to input a base URL when parsing it. The spec seems OK here (and matches Chrome, although not Firefox or IE11). A good case for the test suite.

https://jsbin.com/rilorigufo/edit?html,console,output in case anyone wants to test.

hillbrad:
I don't see how you get to step 7 of scheme state for a url with no ":" in
it.

Here's my manual walkthrough of the algorithm:

base URL:  "http://foobar.com"
URL: "/\\example.jpg"
Non-relative flag is unset
State override is not given

Begin:
scheme start state, pointer at /
scheme start state choice 2: c is not ASCII alpha, state = no scheme state,
decrease pointer
increase pointer, no scheme state, pointer at /
no scheme state choice 3: base URL non-relative flag is unset, so state =
relative state, decrease pointer
[...]

PR #228, opened 2017-02-02
Explain how syntax relates to the parser for hosts and URLs

### LINKED ISSUE #118: It's not immediately clear that "URL syntax" and "URL parser" conflict
URL syntax is a model for valid URLs---basically "authoring requirements". The URL parser section allows parsing URLs which do not follow URL syntax.

An easy example is `https://////example.com`, which is disallowed because the portion after `https:` contradicts

> A scheme-relative URL
> must be "//", followed by a host, optionally followed by ":" and a port, optionally followed by a path-absolute URL. 

/cc @bagder via some Twitter confusion this morning.

annevk:
Fixes #118 and fixes part of #209.

***
[Preview](https://url.spec.whatwg.org//branch-snapshots/annevk/concept-relations/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/url/2f99502..annevk/concept-relations:8fc66a8.html)

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -76,15 +76,14 @@ number.
 
 <h3 id=writing>Writing</h3>
 
-<p>A <dfn oldids=syntax-violation>writing violation</dfn> indicates a non-fatal mismatch between
-input and writing requirements. User agents, especially conformance checkers, are encouraged to
-report them somewhere.
+<p>A <dfn oldids=syntax-violation>validation error</dfn> indicates a mismatch between input and
+valid input. User agents, especially conformance checkers, are encouraged to report them somewhere.
 
 <div class="note no-backref">
- <p>A <a>writing violation</a> does not mean that the parser terminates. Termination of a parser is
+ <p>A <a>validation error</a> does not mean that the parser terminates. Termination of a parser is
  always stated explicitly, e.g., through a return statement.
 
- <p>It is useful to signal <a>writing violations</a> as error-handling can be non-intuitive, legacy
+ <p>It is useful to signal <a>validation errors</a> as error-handling can be non-intuitive, legacy
  user agents might not implement correct error-handling, and the intent of what is written might be
  unclear to other developers.
 </div>
@@ -240,7 +239,7 @@ point <a for=/>URLs</a> from <var>A</var> can come from untrusted sources.
 
  <li><p>A <a for=/>host</a> can be seen as the in-memory representation.
 
- <li><p>A <a>valid host string</a> defines what input would not trigger a <a>writing violation</a>
+ <li><p>A <a>valid host string</a> defines what input would not trigger a <a>validation error</a>
  or failure when given to the <a>host parser</a>. I.e., input that would be considered conforming or
  valid.
 
@@ -318,7 +317,7 @@ or
  <i>UseSTD3ASCIIRules</i> set to false, <i>processing_option</i> set to
  <i>Transitional_Processing</i>, and <i>VerifyDnsLength</i> set to false.
 
- <li><p>If <var>result</var> is a failure value, <a>writing violation</a>, return failure.
+ <li><p>If <var>result</var> is a failure value, <a>validation error</a>, return failure.
 
  <li><p>Return <var>result</var>.
 </ol>
@@ -332,7 +331,7 @@ or
  <i>domain_name</i> set to <var>domain</var>,
  <i>UseSTD3ASCIIRules</i> set to false.
 
- <li><p>Signify <a>writing violations</a> for any returned errors, and then, return
+ <li><p>Signify <a>validation errors</a> for any returned errors, and then, return
  <var>result</var>.
 </ol>
 
@@ -401,8 +400,8 @@ then runs these steps:
   substeps:
 
   <ol>
-   <li><p>If <var>input</var> does not end with
-   "<code>]</code>", <a>writing violation</a>, return failure.
+   <li><p>If <var>input</var> does not end with "<code>]</code>", <a>validation error</a>, return
+   failure.
 
    <li><p>Return the result of
    <a lt="IPv6 parser">IPv6 parsing</a> <var>input</var>
@@ -423,10 +422,10 @@ then runs these steps:
  <li><p>Let <var>asciiDomain</var> be the result of running
  <a>domain to ASCII</a> on <var>domain</var>.
 
- <li><p>If <var>asciiDomain</var> is failure, <a>writing violation</a>, return failure.
+ <li><p>If <var>asciiDomain</var> is failure, <a>validation error</a>, return failure.
 
  <li><p>If <var>asciiDomain</var> contains a <a>forbidden host code point</a>,
- <a>writing violation</a>, return failure.
+ <a>validation error</a>, return failure.
 
  <li><p>Let <var>ipv4Host</var> be the result of <a lt="IPv4 parser">IPv4 parsing</a>
  <var>asciiDomain</var>.
@@ -527,16 +526,16 @@ these steps:
    <li><p>Append <var>n</var> to <var>numbers</var>.
   </ol>
 
- <li><p>If <var>writingViolationFlag</var> is set, <a>writing violation</a>.
+ <li><p>If <var>writingViolationFlag</var> is set, <a>validation error</a>.
 
- <li><p>If any item in <var>numbers</var> is greater than 255, <a>writing violation</a>.
+ <li><p>If any item in <var>numbers</var> is greater than 255, <a>validation error</a>.
 
  <li><p>If any but the last item in <var>numbers</var> is greater than 255, return
  failure.
 
  <li><p>If the last item in <var>numbers</var> is greater than or equal to
- 256<sup>(5 &minus; the number of items in <var>numbers</var>)</sup>,
- <a>writing violation</a>, return failure.
+ 256<sup>(5 &minus; the number of items in <var>numbers</var>)</sup>, <a>validation error</a>,
+ return failure.
 
  <li><p>Let <var>ipv4</var> be the last item in <var>numbers</var>.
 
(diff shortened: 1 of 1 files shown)


url.bs
```diff
@@ -226,9 +226,27 @@ point <a for=/>URLs</a> from <var>A</var> can come from untrusted sources.
...
 
-<!-- Punycode:
-     https://tools.ietf.org/html/rfc3492
-     https://mothereff.in/punycode -->
+<p>At a high level, a <a for=/>host</a>, <a>host string</a>, <a>host parser</a>, and
+<a>host serializer</a> relate as follows:
+
+<ul>
+ <li><p>The <a>host parser</a> takes an arbitrary string and returns either failure or a
+ <a for=/>host</a>. (This <a for=/>host</a> cannot be an <a>opaque host</a>, those can only be
```
domenic: comma should be semicolon; each half is a complete sentence

url.bs
```diff
@@ -226,9 +226,27 @@ point <a for=/>URLs</a> from <var>A</var> can come from untrusted sources.
...
 <h2 id="hosts-(domains-and-ip-addresses)">Hosts (domains and IP addresses)</h2>
 
-<!-- Punycode:
-     https://tools.ietf.org/html/rfc3492
-     https://mothereff.in/punycode -->
+<p>At a high level, a <a for=/>host</a>, <a>host string</a>, <a>host parser</a>, and
+<a>host serializer</a> relate as follows:
+
+<ul>
+ <li><p>The <a>host parser</a> takes an arbitrary string and returns either failure or a
```
domenic: I wonder if you want to link to infra for strings

url.bs
```diff
@@ -823,6 +842,27 @@ unified model would be, please file an issue.
...
+
+ <li><p>A <a for=/>URL</a> can be seen as the in-memory representation.
+
+ <li><p>A <a>URL string</a> defines what input would not trigger a <a>syntax violation</a> or
+ failure when given to the <a>URL parser</a>. I.e., input that would be considered conforming or
+ valid.
+
+ <li><p>The <a>URL serializer</a> takes a <a for=/>URL</a> and returns a string. (If that string
+ is then <a lt="URL parser">parsed</a>, the result will <a for=url>equal</a> the
+ <a lt="URL serializer">serialized</a> <a for=/>host</a>.)
```
domenic: copypasta "host"

url.bs
```diff
@@ -823,6 +842,27 @@ unified model would be, please file an issue.
 <!-- History behind URL as term:
      https://lists.w3.org/Archives/Public/uri/2012Oct/0080.html -->
 
+<p>At a high level, a <a for=/>URL</a>, <a>URL string</a>, <a>URL parser</a>, and
```
domenic: Reading this makes me wonder if it should be "conformant URL string" or "valid URL string" instead. A web programmer probably thinks of "a URL string" as "a string that is a URL", with an ambiguous meaning of "is" that doesn't necessarily have the nuance of conformance involved.

domenic: Similarly maybe renaming the "URL syntax" section to "Conformant URL syntax"

zcorpan: Related: https://github.com/whatwg/html/pull/2318

Since HTML uses "valid URL" (and "valid foo" in general), maybe URL spec should use "valid URL string", and HTML can use that term directly?

annevk:
Thoughts thus far:

* A sibling section to "Parsers" doesn't really work since syntax doesn't require infrastructure. A "syntax violation" is really a parser concept since only the parser calls them out.
* If we start renaming syntax to "Valid URLs" or "URL validity" we'd need a whole lot of accompanying changes. We'd also no longer match how HTML talks about this. I'm not sure that's an improvement.
* I'm not entirely convinced the confusion is as widespread as it's made to appear. There are indeed a couple of folks with a shared background who are confused, but we've had that with HTML as well and as time passed those objections passed.

So I'm not entirely sure if I want to make more drastic changes at this point.

Things I'm open to doing:

* Add a non-normative paragraph at the start of the syntax sections explaining who they are for, similar to what HTML does.

domenic:
> A sibling section to "Parsers" doesn't really work since syntax doesn't require infrastructure. A "syntax violation" is really a parser concept since only the parser calls them out.

Maybe renaming "Infrastructure" to "Shared concepts" then. I disagree that "syntax violation" is a parser concept; it's detected during the course of parsing, but it's much more interesting in the context of conformance checking, and the fact that conformance checking reuses the parser infrastructure is almost incidental. You could imagine a world where conformance checking uses a grammar instead, for example.

> I'm not entirely convinced the confusion is as widespread as it's made to appear. There are indeed a couple of folks with a shared background who are confused, but we've had that with HTML as well and as time passed those objections passed.

I don't think the confusion has passed for HTML over time. We see implementers unaware of the difference on almost a weekly basis. To me anything we can do to make it more explicit helps the situation.

Anyway, it's fine if you don't want to change much; what's in this PR is already an improvement and we shouldn't block it.

annevk:
Yeah, you're right that folks get confused by HTML too and I was wrong that I followed HTML here. HTML calls the whole setup syntax, and then distinguishes between writing and parsing. Maybe that's what URL should do. (See https://html.spec.whatwg.org/multipage/#toc-syntax.)

url.bs
```diff
@@ -74,6 +74,22 @@ DOM, Encoding, IDNA, and Web IDL Standards.
 number.
 
 
+<h3 id=writing>Writing</h3>
+
+<p>A <dfn oldids=syntax-violation>writing violation</dfn> indicates a non-fatal mismatch between
```
domenic: Eh, this turn of phrase just seems awkward... a "violation of writing"? I think it's OK for the section to be about writing URLs, but to call it a "conformance violation" or "syntax violation" still.

url.bs
```diff
@@ -74,6 +74,22 @@ DOM, Encoding, IDNA, and Web IDL Standards.
...
+input and writing requirements. User agents, especially conformance checkers are encouraged to
```
domenic: Missing comma after "conformance checkers"

domenic: Again "writing requirements" is a bit of an odd turn of phrase. It could work with some explanation, probably... Maybe "requirements for writing URLs" would be enough?

url.bs
```diff
@@ -306,16 +333,16 @@ or
...
  <i>UseSTD3ASCIIRules</i> set to false.
 
- <li><p>Signify <a>syntax violations</a> for any returned errors, and then, return
+ <li><p>Signify <a>writing violations</a> for any returned errors, and then, return
  <var>result</var>.
 </ol>
 
 
-<h3 id=host-syntax>Host syntax</h3>
+<h3 id=host-writing oldids=host-syntax>Host writing</h3>
```
domenic: "Writing hosts" or "Writing conformant hosts" maybe?

url.bs
```diff
@@ -820,8 +847,29 @@ unified model would be, please file an issue.
...
+
+ <li><p>A <a for=/>URL</a> can be seen as the in-memory representation.
+
+ <li><p>A <a>valid URL string</a> defines what input would not trigger a <a>writing violation</a> or
+ failure when given to the <a>URL parser</a>. I.e., input that would be considered conforming or
+ valid.
+
+ <li><p>The <a>URL serializer</a> takes a <a for=/>URL</a> and returns a string. (If that string
+ is then <a lt="URL parser">parsed</a>, the result will <a for=url>equal</a> the
+ <a lt="URL serializer">serialized</a> <a for=/>URL</a>.)
```
domenic: I don't think it will equal the serialized URL; the serialized URL is a string. Maybe "the URL that was serialized".

url.bs
annevk: Wouldn't work for hosts.

url.bs
annevk: All other sections lead with "Host".

domenic: I guess "Book writing" is not that much worse than "Writing books".

url.bs
```diff
@@ -223,18 +231,36 @@ point <a for=/>URLs</a> from <var>A</var> can come from untrusted sources.
...
-<!-- Punycode:
-     https://tools.ietf.org/html/rfc3492
-     https://mothereff.in/punycode -->
+<p>At a high level, a <a for=/>host</a>, <a>valid host string</a>, <a>host parser</a>, and
+<a>host serializer</a> relate as follows:
+
+<ul>
+ <li><p>The <a>host parser</a> takes an arbitrary string and returns either failure or a
+ <a for=/>host</a>. (This <a for=/>host</a> cannot be an <a>opaque host</a>; those can only be
+ returned through the <a>URL parser</a>.)
```
TRowbotham: cdbcce6 made it so that the host parser can return an opaque host, so this note is no longer true.

annevk: Excellent point, thanks!

annevk:
So we discussed terms before in #60 with @sideshowbarker and decided to rename from "parse error" then because of https://github.com/whatwg/url/issues/59#issuecomment-129727101.

There's additionally the question of whether to signify both non-fatal and fatal the same way. I think we probably should signify them the same way (since it's still a requirements mismatch), but fatal has additionally the failure return value.

"Conformance violation" would work, but I don't like that we then have both "valid" and "conformance". And "conforming" seems overall like a more complicated word.

Anyone any good ideas?

domenic:
"validity error" or "validation error"? Not sure if that crosses over into the line of "bad" that "validator" does.

annevk:
I still don't really understand the problem with "validator". "Validation error" seems reasonable. @sideshowbarker?

sideshowbarker:
> "Validation error" seems reasonable.

Yes agreed

annevk:
Thank you both! Getting somewhere.

Commit message:
```
Attempt to explain valid input better

In particular, do away with the word "syntax" as that causes lots of confusion and focus on validity instead. Also explain the relationship between the parser, serializer, model, and (valid) input.

"Syntax violation" is now known as "validation error".

Fixes #118 and fixes part of #209.
```

annevk:
I kinda addressed that issue in https://github.com/whatwg/url/pull/228#issuecomment-278617035. I think the duplication is fine and more clear, especially with the revised wording.

PR #229, opened 2017-02-02
Define URL's toJSON()

### LINKED ISSUE #137: toJSON() serialization
I think it'd be handy if all `URL` instances would serialize to the value of `href`.

annevk:
This makes it a little easier to use with JSON.stringify().

Fixes #137.

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -2483,6 +2483,8 @@ interface URL {
            attribute USVString search;
   [SameObject] readonly attribute URLSearchParams searchParams;
            attribute USVString hash;
+
+  USVString toJSON();
 };
 </pre>
 
@@ -2579,8 +2581,9 @@ url.pathname // "/%F0%9F%8F%B3%EF%B8%8F%E2%80%8D%F0%9F%8C%88"</pre>
 
 <h3 id=urlutils-members>{{URL}} members</h3>
 
-<p>The <dfn attribute for=URL><code>href</code></dfn> attribute's getter must return the
-<a lt='URL serializer'>serialization</a> of <a>context object</a>'s <a for=URL>url</a>.
+<p>The <dfn attribute for=URL><code>href</code></dfn> attribute's getter and the
+<dfn method for=URL>toJSON()</dfn> method, when invoked, must return the
+<a lt="URL serializer">serialization</a> of <a>context object</a>'s <a for=URL>url</a>.
 
 <p>The <code><a attribute for=URL>href</a></code> attribute's setter must run these steps:
 


url.bs
```diff
@@ -2483,6 +2483,8 @@ interface URL {
            attribute USVString search;
   [SameObject] readonly attribute URLSearchParams searchParams;
            attribute USVString hash;
+
+  USVString toJSON();
```
TimothyGu: Even though I'm not a fan of the complicated serializer syntax, I do like the conciseness of just writing

```idl
serializer;
```

which is more parallel to the existing `iterable<USVString, USVString>;` for `URLSearchParams`.

annevk: It would be, but this is the way IDL is going for `toJSON()`. I checked with @bzbarsky and @tobie.

bzbarsky: Is the question one of having `serializer;` basically be a thing that produces a then-custom-defined toJSON, kinda like `stringifier;`?

The problem is that unlike `stringifier;`, where the return value is clear, the return value of toJSON could be all sorts of things.  We could define `serializer;` to induce the equivalent of `any toJSON();` but that's more annoying to work with than being able to usefully declare your return type...

TimothyGu: Indeed, I haven't thought of that. How about `serializer<USVString>;`? It also has the advantage of being more language-independent, while there are precedents for `<>` meaning "produces" with `iterable`.

domenic: Why? We should be moving away from this baroque, hard-to-understand, "language independent" syntax in favor of simple, do-what-it-says JavaScript-derived syntax.

The only reason iterable is special is because it's a macro that expands to ~4 JS methods. For a simple stand-in for a JS method, we should not create an extra layer of abstraction that requires looking up the Web IDL spec to understand what it does.

esperecyan: If “URL” interface will be implemented on PHP, should this method be implemented by the name of “jsonSerialize()” instead of “toJSON()”?
https://secure.php.net/manual/class.jsonserializable.php

domenic: If you're porting the JavaScript URL class to PHP, you can do whatever you like, as there's no spec for that. It's up to you what you think is most intuitive for your developers (e.g. probably PHP has different naming conventions than JavaScript in many areas, not just JSON serialization).

esperecyan: I get it. Thank you for your advice.

annevk: I'm going to assume @TimothyGu is okay with the PR as-is given five days have passed. Hope that's okay. (Still I'll wait a little longer with landing since the test required a further change.)

annevk:
Tests: https://github.com/w3c/web-platform-tests/pull/4702.

url.bs
TimothyGu: @annevk, yeah I'm okay with the PR in general. The bit about IDL notation is minor and does not affect the semantics anyway.

PR #242, opened 2017-02-10
Define which URLs are valid in the parser examples

### LINKED ISSUE #209: Is file:///C|/foo/bar meant to be a syntax violation?

annevk:
Also make sure the base URLs are actually represented in serialized
form by adding trailing slashes.

Fixes #209.

***
[Preview](https://url.spec.whatwg.org//branch-snapshots/annevk/validity-example/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/url/50cb9ab..annevk/validity-example:4000ed7.html)

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -894,7 +894,7 @@ unified model would be, please file an issue.
   <tr>
    <td><code>https:example.org</code>
    <td>
-   <td>✅
+   <td>❌
    <td><code>https://example.org/</code>
   <tr>
    <td><code>https://////example.com///</code>
--- a/url.bs
+++ b/url.bs
@@ -942,6 +942,26 @@ unified model would be, please file an issue.
    <td>✅
    <td><code>file:///</code>
   <tr>
+   <td><code>https://user:password@example.org/</code>
+   <td>
+   <td>❌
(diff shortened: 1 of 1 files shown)


zcorpan:
I think valid would be better as its own column.

annevk:
Yeah, that would look better probably, this is a mess. On the other hand, the main reason I did this way was because it might be confusing what it's applicable to.

annevk:
Gave the third column suggestion a go. Probably clear enough.

zcorpan:
Good ones to add from https://github.com/whatwg/url/issues/118

```
https://user:password@example.org/
https://ex ample.org/
https://example.org/foo bar
https://EXAMPLE.com/../x
```

PR #243, opened 2017-02-10
Editorial: use specific names for encode sets

### LINKED ISSUE #201: Avoid using "simply"

annevk:
“Simple encode set” is effectively “C0 control encode set” so name it
that. And “default encode set” is only used for paths, so name it “path
encode set”.

Fixes #201.

***
[Preview](https://url.spec.whatwg.org//branch-snapshots/annevk/encode-set-names/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/url/cdbe5dc..annevk/encode-set-names:c60ff34.html)

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -164,10 +164,11 @@ contains bytes that are not <a>ASCII bytes</a> might be insecure and is not reco
 
      note that query and application/x-www-form-urlencoded use their own
      local sets -->
-<p>The <dfn>C0 control encode set</dfn> are <a>C0 controls</a> and all code points greater than
-U+007E.
+<p>The <dfn oldids=simple-encode-set>C0 control percent-encode set</dfn> are <a>C0 controls</a> and
+all code points greater than U+007E.
 
-<p>The <dfn>path encode set</dfn> is the <a>C0 control encode set</a> and code points
+<p>The <dfn oldids=default-encode-set>path percent-encode set</dfn> is the
+<a>C0 control percent-encode set</a> and code points
 U+0020,
 '<code>"</code>', <!-- 0x22 -->
 "<code>#</code>", <!-- 0x23 -->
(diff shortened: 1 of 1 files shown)


url.bs
```diff
@@ -164,11 +164,11 @@ contains bytes that are not <a>ASCII bytes</a> might be insecure and is not reco
 
      note that query and application/x-www-form-urlencoded use their own
      local sets -->
-<p>The <dfn>simple encode set</dfn> are <a>C0 controls</a> and all code points greater
-than U+007E.
+<p>The <dfn>C0 control encode set</dfn> are <a>C0 controls</a> and all code points greater than
```
domenic: Add oldids=""

domenic: Would a better name be around its usage, which appears to have something to do with percent encoding? Just an idea.

annevk: You want them renamed to percent-encode sets?

annevk: So this would become C0 control percent-encode set?

domenic: No, I didn't remember both were used for percent encoding. Is there a succinct description of where this is used, like "path" or "userinfo"? "C0 control", is what it *is*, not how it's used.

annevk: All the encode sets are used for percent-encoding.

This one is used by the opaque-host parser, the cannot-be-a-base-URL path state, and the fragment state.

domenic: It sounds like there's no good unifying name for those concepts, so I guess naming it after C0 controls is fine. *Shrug*.

annevk: Should I name them percent-encode sets though? I agree that there's no good name for this one other than describing what's in it, which seems a little better than "default" or "simple".

domenic: Hmm yeah, I guess while you're here, that might be a nice change as well.

PR #244, opened 2017-02-10
Paths need to be copied from base URLs

### LINKED ISSUE #231: Setting path to base's path seems wrong
When reviewing https://github.com/whatwg/url/pull/225 I noticed this, and I imagine it's widespread:

> Set url’s host to base’s host, **url’s path to base’s path**, and url’s query to base’s query.

A strict reading of this is that it sets url's path to the same list as base's path, so any future modifications to url's path get reflected in the base URL. That seems pretty bad. It should instead be a copy, right?

annevk:
Otherwise things get a little weird.

Fixes #231.

***
[Preview](https://url.spec.whatwg.org//branch-snapshots/annevk/copy-paths/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/url/50cb9ab..annevk/copy-paths:fd8895f.html)

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -2086,7 +2086,7 @@ string <var>input</var>, optionally with a <a>base URL</a> <var>base</var>, opti
        not start with two <a>ASCII hex digits</a>, <a>validation error</a>.
 
        <li><p>If <a>c</a> is not <a>EOF code point</a>, <a>UTF-8 percent encode</a> <a>c</a> using
-       the <a>simple encode set</a>, and <a for=list>append</a> the result to <var>url</var>'s
+       the <a>simple encode set</a>, and append the result to <var>url</var>'s
        <a for=url>path</a>[0].
       </ol>
     </ol>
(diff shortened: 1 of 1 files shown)


url.bs
```diff
@@ -2086,8 +2086,8 @@ string <var>input</var>, optionally with a <a>base URL</a> <var>base</var>, opti
        not start with two <a>ASCII hex digits</a>, <a>validation error</a>.
 
        <li><p>If <a>c</a> is not <a>EOF code point</a>, <a>UTF-8 percent encode</a> <a>c</a> using
-       the <a>simple encode set</a>, and <a for=list>append</a> the result to the first string in
-       <var>url</var>'s <a for=url>path</a>.
+       the <a>simple encode set</a>, and <a for=list>append</a> the result to <var>url</var>'s
```
domenic: This append should not be for=list, it's actually string append, right?

annevk: You're right.

PR #251, opened 2017-02-15
Meta: make .pr-preview.json valid json

tobie:


annevk:
Our build script should start linting JSON just like we validate HTML.

PR #260, opened 2017-02-24
Use empty host for file URL's

### LINKED ISSUE #258: Empty or null host when scheme is "file:"?
In the ["file host state"](https://url.spec.whatwg.org/#file-host-state) there are steps where <var>host</var> is set to the empty string, but in the ["path state"](https://url.spec.whatwg.org/#path-state) at some step <var>host</var> is tested for null (and also set to null).

Because of this difference parser will report [validation error](https://url.spec.whatwg.org/#validation-error) for such valid URL:
`file:///C:/path`
In the ["file host state"](https://url.spec.whatwg.org/#file-host-state) <var>host</var> is set to empty string, but in the ["path state"](https://url.spec.whatwg.org/#path-state): "If url’s host is non-null, validation error."

So I think, <var>host</var> must be set to (and tested for) the same value in both cases; Maybe this value must be null?

rmisev:
Summary of changes:
1. Fixes #258
2. Introduces <b>empty host</b> term for file and non-special URL's
3. Changes definition: <b>opaque host</b> is non-empty ASCII string

***
[Preview](https://s3.amazonaws.com/pr-preview/rmisev/url/empty-host-for-file.html) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/url/4bf85a0...rmisev:3893222.html)

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -385,7 +385,7 @@ up to three <a>ASCII digits</a> per sequence, each representing a decimal number
 
      XXX should we define the format inline instead just like STD 66? -->
 
-<p>An <dfn export>valid opaque-host string</dfn> must be one or more <a>URL units</a> or:
+<p>A <dfn export>valid opaque-host string</dfn> must be one or more <a>URL units</a> or:
 "<code>[</code>", followed by a <a>valid IPv6-address string</a>, followed by "<code>]</code>".
 
 <p class="note no-backref">This is not part of the definition of <a>valid host string</a> as it
@@ -770,8 +770,8 @@ no purpose other than being a location the algorithm can jump to.
  <a>IPv6 serializer</a> on <var>host</var>,
  followed by "<code>]</code>".
 
- <li><p>Otherwise, <var>host</var> is a <a>domain</a>, an <a>opaque host</a>, or an <a>empty
- host</a>, return <var>host</var>.
+ <li><p>Otherwise, <var>host</var> is a <a>domain</a>, <a>opaque host</a>, or <a>empty host</a>,
+ return <var>host</var>.
 </ol>
 
 The <dfn id=concept-ipv4-serializer>IPv4 serializer</dfn> takes an
@@ -1175,8 +1175,8 @@ switching on <a>base URL</a>'s <a for=url>scheme</a>:
  <dd><p>a <a>path-relative-scheme-less-URL string</a>
  <dt>"<code>file</code>"
  <dd><p>a <a>scheme-relative-file-URL string</a>
- <dd><p>a <a>path-absolute-URL string</a> if <a>base URL</a>'s <a for=url>host</a> is an <a>empty
- host</a>
+ <dd><p>a <a>path-absolute-URL string</a> if <a>base URL</a>'s <a for=url>host</a> is an
+ <a>empty host</a>
  <dd><p>a <a>path-absolute-non-Windows-file-URL string</a> if <a>base URL</a>'s <a for=url>host</a>
  is not an <a>empty host</a>
  <dd><p>a <a>path-relative-scheme-less-URL string</a>
@@ -2073,8 +2073,8 @@ string <var>input</var>, optionally with a <a>base URL</a> <var>base</var>, opti
            <li><p>If <var>url</var>'s <a for=url>host</a> is not the empty string,
            <a>validation error</a>.
 
-           <li><p>Set <var>url</var>'s <a for=url>host</a> to the empty string and replace the second
-           code point in <var>buffer</var> with "<code>:</code>".
+           <li><p>Set <var>url</var>'s <a for=url>host</a> to the empty string and replace the
+           second code point in <var>buffer</var> with "<code>:</code>".
           </ol>
 
           <p class=note>This is a (platform-independent) Windows drive letter quirk.
--- a/url.bs
+++ b/url.bs
@@ -283,9 +283,6 @@ eight <dfn id=concept-ipv6-piece lt='IPv6 piece'>16-bit pieces</dfn>.
 <p>An <dfn export>opaque host</dfn> is a non-empty <a>ASCII string</a> holding data that can be used
 for further processing.
 
-<p class="note no-backref">An <a>opaque host</a> is only used by <a lt="is special">non-special</a>
-<a for=/>URLs</a>.
-
 <p>An <dfn export>empty host</dfn> is the empty string.
 
 
@@ -1010,17 +1007,40 @@ It is initially the empty string.
  <a for=url>host</a> combinations.
 
  <table>
-  <tr><th rowspan="2"><a for=/>URL</a>'s <a for=url>scheme</a>
-      <th colspan="6"><a for=/>URL</a>'s <a for=url>host</a>
-  <tr><th><a>domain</a>
-      <th><a>IPv4 address</a>
-      <th><a>IPv6 address</a>
-      <th><a>opaque host</a>
-      <th><a>empty host</a>
-      <th>null
-  <tr><td>non-file <a lt="special scheme">special</a><td>✅<td>✅<td>✅<td><td><td>
-  <tr><td>"<code>file</code>"<td>✅<td>✅<td>✅<td><td>✅<td>✅
-  <tr><td><a lt="special scheme">non-special</a><td><td><td>✅<td>✅<td>✅<td>✅
+  <tr>
+   <th rowspan=2><a for=url>scheme</a>
+   <th colspan=6><a for=url>host</a>
+  <tr>
+   <th><a>domain</a>
+   <th><a>IPv4 address</a>
+   <th><a>IPv6 address</a>
+   <th><a>opaque host</a>
+   <th><a>empty host</a>
+   <th>null
+  <tr>
+   <td>non-"<code>file</code>" <a lt="special scheme">special</a>
+   <td>✅
+   <td>✅
+   <td>✅
+   <td>❌
+   <td>❌
+   <td>❌
+  <tr>
+   <td>"<code>file</code>"
+   <td>✅
+   <td>✅
+   <td>✅
+   <td>❌
+   <td>✅
+   <td>✅
+  <tr>
+   <td><a lt="special scheme">non-special</a>
+   <td>❌
+   <td>❌
+   <td>✅
+   <td>✅
+   <td>✅
+   <td>✅
  </table>
 </div>
 
--- a/url.bs
+++ b/url.bs
@@ -280,7 +280,7 @@ eight <dfn id=concept-ipv6-piece lt='IPv6 piece'>16-bit pieces</dfn>.
 <p class="note">Support for <code>&lt;zone_id></code> is
 <a href="https://www.w3.org/Bugs/Public/show_bug.cgi?id=27234#c2">intentionally omitted</a>.
 
-<p>An <dfn export>opaque host</dfn> is an <a>ASCII string</a> holding data that can be used for
+<p>An <dfn export>opaque host</dfn> is a non-empty <a>ASCII string</a> holding data that can be used for
 further processing.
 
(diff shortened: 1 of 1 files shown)


url.bs
```diff
@@ -1172,9 +1174,9 @@ switching on <a>base URL</a>'s <a for=url>scheme</a>:
  <dd><p>a <a>path-relative-scheme-less-URL string</a>
  <dt>"<code>file</code>"
  <dd><p>a <a>scheme-relative-file-URL string</a>
- <dd><p>a <a>path-absolute-URL string</a> if <a>base URL</a>'s <a for=url>host</a> is null
+ <dd><p>a <a>path-absolute-URL string</a> if <a>base URL</a>'s <a for=url>host</a> is either null or the empty string
  <dd><p>a <a>path-absolute-non-Windows-file-URL string</a> if <a>base URL</a>'s <a for=url>host</a>
- is non-null
+ is non-empty string
```
annevk: It's not always a string though. It could be an IPv4 address.

url.bs
```diff
@@ -254,7 +254,7 @@ point <a for=/>URLs</a> from <var>A</var> can come from untrusted sources.
 <h3 id=host-representation>Host representation</h3>
 
 <p>A <dfn export id=concept-host>host</dfn> is a <a>domain</a>, an
-<a>IPv4 address</a>, an <a>IPv6 address</a>, or an <a>opaque host</a>. Typically a <a for=/>host</a>
+<a>IPv4 address</a>, an <a>IPv6 address</a>, an <a>opaque host</a>, or an <a>empty host</a>. Typically a <a for=/>host</a>
```
annevk: This doesn't meet the 100 column wrapping requirements.

url.bs
```diff
@@ -280,12 +280,14 @@ eight <dfn id=concept-ipv6-piece lt='IPv6 piece'>16-bit pieces</dfn>.
 <p class="note">Support for <code>&lt;zone_id></code> is
 <a href="https://www.w3.org/Bugs/Public/show_bug.cgi?id=27234#c2">intentionally omitted</a>.
 
-<p>An <dfn export>opaque host</dfn> is an <a>ASCII string</a> holding data that can be used for
+<p>An <dfn export>opaque host</dfn> is a non-empty <a>ASCII string</a> holding data that can be used for
```
annevk: Same here, I think.

url.bs
```diff
@@ -768,7 +770,7 @@ no purpose other than being a location the algorithm can jump to.
  <a>IPv6 serializer</a> on <var>host</var>,
  followed by "<code>]</code>".
 
- <li><p>Otherwise, <var>host</var> is a <a>domain</a> or <a>opaque host</a>, return <var>host</var>.
+ <li><p>Otherwise, <var>host</var> is a <a>domain</a>, an <a>opaque host</a>, or an <a>empty host</a>, return <var>host</var>.
```
annevk: And here.

url.bs
```diff
@@ -1172,9 +1174,9 @@ switching on <a>base URL</a>'s <a for=url>scheme</a>:
  <dd><p>a <a>path-relative-scheme-less-URL string</a>
  <dt>"<code>file</code>"
  <dd><p>a <a>scheme-relative-file-URL string</a>
- <dd><p>a <a>path-absolute-URL string</a> if <a>base URL</a>'s <a for=url>host</a> is null
+ <dd><p>a <a>path-absolute-URL string</a> if <a>base URL</a>'s <a for=url>host</a> is either null or the empty string
```
annevk: This should say "an empty host" I think. A base URL is a URL record after all. But also, I thought the idea was that for file URLs the host could never be null so then we should leave that out too.

url.bs
annevk: This should probably just say not an empty host.

url.bs
```diff
@@ -1199,7 +1201,7 @@ optionally followed by a <a>path-absolute-URL string</a>.
 <a>path-absolute-URL string</a>.
 
 <p>An <dfn export>opaque-host-and-port string</dfn> must be either an empty
-<a>valid opaque-host string</a> or: a non-empty <a>valid opaque-host string</a>, optionally followed
+string or: a <a>valid opaque-host string</a>, optionally followed
```
annevk: It's "the empty string" (and an empty host).

url.bs
```diff
@@ -2066,10 +2068,10 @@ string <var>input</var>, optionally with a <a>base URL</a> <var>base</var>, opti
           <a>Windows drive letter</a>, then:
 
           <ol>
-           <li><p>If <var>url</var>'s <a for=url>host</a> is non-null,
+           <li><p>If <var>url</var>'s <a for=url>host</a> is not the empty string,
```
annevk: Here again we're talking about a URL record so we should say "not an empty host". Same below.

url.bs
rmisev: `file:/path` still parses to null host (with validation error), so maybe it's safe to leave null?

url.bs
rmisev: Both an empty host and an opaque host are strings by definition. So maybe it's fine to set host to the empty string in the parser? For example the opaque host parser returns string, but not an opaque host record.

annevk: Fair, I guess some level of implicit casting is probably okay.

url.bs
```diff
@@ -383,7 +385,7 @@ up to three <a>ASCII digits</a> per sequence, each representing a decimal number
 
      XXX should we define the format inline instead just like STD 66? -->
 
-<p>An <dfn export>valid opaque-host string</dfn> must be zero or more <a>URL units</a> or:
+<p>An <dfn export>valid opaque-host string</dfn> must be one or more <a>URL units</a> or:
```
annevk: Can you change "An" to "A" here while we're changing this? I can also do it before merging.

url.bs
```diff
@@ -768,7 +770,8 @@ no purpose other than being a location the algorithm can jump to.
  <a>IPv6 serializer</a> on <var>host</var>,
  followed by "<code>]</code>".
 
- <li><p>Otherwise, <var>host</var> is a <a>domain</a> or <a>opaque host</a>, return <var>host</var>.
+ <li><p>Otherwise, <var>host</var> is a <a>domain</a>, an <a>opaque host</a>, or an <a>empty
+ host</a>, return <var>host</var>.
```
annevk: I think it reads more naturally if you leave out the "an". So "A domain, opaque host, or empty host".

url.bs
```diff
@@ -1172,9 +1175,10 @@ switching on <a>base URL</a>'s <a for=url>scheme</a>:
  <dd><p>a <a>path-relative-scheme-less-URL string</a>
  <dt>"<code>file</code>"
  <dd><p>a <a>scheme-relative-file-URL string</a>
- <dd><p>a <a>path-absolute-URL string</a> if <a>base URL</a>'s <a for=url>host</a> is null
+ <dd><p>a <a>path-absolute-URL string</a> if <a>base URL</a>'s <a for=url>host</a> is an <a>empty
+ host</a>
```
annevk: No newlines inside phrasing/inline elements.

url.bs
```diff
@@ -2066,10 +2070,10 @@ string <var>input</var>, optionally with a <a>base URL</a> <var>base</var>, opti
           <a>Windows drive letter</a>, then:
 
           <ol>
-           <li><p>If <var>url</var>'s <a for=url>host</a> is non-null,
+           <li><p>If <var>url</var>'s <a for=url>host</a> is not the empty string,
            <a>validation error</a>.
 
-           <li><p>Set <var>url</var>'s <a for=url>host</a> to null and replace the second
+           <li><p>Set <var>url</var>'s <a for=url>host</a> to the empty string and replace the second
```
annevk: This is >100 columns. "second" needs to go to the next line.

rmisev:
Maybe explanatory text is good idea, I don't know. But I would like to note, what file URL still can have null host (two invalid URL's cases: `file:path` and `file:/path`).

Also I think about such note below the <b>empty host</b> definition: "An empty host is used by non-special and file URLs."

annevk:
I guess there's no real point trying to make those invalid cases consistent. And you're right about where the note should go. Although maybe we should have it in both places eventually, once we're certain this design can be widely implemented (i.e., once two browsers ship it).

rmisev:
I think the such table is clearer than a long explanatory text https://github.com/whatwg/url/pull/260#pullrequestreview-23976295, https://github.com/whatwg/url/pull/260#issuecomment-282751627:

| Scheme           | domain | IPv4 address | IPv6 address | opaque host | empty host | null |
|------------------|--------|--------------|--------------|-------------|------------|------|
| non-file special |    +   |      +       |      +       |             |            |      |
| "file"           |    +   |      +       |      +       |             |     +      |   +  |
| non-special      |        |              |      +       |      +      |     +      |   +  |

Then maybe no need for additional notes. What do you think?

annevk:
And have that below the definition of URL's host? I think that's reasonable.

url.bs
```diff
@@ -280,12 +280,14 @@ eight <dfn id=concept-ipv6-piece lt='IPv6 piece'>16-bit pieces</dfn>.
 <p class="note">Support for <code>&lt;zone_id></code> is
 <a href="https://www.w3.org/Bugs/Public/show_bug.cgi?id=27234#c2">intentionally omitted</a>.
 
-<p>An <dfn export>opaque host</dfn> is an <a>ASCII string</a> holding data that can be used for
-further processing.
+<p>An <dfn export>opaque host</dfn> is a non-empty <a>ASCII string</a> holding data that can be used
+for further processing.
 
 <p class="note no-backref">An <a>opaque host</a> is only used by <a lt="is special">non-special</a>
 <a for=/>URLs</a>.
```
annevk: I think either we should have a note like this for empty hosts too or we should drop this one, no?

url.bs
```diff
@@ -1002,6 +1005,25 @@ It is initially the empty string.
...
+ <table>
+  <tr><th rowspan="2"><a for=/>URL</a>'s <a for=url>scheme</a>
+      <th colspan="6"><a for=/>URL</a>'s <a for=url>host</a>
+  <tr><th><a>domain</a>
+      <th><a>IPv4 address</a>
+      <th><a>IPv6 address</a>
+      <th><a>opaque host</a>
+      <th><a>empty host</a>
+      <th>null
+  <tr><td>non-file <a lt="special scheme">special</a><td>✅<td>✅<td>✅<td><td><td>
```
annevk: `non-"<code>file</code>"`

annevk:
I ended up fixing the nits, let me know what you think.

rmisev:
I think your changes are OK.
   
Yet I have changed the "Windows drive letter quirk" in the path state, to be more logical (no validation error if host null) and consistent (null host is left the null):

For example, now: `file:/C:/p` --> `host === null` as for `file:/a/p` also `host === null`, but `file://example.net/C:/` --> `host === ""`

Now LGTM.

annevk:
Looks okay, blocked on https://github.com/tabatkins/bikeshed/issues/941 now.

domenic:
To confirm, this only impacts validation, not parsing/serializing?

annevk:
Yeah, this should not have changed either of those.

rmisev:
Yes, fixes validation bug. Parsing: `file://host/C:/` and `file:///C:/` now parses to "" host, previously to null. Serializing for file URL is the same host is null or "" - so no impact.

domenic:
Note that there is no test coverage for the case introduced here: in

```js
        if (this.url.host !== "" && this.url.host !== null) {
          this.parseError = true;
          this.url.host = "";
        }
```

the condition never evaluates to true so those two lines are never reached. Maybe a coverage gap, maybe a spec bug.

rmisev:
@domenic test case, which covers [this code](https://github.com/whatwg/url/pull/260#issuecomment-286502916) is in previous comment: https://github.com/whatwg/url/pull/260#issuecomment-284814380:
`file://host/C:/`

It looks like WPT lacks this. I can prepare PR.

PR #268, opened 2017-03-09
Clarify IPv6 serializer

### LINKED ISSUE #266: IPv6 serializer unclear
I'm having a hard time making sense of this:
> If compress pointer points to piece, append "::" to output if piece is address’s first piece and append ":" otherwise, and then run these substeps again with all subsequent pieces in address’s pieces that are 0 skipped or go the next step in the overall set of steps if that leaves no pieces.

annevk:
Fixes #266.

***
[Preview](https://url.spec.whatwg.org/branch-snapshots/annevk/ipv6-serializer/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/url/8386c7a...4af559d.html)

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -8,7 +8,7 @@ No Editor: true
 Abstract: The URL Standard defines URLs, domains, IP addresses, the <code title>application/x-www-form-urlencoded</code> format, and their API.
 Logo: https://resources.whatwg.org/logo-url.svg
 Boilerplate: omit feedback-header, omit conformance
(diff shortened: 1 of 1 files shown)


domenic:
However you may find whatwg-url's former way of doing things to be simpler (no ignore0 variable):

```js
  for (let i = 0; i < address.length; ++i) {
    if (compressPtr === i) {
      if (i === 0) {
        output += "::";
      } else {
        output += ":";
      }

      i += seqResult.len - 1;
      continue;
    }

    output += address[i].toString(16);
    if (i !== address.length - 1) {
      output += ":";
    }
  }
```

annevk:
What is seqResult.len?

domenic:
It looks like it's the length of the longest zero sequence.

annevk:
If it's immaterial to you I'd like to go with the patch. I don't really see a good translation of your code to English. We don't really have good for loop infrastructure.

domenic:
Yeah, no problem for me.

PR #269, opened 2017-03-10
Restrict setting protocol to "file"

### LINKED ISSUE #259: username/password/port should be removed when "file" is added as scheme
According to [this section](https://url.spec.whatwg.org/#url-miscellaneous) in the spec, the file scheme can't have `username/password/port`. I think they also should be removed automatically when `file:` is added as the scheme. Currently, they are left in the `href` even they cannot be overwritten.

e.g.
```js
var url = new URL('http://user:pass@foo.bar.com:21/aaa/zzz?l=24#test');
url.protocol = 'file:';
// url.href:
// + expected: file://foo.bar.com/aaa/zzz?l=24#test
// + actual: "file://user:pass%40foo.bar.com:21/aaa/zzz?l=24#test"
```

annevk:
As file URLs cannot have username/password/port we don’t want to allow
changing the scheme of a URL that contains one or more of those
components.

Fixes #259.

***
[Preview](https://url.spec.whatwg.org/branch-snapshots/annevk/protocol-setter/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/url/5807b28...a8ed8d2.html)

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -1490,6 +1490,9 @@ string <var>input</var>, optionally with a <a>base URL</a> <var>base</var>, opti
         <p>If <var>state override</var> is given, run these subsubsteps:
 
         <ol>
+         <li><p>If <var>url</var> <a>includes credentials</a> or has a non-null <a for=url>port</a>,
+         and <var>buffer</var> is "<code>file</code>", then return.
+
          <li><p>If <var>url</var>'s <a for=url>scheme</a> is a <a>special scheme</a> and
          <var>buffer</var> is not, then return.
 
--- a/url.bs
+++ b/url.bs
@@ -1490,14 +1490,17 @@ string <var>input</var>, optionally with a <a>base URL</a> <var>base</var>, opti
         <p>If <var>state override</var> is given, run these subsubsteps:
 
         <ol>
-         <li><p>If <var>url</var> <a>includes credentials</a> or has a non-null <a for=url>port</a>,
-         and <var>buffer</var> is "<code>file</code>", then return.
-
          <li><p>If <var>url</var>'s <a for=url>scheme</a> is a <a>special scheme</a> and
          <var>buffer</var> is not, then return.
 
          <li><p>If <var>url</var>'s <a for=url>scheme</a> is not a <a>special scheme</a> and
          <var>buffer</var> is, then return.
+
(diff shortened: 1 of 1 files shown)


watilde:
This looks good to me to fix the issue, but I'm worried about the change seems a new behaviour that browsers don't have yet.

annevk:
Tests at https://github.com/w3c/web-platform-tests/pull/5112.

Yeah, we'll need to file browser bugs. What browsers currently do doesn't make any sense, so hopefully this can still change.

url.bs
```diff
@@ -1490,14 +1490,17 @@ string <var>input</var>, optionally with a <a>base URL</a> <var>base</var>, opti
...
          <var>buffer</var> is not, then return.
 
          <li><p>If <var>url</var>'s <a for=url>scheme</a> is not a <a>special scheme</a> and
          <var>buffer</var> is, then return.
+
+         <li><p>If <var>url</var> <a>includes credentials</a> or has a non-null <a for=url>port</a>,
+         and <var>buffer</var> is "<code>file</code>", then return.
+
+         <li><p>If <var>url</var>'s <a for=url>scheme</a> is "<code>file</code>" and its
+         <a for=url>host</a> is an <a>empty host</a> or null, then return.
```
rmisev: I think the last item must be: 
<p>4. If <var>url</var>'s <a for=url>host</a> is an <a>empty host</a> or null,
and <var>buffer</var> is a <a>special scheme</a>, then return.

annevk: It's only "file" that can come here and hit this. Other non-special schemes cannot be set to special schemes due to the earlier steps.

rmisev: Yes you are right.

annevk: I guess I could remove the check for scheme being "file" though, since only file can have an empty/null host.

annevk: No, that would be bad, since that would also block non-special URLs without host from changing schemes. Oops.

annevk:
Bugs filed:
* https://bugs.chromium.org/p/chromium/issues/detail?id=701685
* https://bugs.webkit.org/show_bug.cgi?id=169667
* https://bugzilla.mozilla.org/show_bug.cgi?id=1347459
* https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/11280885/

PR #276, opened 2017-03-20
Move some terminology to the Fetch Standard

### LINKED ISSUE #241: All the other network schemes
A quote from the text:

> A network scheme is a scheme that is "ftp" or an HTTP(S) scheme.

but even this spec itself refers to "gopher" just lines above. Is gopher not a network scheme? How about IMAP, TELNET or any of the large amount of other clearly networked schemes that are standardized?

annevk:
This wasn’t used by the URL Standard directly so might as well go to
Fetch.

Change to Fetch: https://github.com/whatwg/fetch/issues/512.

Fixes #241.

***
[Preview](https://url.spec.whatwg.org/branch-snapshots/annevk/url-terminology/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/url/7286569...2d8b4d7.html)

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -1112,29 +1112,6 @@ corresponding <a for=url>port</a> and is listed in the second column on the same
 <p>A <a for=/>URL</a> <dfn export>is special</dfn> if its <a for=url>scheme</a> is a
 <a>special scheme</a>.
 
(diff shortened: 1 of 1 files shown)


annevk:
This can land a day after the corresponding change to Fetch lands.

annevk:
Maybe state explicitly which terms moved for easier commit message hunting.

PR #234, opened 2017-02-07
Repeated slashes should be ignored in file scheme

watilde:
Fixes https://github.com/whatwg/url/issues/232.

***
[Preview](https://s3.amazonaws.com/pr-preview/watilde/url/feature/check-file-slashes.html) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/url/6688baa..watilde:feature/check-file-slashes:b46bf40.html)

PR #278, opened 2017-03-22
URL: trim leading slashes of file URL paths

### LINKED ISSUE #232: Needs to ignore repeated slashes in file URL
Hello!

Since the file state doesn't have a way to jump to the special authority slashes state, once the code enters into the file state, repeated slashes in the beginning of a file URL(e.g. `file://///foo/bar` should be `file:///foo/bar`) can't be ignored.

To cover the case, it could be done by the following steps:

1. In scheme state, stop jumping to file state directly before checking special authority slashes state.
2. In special authority ignore slashes state, add file state check.
3. In scheme state, add file state check before jumping to special relative or authority state.

Resources:
+ In the bs file, it can be like this: https://github.com/watilde/url/commit/b5871518a54ad7331de3b457dcb6a730f5d00234
+ Here is an example impl: https://github.com/watilde/node/commit/247f960f7125c1dfd24599a3800117d19e4ccb55

What do you think?

annevk:
Tests: https://github.com/w3c/web-platform-tests/pull/5195.

Closes #234 by superseding it. Fixes #232.

***
[Preview](https://url.spec.whatwg.org/branch-snapshots/annevk/file-url-path-slashes/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/url/91cb2aa...91bcea4.html)

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -1909,14 +1909,22 @@ string <var>input</var>, optionally with a <a>base URL</a> <var>base</var>, opti
 
       <ol>
        <li>
-        <p>If <var>base</var> is non-null, <var>base</var>'s <a for=url>scheme</a> is
-        "<code>file</code>", and <var>base</var>'s <a for=url>path</a>[0] is a
-        <a>normalized Windows drive letter</a>, <a for=list>append</a> <var>base</var>'s
-        <a for=url>path</a>[0] to <var>url</var>'s <a for=url>path</a>.
+        <p>If <var>base</var> is non-null and <var>base</var>'s <a for=url>scheme</a> is
+        "<code>file</code>", then:
 
-        <p class=note>This is a (platform-independent) Windows drive letter quirk. Both
-        <var>url</var>'s and <var>base</var>'s <a for=url>host</a> are null under
-        these conditions and therefore not copied.
+        <ol>
+         <li>
+          <p>If <var>base</var>'s <a for=url>path</a>[0] is a
+          <a>normalized Windows drive letter</a>, then <a for=list>append</a> <var>base</var>'s
+          <a for=url>path</a>[0] to <var>url</var>'s <a for=url>path</a>.
+
+          <p class=note>This is a (platform-independent) Windows drive letter quirk. Both
+          <var>url</var>'s and <var>base</var>'s <a for=url>host</a> are null under these conditions
+          and therefore not copied.
+
+         <li><p>Otherwise, set <var>url</var>'s <a for=url>host</a> to <var>base</var>'s
+         <a for=url>host</a>.
+        </ol>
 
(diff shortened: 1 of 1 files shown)


annevk:
When I implement this in whatwg-url I'm left with a single failing test and currently unsure as how to fix it:
```
  1) Web platform tests parsing </..//localhost//pig> against <file://lion/>:

      AssertionError: href
      + expected - actual

      -file:///localhost//pig
      +file://lion/localhost//pig
```

annevk:
I think the problem is that https://url.spec.whatwg.org/#file-slash-state doesn't copy the base's host over and we never tested that path before... I guess code coverage isn't everything.

annevk:
Should be all good now. Also updated the tests PR.

url.bs
```diff
@@ -2113,6 +2121,12 @@ string <var>input</var>, optionally with a <a>base URL</a> <var>base</var>, opti
 
        <li><p>Set <var>buffer</var> to the empty string.
 
+       <li><p>If <var>url</var>'s <a for=url>scheme</a> is "<code>file</code>" and <a>c</a> is
+       <a>EOF code point</a>, "<code>?</code>", or "<code>#</code>", then while <var>url</var>'s
+       <a for=url>path</a>'s <a for=list>size</a> is not 1 and <var>url</var>'s
+       <a for=url>path</a>[0] is the empty string, <a>validation error</a>, <a for=list>remove</a>
```
rmisev: Path initially can be empty (e.g.: `file://`) and then path[0] can not be accessed. So I think "is not 1" must be changed to "is greater than 1",

annevk: Either we don't have coverage or that doesn't cause failure in the JavaScript implementation.

rmisev: In JavaScript path[0] in such case is undefined, so test `path[0] === ""` fails. But for example in C++ this can cause exception or runtime error.

annevk: Yeah, that's what I figured. Should maybe assert for length being 0 to make sure we have coverage.

We should probably also clarify in the Infra Standard that [] indexing only works for actual entries (or make it work like JavaScript somehow, but not sure that's better).

rmisev: I think just changing "path's size is not 1" to "path's size is greater than 1" solves this problem, because path[0] then will be not accessed.

annevk: So with `file://` I think path will still be non-empty due to 1.4.2. We don't have any testcase that triggers path's size being 0. Maybe we don't test `file://` explicitly though.

(That's the reason I wanted to add the assert first. I want to make sure it would actually be triggered. I agree that greater than is what we want here.)

annevk: Yeah, I can't trigger it with `file://`. I'll change to greater than though for clarity.

PR #285, opened 2017-03-28
Editorial: remove labels from IPv6 parser

annevk:
Fixes part of #282.

***
[Preview](https://url.spec.whatwg.org/branch-snapshots/annevk/ipv6/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/url/91cb2aa...b41b65a.html)

PR #286, opened 2017-03-28
Refactor IPv6 in terms of lists

annevk:
***
[Preview](https://url.spec.whatwg.org/branch-snapshots/annevk/ipv6-more/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/url/20c3257...a4f2eba.html)

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -744,7 +744,7 @@ The <dfn id=concept-ipv4-serializer>IPv4 serializer</dfn> takes an
   the second 0.
 
  <li><p>If there is no sequence of <var>address</var>'s <a>IPv6 pieces</a> that are 0 that is
- greater than 1, then set <var>compress</var> to null.
+ longer than 1, then set <var>compress</var> to null.
 
  <li><p>Let <var>ignore0</var> be false.
 
--- a/url.bs
+++ b/url.bs
@@ -250,12 +250,14 @@ network.
 <p class=note>The <code>example.com</code> and <code>example.com.</code> <a for=/>domains</a> are
 not equivalent and typically treated as distinct.
 
-<p>An <dfn export id=concept-ipv4>IPv4 address</dfn> is a 32-bit identifier.
+<p>An <dfn export id=concept-ipv4>IPv4 address</dfn> is a 32-bit unsigned integer that identifies a
+network address.
 [[RFC791]]
 
-<p>An <dfn export id=concept-ipv6>IPv6 address</dfn> is a 128-bit identifier and
-for the purposes of this specification represented as an ordered list of
-eight <dfn id=concept-ipv6-piece lt='IPv6 piece'>16-bit pieces</dfn>.
+<p>An <dfn export id=concept-ipv6>IPv6 address</dfn> is a 128-bit unsigned integer that identifies a
+network address. For the purposes of this standard it is represented as a <a for=/>list</a> of eight
+16-bit unsigned integers, also known as
+<dfn export lt="IPv6 piece" id=concept-ipv6-piece>IPv6 pieces</dfn>.
 [[RFC4291]]
 
 <p class="note">Support for <code>&lt;zone_id></code> is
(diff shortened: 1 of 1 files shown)


annevk:
This requires #285 to land first.

url.bs
```diff
@@ -780,11 +751,11 @@ The <dfn id=concept-ipv4-serializer>IPv4 serializer</dfn> takes an
    <li><p>Otherwise, if <var>ignore0</var> is true, set <var>ignore0</var> to false.
 
    <li>
-    <p>If <var>compress pointer</var> points to <var>piece</var>, then:
+    <p>If <var>compress</var> points to <var>piece</var>, then:
 
     <ol>
      <li><p>Let <var>separator</var> be "<code>::</code>" if <var>piece</var> is
-     <var>address</var>'s first <a lt='IPv6 piece'>piece</a>, and U+003A (:)  otherwise.
+     <var>address</var>'s first <a>IPv6 piece</a>, and U+003A (:)  otherwise.
```
zcorpan: Double space before "otherwise"

url.bs
```diff
@@ -545,153 +539,133 @@ then runs these steps:
...
-  <var>compress pointer</var> is non-null, then:
+  <p>If <var>compress</var> is non-null, then:
 
   <ol>
-   <li><p>Let <var>swaps</var> be
-   <var>piece pointer</var> &minus; <var>compress pointer</var>.
+   <li><p>Let <var>swaps</var> be <var>piece</var> &minus; <var>compress</var>.
 
-   <li><p>Set <var>piece pointer</var> to seven.
+   <li><p>Set <var>piece</var> to seven.
```
zcorpan: 7

rmisev:
For me it's a bit confusing that **IPv6 parser's** variable <var>piece</var> actually stores not piece's value, but piece's index, while similar variable <var>ipv4Piece</var> stores value. So maybe rename it to more clear <var>pieceIndex</var>; e.g.: `address[pieceIndex]`.

Also I see some confusion in **IPv6 serializer**: the <var>piece</var> means piece's value in 5, 5.1, 5.4 steps, but the same variable is used as index in 5.3, 5.3.1, 5.5.

So I suggest to rewrite some **IPv6 serializer** steps as follows (changes are marked):

...
2. Let compress be <del>a pointer to</del><ins>an index of</ins> the first IPv6 piece in the first longest sequences of address’s IPv6 pieces that are 0.
...
5. For each <del>piece in address’s IPv6 pieces</del><ins>pieceIndex in the range 0 to 7, inclusive</ins>:
 1. If ignore0 is true and <del>piece</del><ins>address[pieceIndex]</ins> is 0, then continue.
 2. Otherwise, if ignore0 is true, set ignore0 to false.
 3. If compress <del>points to piece</del><ins>is equal to pieceIndex</ins>, then:
    1. Let separator be "::" if <ins>pieceIndex is 0 (</ins><del>piece is </del>address’s first IPv6 piece<ins>)</ins>, and U+003A (:) otherwise.
    2. Append separator to output.
    3. Set ignore0 to true and continue.
 4. Append <del>piece</del><ins>address[pieceIndex]</ins>, represented as the shortest possible lowercase hexadecimal number, to output.
  5. If <ins>pieceIndex is not 7 (</ins><del>address[piece] is not </del>address’s last IPv6 piece<ins>)</ins>, then append U+003A (:) to output.

...

url.bs
```diff
@@ -521,22 +523,14 @@ these steps:
 then runs these steps:
 
 <ol>
- <li><p>Let <var>address</var> be a new <a>IPv6 address</a> with its
- <a lt='IPv6 piece'>16-bit pieces</a> initialized to 0.
+ <li><p>Let <var>address</var> be a new <a>IPv6 address</a> whose <a>IPv6 pieces</a> are 0.
```
domenic: I think this would be clearer as "whose IPv6 pieces are all 0"

annevk:
@domenic the main motivation is to use more of Infra and have less innovative language.

annevk:
Sorry for the delay, addressed.

domenic:
FWIW the latest changes bring this much more in line with the jsdom/whatwg-url implementation, by using indexing into a string/array instead of somewhat-magic variables that point to various places. So :+1: from me.

PR #292, opened 2017-04-12
IPv4 in IPv6 cleanup

rmisev:
Changes:

1. Moves _ipv4Piece_ verification step right after the _ipv4Piece_ calculation. Because if the "_ipv4Piece_ > 255" verification returns failure, then no need to increase _pointer_.

2. Moves the last (8.) step to the outside of 6.5.5 loop. This lets drop `c is the EOF` check in this moved step, because loop stops on `EOF`. This obviously gives a better performance.

These changes are based on https://github.com/nodejs/node/pull/12315#discussion_r110858300 and https://github.com/nodejs/node/pull/12315#discussion_r110879264

***
[Preview](https://s3.amazonaws.com/pr-preview/rmisev/url/ipv4-in-ipv6-cleanup.html) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/url/703fcd0...rmisev:a1e1e80.html)

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -623,10 +623,10 @@ then runs these steps:
           <p>Otherwise, set <var>ipv4Piece</var> to <var>ipv4Piece</var> &times; 10 +
           <var>number</var>.
 
-         <li><p>Increase <var>pointer</var> by one.
-
          <li><p>If <var>ipv4Piece</var> is greater than 255, <a>validation error</a>, return
          failure.
+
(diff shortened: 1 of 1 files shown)


annevk:
I'll let @domenic verify this in whatwg-url and then land it as an editorial change. LGTM otherwise though. Thanks!

watilde:
Thanks for putting the discussion into here! +1

PR #299, opened 2017-04-27
Editorial: cleanup references

TimothyGu:
- Use abstract-op to refer to ToASCII and ToUnicode
- Make all terms defined in another specification visible in "Terms defined by
  reference" section
- Use biblio from SpecRef
- List all dependent standards in "Infrastructure" chapter

***
[Preview](https://s3.amazonaws.com/pr-preview/TimothyGu/url/ref-cleanup.html) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/url/488c459...TimothyGu:312d674.html)

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -62,12 +62,18 @@ might increase in scope somewhat.
 
 <p>This specification depends on the Infra Standard. [[!INFRA]]
 
-<p>Some terms used in this specification are defined in the
-DOM, Encoding, IDNA, and Web IDL Standards.
-[[!DOM]]
-[[!ENCODING]]
-[[!IDNA]]
-[[!WEBIDL]]
+<p>Some terms used in this specification are defined in the following standards and specifications:
+
+<ul class=brief>
+ <li>DOM Standard [[!DOM]]
+ <li>Encoding Standard [[!ENCODING]]
+ <li>File API [[!FILEAPI]]
+ <li>HTML Standard [[!HTML]]
+ <li>Media Source Extensions [[!MEDIA-SOURCE]]
+ <li>Media Capture and Streams [[!MEDIACAPTURE-STREAMS]]
+ <li>Unicode IDNA Compatibility Processing [[!UTS46]]
+ <li>Web IDL [[!WEBIDL]]
+</ul>
 
 <hr>
 
@@ -196,7 +202,7 @@ taken when rendering, interpreting, and passing <a for=/>URLs</a> around.
 <p>When rendering and allocating new <a for=/>URLs</a> "spoofing" needs to be considered. An attack
 whereby one <a for=/>host</a> or <a for=/>URL</a> can be confused for another. For instance,
 consider how 1/l/I, m/rn/rri, 0/O, and а/a can all appear eerily similar. Or worse, consider how
-U+202A LEFT-TO-RIGHT EMBEDDING and similar <a>code points</a> are invisible. [[!UTS36]]
+U+202A LEFT-TO-RIGHT EMBEDDING and similar <a>code points</a> are invisible. [[UTR36]]
 
 <p>When passing a <a for=/>URL</a> from party <var>A</var> to <var>B</var>, both need to
 carefully consider what is happening. <var>A</var> might end up leaking data it does not
@@ -275,7 +281,7 @@ U+005C (\), or U+005D (]).
 <a>domain</a> <var>domain</var>, runs these steps:
 
 <ol>
- <li><p>Let <var>result</var> be the result of running <a lt=ToASCII>Unicode ToASCII</a> with
+ <li><p>Let <var>result</var> be the result of running <a abstract-op lt=ToASCII>Unicode ToASCII</a> with
  <i>domain_name</i> set to <var>domain</var>, <i>UseSTD3ASCIIRules</i> set to false,
  <i>processing_option</i> set to <i>Nontransitional_Processing</i>, and <i>VerifyDnsLength</i> set
  to false.
@@ -290,7 +296,7 @@ U+005C (\), or U+005D (]).
 
 <ol>
  <li><p>Let <var>result</var> be the result of running
- <a lt=ToUnicode>Unicode ToUnicode</a> with
+ <a abstract-op lt=ToUnicode>Unicode ToUnicode</a> with
  <i>domain_name</i> set to <var>domain</var>,
  <i>UseSTD3ASCIIRules</i> set to false.
 
@@ -309,7 +315,7 @@ U+005C (\), or U+005D (]).
 
 <ol>
  <li><p>Let <var>result</var> be the result of running
- <a lt=ToASCII>Unicode ToASCII</a> with
+ <a abstract-op lt=ToASCII>Unicode ToASCII</a> with
  <i>domain_name</i> set to <var>domain</var>,
  <i>UseSTD3ASCIIRules</i> set to true, <i>processing_option</i> set to
  <i>Nontransitional_Processing</i>, and <i>VerifyDnsLength</i> set to true.
@@ -317,7 +323,7 @@ U+005C (\), or U+005D (]).
  <li><p>If <var>result</var> is a failure value, return failure.
 
  <li><p>Set <var>result</var> to the result of running
- <a lt=ToUnicode>Unicode ToUnicode</a> with
+ <a abstract-op lt=ToUnicode>Unicode ToUnicode</a> with
  <i>domain_name</i> set to <var>result</var>,
  <i>UseSTD3ASCIIRules</i> set to true.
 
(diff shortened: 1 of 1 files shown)


url.bs
```diff
@@ -3172,23 +3180,26 @@ work.
...
-urlPrefix: https://w3c.github.io/FileAPI/; type: dfn
-    text: blob url store; url: #BlobURLStore
-urlPrefix: https://w3c.github.io/media-source/#idl-def-; type: interface
-    text: MediaSource
-urlPrefix: https://w3c.github.io/mediacapture-main/#idl-def-; type: interface
-    text: MediaStream
-url: http://www.unicode.org/reports/tr46/#ToASCII; type: dfn; text: toascii; spec: IDNA
-url: http://www.unicode.org/reports/tr46/#ToUnicode; type: dfn; text: tounicode; spec: IDNA
+# XXX Cannot use spec attribute here, as a Bikeshed bug currently prevents it
+# from coalescing autolinked and manually linked FILEAPI entries in the index.
```
annevk: @tabatkins do you know about this?

url.bs
```diff
@@ -62,12 +62,18 @@ might increase in scope somewhat.
...
+<ul class=brief>
+ <li>DOM Standard [[!DOM]]
+ <li>Encoding Standard [[!ENCODING]]
+ <li>File API [[!FILEAPI]]
+ <li>HTML Standard [[!HTML]]
+ <li>Media Source Extensions [[!MEDIA-SOURCE]]
+ <li>Media Capture and Streams [[!MEDIACAPTURE-STREAMS]]
+ <li>Unicode IDNA Compatibility Processing [[!IDNA]]
+ <li>Web IDL [[!WEBIDL]]
+</ul>
```
annevk: I wonder, have you found this section useful? I'm never quite sure it adds value given that there's so many ways to discover we use terms from other standards already.

I guess it gives a heads up that some things might be unfamiliar if you haven't read these standards...

TimothyGu: I wouldn't mind if we are using the approach taken by the [HTML Standard](https://html.spec.whatwg.org/#dependencies) where the [terms defined by reference](https://url.spec.whatwg.org/#index-defined-elsewhere) is part of the prose, though I can imagine this being additional work to keep up to date.

But yes, I do think it is a good idea to inform the reader of where definitions of terms can be found in a chronological order -- i.e. before the terms are actually used.

annevk: Okay, I guess this revision is okay with me then. (HTML Standard variant is indeed too much work.)

TimothyGu:
Another thing I noticed is the disparity between the version of specification in the [Bibliography](https://url.spec.whatwg.org/#normative) chapter and in the [term definitions](https://url.spec.whatwg.org/#index-defined-elsewhere). For example, while [FILEAPI] in the bibliography points to the published https://www.w3.org/TR/FileAPI/, the term definition (which uses Bikeshed's autolinking database) points to https://w3c.github.io/FileAPI/. Same issue applies to Web IDL right now, and (after this PR) MSE and Media Capture and Streams as well.

url.bs
TimothyGu: To elaborate, if this patch is applied:

```diff
diff --git a/url.bs b/url.bs
index 6ed22f1..a1e8d5f 100644
--- a/url.bs
+++ b/url.bs
@@ -1303,10 +1303,8 @@ <h3 id=url-parsing>URL parsing</h3>
  <li><p>If <var>url</var>'s <a for=url>scheme</a> is not
  "<code>blob</code>", return <var>url</var>.
 
- <!-- XXX spec attribute needed to work around a Bikeshed bug. See comment in pre.anchors for more
-      info -->
  <li><p>If <var>url</var>'s <a for=url>path</a> <a for=list>is empty</a> or <var>url</var>'s
- <a for=url>path</a>[0] is not in the <a spec=FILEAPI>Blob URL Store</a>, then return <var>url</var>.
+ <a for=url>path</a>[0] is not in the <a>Blob URL Store</a>, then return <var>url</var>.
  [[!FILEAPI]]
 
  <li><p>Set <var>url</var>'s <a for=url>object</a> to the entry in the <a>Blob URL Store</a>
@@ -3189,9 +3187,7 @@ <h2 id=acknowledgments class=no-num>Acknowledgments</h2>
 </pre>
 
 <pre class=anchors>
-# XXX Cannot use spec attribute here, as a Bikeshed bug currently prevents it
-# from coalescing autolinked and manually linked FILEAPI entries in the index.
-urlPrefix: https://w3c.github.io/FileAPI/
+spec: FILEAPI; urlPrefix: https://w3c.github.io/FileAPI/
     type: dfn; text: Blob URL Store; url: #BlobURLStore
 spec: IDNA; urlPrefix: http://www.unicode.org/reports/tr46/
     type: abstract-op; text: ToASCII; url: #ToASCII
```

the HTML changes in the following fashion:

```diff
diff --git a/url-pr-299.html b/url.html
index 088f353..6b1cd18 100644
--- a/url-pr-299.html
+++ b/url.html
@@ -2605,10 +2605,14 @@ work. </p>
    <li>
     <a data-link-type="biblio">[FILEAPI]</a> defines the following terms:
     <ul>
-     <li><a href="https://w3c.github.io/FileAPI/#dfn-Blob">Blob</a>
      <li><a href="https://w3c.github.io/FileAPI/#BlobURLStore">blob url store</a>
     </ul>
    <li>
+    <a data-link-type="biblio">[FILEAPI]</a> defines the following terms:
+    <ul>
+     <li><a href="https://w3c.github.io/FileAPI/#dfn-Blob">Blob</a>
+    </ul>
+   <li>
     <a data-link-type="biblio">[HTML]</a> defines the following terms:
     <ul>
      <li><a href="https://html.spec.whatwg.org/multipage/comms.html#eventsource">EventSource</a>
```

annevk:
Hopefully @tabatkins can help with that as well.

url.bs
domenic: I would remove this list. It's redundant with the normative references list.

url.bs
domenic: We should also just be able to PR File API to export the term.

rmisev:
Yet another change: rename [UTS36] to [UTR36], as latter is used in [IDNA] to refer http://www.unicode.org/reports/tr36/.

See: http://www.unicode.org/reports/tr46/#UTR36

annevk:
We don't have consistent naming of those reference shortnames even within the WHATWG, but if someone wants to make that change I'll take it I suppose.

rmisev:
Not only, [UTR36] refers to **U**nicode **T**echnical **R**eport, so I think [UTS36] is an editorial mistake.

annevk:
Well, but http://www.unicode.org/reports/tr46/ calls itself "Unicode® Technical Standard #46". So I don't think there's a clear story there.

rmisev:
> ... UTR like Unicode Security Considerations (UTR #36) is just a piece of informative material.

There is what I'm talking about: [UTR36], but not [UTS36].

annevk:
Ah okay. Thanks, I learned something and I agree we should fix it.

url.bs
tabatkins: Huh, I wasn't aware of this bug.

tabatkins: Ah, found the issue - FileAPI is, per usual in W3C, a versioned spec, so its full shortname is `fileapi-1`. However, the biblio name is spelled "FileAPI". They happened to run down slightly different codepaths that stored them separately, but then both get corrected to the biblio spelling upon printing.

I've now fixed it to coalesce properly.

url.bs
tabatkins: I sometimes find it valuable to list things that are explicitly "required reading" before a spec will make sense, or things that a spec builds on top of, but otherwise yeah, this is mostly just redundant with the normative refs list and can be deleted.

tabatkins:
> Another thing I noticed is the disparity between the version of specification in the Bibliography chapter and in the term definitions.

Ah, looks like I wasn't paying attention to the document's `Default Biblio Status` metadata when gathering biblio refs due to autolinking.  Fixed now.

tabatkins:
> Yet another change: rename [UTS36] to [UTR36],

One might also consider just adding this ref (and IDNA) to SpecRef via a quick PR, so other people can refer to it as well. ^_^

TimothyGu:
Updated, kudos to @tabatkins for fixing the Bikeshed issues. The preview links don't work right now as the Bikeshed API has not updated yet.

I also changed the [IDNA] reference to [UTS46], since the name "IDNA" can refer to [many](http://www.unicode.org/reports/tr46/#IDNA2003) published documents and thus unfit for inclusion in SpecRef.

TimothyGu:
Okay this PR should be ready to merge now. Now that _Blob URL Store_ is exported from File API, I've removed custom anchors. For now, I left the spec list in Infrastructure alone besides adding other used specifications, but if we were to remove it we should do it in another PR in conjunction with other specs.

PR #300, opened 2017-04-27
Editorial: use Infra in a/x-www-form-urlencoded

TimothyGu:
Also simplifies the parser and removes mention of a nonexistent variable.

***
[Preview](https://s3.amazonaws.com/pr-preview/TimothyGu/url/form-urlencoded-infra.html) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/url/488c459...TimothyGu:542b196.html)

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -129,10 +129,9 @@ contains bytes that are not <a>ASCII bytes</a> might be insecure and is not reco
   <p>For each byte <var>byte</var> in <var>input</var>:
 
   <ol>
-   <li><p>If <var>byte</var> is not `<code>%</code>`, append
-   <var>byte</var> to <var>output</var>.
+   <li><p>If <var>byte</var> is not 0x25 (%), then append <var>byte</var> to <var>output</var>.
 
-   <li><p>Otherwise, if <var>byte</var> is `<code>%</code>` and the next two bytes after
+   <li><p>Otherwise, if <var>byte</var> is 0x25 (%) and the next two bytes after
    <var>byte</var> in <var>input</var> are not in the ranges 0x30 (0) to 0x39 (9),
    0x41 (A) to 0x46 (F), and 0x61 (a) to 0x66 (f), all inclusive, append <var>byte</var> to
    <var>output</var>.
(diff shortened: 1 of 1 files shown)


annevk:
This looks reasonable. The main thing I've been meaning to investigate is if we can decode to scalar values first and only then start worrying about splitting and such.

I think that's what implementations might already be doing, though it has some implications for various algorithms throughout the standard so I haven't done it yet.

annevk:
I should have mentioned that I looked into that in an issue and I feel bad that I haven't now. Sorry about that.

TimothyGu:
Am I correct in understanding that that endeavor is independent from the changes I'm making here?

url.bs
```diff
@@ -2403,16 +2402,12 @@ takes a byte sequence <var>input</var>, and then runs these steps:
...
- string.
+   <li><p>Let <var>nameString</var> and <var>valueString</var> be the result of running <a>UTF-8
+   decode without BOM</a> on the <a lt="percent decode">percent decoding</a> of <var>name</var> and
+   <var>value</var>, respectively.
 
- <li><p>For each name-value tuple in <var>tuples</var>, append a name-value tuple to
- <var>output</var> where the new name and value appended to <var>output</var> are the result of
- running <a>UTF-8 decode without BOM</a> on the <a lt="percent decode">percent decoding</a> of the
- name and value from <var>tuples</var>, respectively, using <var>encoding</var>.
+   <li><p><a for=list>Append</a> <var>nameString</var>/<var>valueString</var> to <var>output</var>.
```
annevk: Could you use the tuple syntax instead? See https://github.com/whatwg/infra/issues/127.

url.bs
```diff
@@ -2378,11 +2377,11 @@ takes a byte sequence <var>input</var>, and then runs these steps:
 
 <ol>
  <li><p>Let <var>sequences</var> be the result of splitting <var>input</var> on
- `<code>&amp;</code>`.
- <!-- XXX define splitting? DOM does not do it -->
+ 0x26 (&amp;).
+ <!-- XXX define strictly splitting for byte sequences in Infra -->
```
annevk: Could you add "or decode earlier" to this comment, since we haven't decided on which is better.

url.bs
```diff
@@ -129,10 +129,9 @@ contains bytes that are not <a>ASCII bytes</a> might be insecure and is not reco
   <p>For each byte <var>byte</var> in <var>input</var>:
 
   <ol>
-   <li><p>If <var>byte</var> is not `<code>%</code>`, append
-   <var>byte</var> to <var>output</var>.
+   <li><p>If <var>byte</var> is not 0x25 (%), append <var>byte</var> to <var>output</var>.
```
annevk: then append*

TimothyGu:
@annevk Done.

annevk:
LGTM, except for the commit message. I don't understand which variable is nonexistent.

TimothyGu:
@annevk currently the [parser](https://url.spec.whatwg.org/#concept-urlencoded-parser) has the following step:

> 5. For each name-value tuple in _tuples_, append a name-value tuple to _output_ where the new name and value appended to output are the result of running UTF-8 decode without BOM on the percent decoding of the name and value from _tuples_, respectively, using **_encoding_**.

The _encoding_ variable is unused and undeclared in the rest of the algorithm, nor is it clear for which algorithm this variable is used as a parameter. Therefore I simply removed it.

annevk:
Thanks, I missed that while scanning.

Issue #53, opened 2015-07-30
IDNA


annevk:
This issue tracks faults in http://www.unicode.org/reports/tr46/ since Unicode doesn't really do that well. If you find an issue, use http://www.unicode.org/reporting.html to report it and then report back here.

annevk:
@SimonSapin reported an issue with the "xn--" prefix not being added for Unicode-only input. No known feedback URL yet.

SimonSapin:
I’ve just submitted the following http://www.unicode.org/reporting.html . I’ll update this when I get a response.

> Subject: xn-- prefix never added in UTS # 46
>
> In http://www.unicode.org/reports/tr46/tr46-15.html#ProcessingStepConvertValidate , the algorithm looks for a xn-- prefix and decodes the rest of the label per Punycode when it is present.
> 
> In http://www.unicode.org/reports/tr46/tr46-15.html#ToASCII however, the xn-- prefix is never added:
> 
> > Convert each label with non-ASCII characters into Punycode [RFC3492]. This may record an error.
> 
> This should probably be replaced with something like:
> 
> > For each label with non-ASCII characters, replace the label with “xn--” followed by the encoding of the label according to Punycode [RFC3492]. This may record an error.

Sebmaster:
My report to Unicode from some time ago which seems to not be fixed yet:

> The Format section (8.1) under Conformance Testing in UTS46 is confusing.
>
> The explanation for the toASCII and toUnicode explains to use the provided processing_option for toUnicode, and always use nontransitional for toASCII.
> However, in the implementation section of toUnicode (4.3), it explains to always call the processing step with nontransitional. The toASCII parameter list provides a processing_option, though.
>
> It looks to me, as if the descriptions for toASCII and toUnicode in the conformance testing section got mixed up. This also applies to the descriptions in the header of IdnaTest.txt.

The other thing is that there's only a single IdnaTest file, but there's no explanation to which algorithm it applies. Is it for IDNA2008, IDNA2003 or UTS46? It seems to be categorized according to Unicode standard instead of IDNA reference, which makes this really confusing. Haven't reported that one yet though.

SimonSapin:
@Sebmaster regarding the other thing, http://www.unicode.org/reports/tr46/#Conformance_Testing explains how "To test for conformance to UTS46" using `IdnaTest.txt`.

Sebmaster:
@SimonSapin I'm not sure that's totally correct either since: 
[...]

Issue #110, opened 2016-04-03
IDNA / UTS #46 "should" requirements (Bidi and Joiners)


SimonSapin:
https://url.spec.whatwg.org/#idna refers (through the “Unicode ToAscii” and “Unicode ToUnicode” algorithms) to http://www.unicode.org/reports/tr46/#Processing and rely on the error flag.

>  The following steps, performed in order, successively alter the input domain_name string and then output it as a converted Unicode string, plus a flag to indicate whether there was an error.

This it turns refers to Section 4.1 http://www.unicode.org/reports/tr46/#Validity_Criteria which has a series of “must” requirements. For example:

> The label must be in Unicode Normalization Form NFC.

This section also has a subsection 4.1.2 http://www.unicode.org/reports/tr46/#Right_to_Left_Scripts

>  In addition, the label **should** meet the requirements for right-to-left characters specified in the Right-to-Left Scripts document of [IDNA2008], and for the CONTEXTJ requirements in the Protocol document of [IDNA2008]. It is strongly recommended that Unicode Technical Report #36, Unicode Security Considerations [UTR36] and Unicode Technical Standard #39, Unicode Security Mechanisms [UTS39] be consulted for information on dealing with confusables, and for characters that should be excluded from identifiers. Note that the recommended exclusions are a superset of those in [IDNA2008]. 

Note “should” (emphasis added) and “strongly recommended” rather than “must”.

If the URL Standard is to define interoperable algorithms, I think it needs to define in which requirements Section 4.1.2 sets the error flag.

(Related: https://github.com/servo/rust-url/issues/179)

annevk:
Interesting, yeah, I don't think that should be enforced. We should probably make this configurable in the IDNA standard.

I noticed the line you quoted mentioned transitional processing. It seems Gecko is successfully shipping non-transitional processing these days. Perhaps Servo can do so too? And maybe the URL Standard should start requiring that? It's still rather contentious whether it's a good idea though...

SimonSapin:
> I noticed the line you quoted mentioned transitional processing. It seems Gecko is successfully shipping non-transitional processing these days. Perhaps Servo can do so too? And maybe the URL Standard should start requiring that? It's still rather contentious whether it's a good idea though...

What do other browsers do?

CC @valenting

annevk:
@SimonSapin all other browsers do transitional as far as I know. See https://bugzilla.mozilla.org/show_bug.cgi?id=1218179 for details.

[...]

PR #309, opened 2017-05-12
Address several IDNA issues

annevk:
As reported at
https://github.com/whatwg/url/issues/53#issuecomment-300718860 this is
causing issues in non-browser implementations.

***
[Preview](https://url.spec.whatwg.org/branch-snapshots/annevk/ToASCII-warning/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/url/a622235...b5bb3d0.html)

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -278,16 +278,18 @@ U+005C (\), or U+005D (]).
 
 <h3 id=idna>IDNA</h3>
 
-<p>The <dfn id=concept-domain-to-ascii>domain to ASCII</dfn> given a
-<a>domain</a> <var>domain</var>, runs these steps:
+<p>The <dfn id=concept-domain-to-ascii>domain to ASCII</dfn> algorithm, given a <a>domain</a>
+<var>domain</var> and optionally a boolean <var>beStrict</var>, runs these steps:
 
 <ol>
+ <li><p>If <var>beStrict</var> is not given, set it to false.
+
  <li>
   <p>Let <var>result</var> be the result of running <a abstract-op lt=ToASCII>Unicode ToASCII</a>
-  with <i>domain_name</i> set to <var>domain</var>, <i>UseSTD3ASCIIRules</i> set to false,
-  <i>CheckHyphens</i> set to false, <i>CheckBidi</i> set to true, <i>CheckJoiners</i> set to true,
-  <i>processing_option</i> set to <i>Nontransitional_Processing</i>, and <i>VerifyDnsLength</i> set
-  to false.
+  with <i>domain_name</i> set to <var>domain</var>, <i>UseSTD3ASCIIRules</i> set to
+  <var>beStrict</var>, <i>CheckHyphens</i> set to false, <i>CheckBidi</i> set to true,
+  <i>CheckJoiners</i> set to true, <i>processing_option</i> set to
+  <i>Nontransitional_Processing</i>, and <i>VerifyDnsLength</i> set to <var>beStrict</var>.
 
   <p class="XXX">This and <a>domain to Unicode</a> below are based on a proposed revision. See
   <a href="https://github.com/whatwg/url/issues/313">issue #313</a>.
@@ -297,8 +299,8 @@ U+005C (\), or U+005D (]).
  <li><p>Return <var>result</var>.
 </ol>
 
-<p>The <dfn id=concept-domain-to-unicode>domain to Unicode</dfn> given a
-<a>domain</a> <var>domain</var>, runs these steps:
+<p>The <dfn id=concept-domain-to-unicode>domain to Unicode</dfn> algorithm, given a <a>domain</a>
+<var>domain</var>, runs these steps:
 
 <ol>
  <li><p>Let <var>result</var> be the result of running
@@ -320,13 +322,10 @@ U+005C (\), or U+005D (]).
 <p>A <var>domain</var> is a <dfn>valid domain</dfn> if these steps return success:
 
 <ol>
- <li><p>Let <var>result</var> be the result of running
- <a abstract-op lt=ToASCII>Unicode ToASCII</a> with
- <i>domain_name</i> set to <var>domain</var>,
- <i>UseSTD3ASCIIRules</i> set to true, <i>processing_option</i> set to
- <i>Nontransitional_Processing</i>, and <i>VerifyDnsLength</i> set to true.
+ <li><p>Let <var>result</var> be the result of running <a>domain to ASCII</a> with <var>domain</var>
+ and true.
 
- <li><p>If <var>result</var> is a failure value, return failure.
+ <li><p>If <var>result</var> is failure, then return failure.
 
  <li><p>Set <var>result</var> to the result of running
  <a abstract-op lt=ToUnicode>Unicode ToUnicode</a> with
--- a/url.bs
+++ b/url.bs
@@ -328,7 +328,10 @@ U+005C (\), or U+005D (]).
  <li><p>If <var>result</var> is failure, then return failure.
 
  <li><p>Set <var>result</var> to the result of running
- <a abstract-op lt=ToUnicode>Unicode ToUnicode</a> with
+ <a abstract-op lt=ToUnicode>Unicode ToUnicode</a> with <i>domain_name</i> set to <var>result</var>,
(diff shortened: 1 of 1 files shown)


domenic:
So what should we do in Node.js/whatwg-url? Should we take this as license to violate the spec, so that our users can use our libraries on YouTube?

I'd kind of prefer the spec just match browsers here (and in most places); we've never had success getting implementers to comment on URL parsing issues before, and I doubt it's going to start here.

annevk:
Sigh. I guess you're right. It's just such dumb Latin-alphabet-centric behavior that I'm having a hard time accepting this is what we want long term. And I'd rather fix issues once than multiple times.

annevk:
Final attempt: https://groups.google.com/d/msg/mozilla.dev.platform/Hd8ATeJcV1A/xEeGH--tAQAJ

rmisev:
Following tests show, what not all browsers use fast path for ASCII only input:

Input           | URL Standard    | Chrome          | Edge                     | Firefox
----------------|-----------------|-----------------|--------------------------|------------------
ab--c           | _(failure)_     | ab--c           | ab--c                    | ab--c
xn--a           | _(failure)_     | xn--a           | _(failure)_              | xn--a
xn--nxa.xn--nxa | xn--nxa.xn--nxa | xn--nxa.xn--nxa | xn--nxa.xn--nxa<br>β.β   | β.β
xn--nxa.β       | xn--nxa.xn--nxa | xn--nxa.xn--nxa | xn--nxa.xn--nxa<br>β.β   | β.β
ab--c.xn--nxa   | _(failure)_     | ab--c.xn--nxa   | ab--c.xn--nxa<br>ab--c.β | ab--c.β
ab--c.β         | _(failure)_     | _(failure)_     | ab--c.xn--nxa<br>ab--c.β | ab--c.β
xn--a.xn--nxa   | _(failure)_     | xn--a.xn--nxa   | _(failure)_              | xn--a.xn--nxa
xn--a.β         | _(failure)_     | _(failure)_     | _(failure)_              | xn--xn--a-pm43a.β

As you see Edge checks the `xn--` labels for ASCII only input, while Chrome doesn't. The Chrome's behavior doesn't look consistent for me. I think Edge's is better.

Firefox doesn't fail on invalid labels at all and the last test result shows browser's bug, which must be fixed anyway.

_Note:_ I tested with `URL` object, prefixing the input with `http://`. If Edge's result has two lines, then first is host returned in `href`, second - `hostname`.

annevk:
Thanks for those tests. Chrome's behavior looks consistent? If all of the input is ASCII, don't run ToASCII. Label splitting happens as part of ToASCII so is not a consideration until then.

rmisev:
If follow Chrome's behavior, then it's unclear `xn--a.xn--nxa` is valid or not. It is accepted when I enter it in the address bar. But then Chrome converts it to `xn--a.β`. So valid becomes invalid... 😕

annevk:
I see, the problem is that they use ToUnicode (which never fails) after which it no longer roundtrips. That does indeed seem bad.

annevk:
@rmisev sorry for leaving this. So what is the processing model we want? ASCII lowercase input; split input on "."; if a label starts with "xn--" or contains non-ASCII, run input through ToASCII? That's getting rather complicated. Always invoking ToASCII and setting the ignore hyphen flag might end up being better, no?

rmisev:
I think simpler is better, so always `ToASCII`.

annevk:
I updated the PR to point to the proposed update of UTS46 and set CheckHyphens to false.

url.bs
```diff
@@ -280,10 +280,14 @@ U+005C (\), or U+005D (]).
...
 <ol>
- <li><p>Let <var>result</var> be the result of running <a abstract-op lt=ToASCII>Unicode ToASCII</a> with
- <i>domain_name</i> set to <var>domain</var>, <i>UseSTD3ASCIIRules</i> set to false,
- <i>processing_option</i> set to <i>Nontransitional_Processing</i>, and <i>VerifyDnsLength</i> set
- to false.
+ <li>
+  <p>Let <var>result</var> be the result of running <a abstract-op lt=ToASCII>Unicode ToASCII</a>
+  with <i>domain_name</i> set to <var>domain</var>, <i>UseSTD3ASCIIRules</i> set to false,
+  <i>CheckHyphens</i> set to false, <i>processing_option</i> set to
+  <i>Nontransitional_Processing</i>, and <i>VerifyDnsLength</i> set to false.
```
TimothyGu: Hmm we technically need to specify *CheckBidi* and *CheckJoiners* too since the spec doesn't offer any default values for those.

annevk: Yeah. I guess I could add a pointer to #110 in the XXX comment about them. Although CheckBidi seems uncontroversial.

Basically I'm not getting feedback from browsers on this which is rather frustrating.

annevk:
Okay, I changed the approach to just embracing the proposed revision of UTS46 fully and thereby addressing the outstanding issues. I filed a follow-up issue to address any fallout if this is still problematic somehow (there might be some more emoji fallout I suspect, but emojis with joiners are already not interoperable in domain names).

rmisev:
I see that the _CheckHyphens_, _CheckBidi_ and _CheckJoiners_ are missing in the "**3.4. Host writing**" -> "**valid domain**".
I'm not sure what _CheckHyphens_ value must be here.

annevk:
Oh, thanks for pointing that out! I guess that depends on whether we think what YouTube is doing is wrong or okay. It's probably fine now everyone is forced to accept it. Fix coming up.

url.bs
```diff
@@ -320,13 +322,10 @@ U+005C (\), or U+005D (]).
...
- <i>UseSTD3ASCIIRules</i> set to true, <i>processing_option</i> set to
- <i>Nontransitional_Processing</i>, and <i>VerifyDnsLength</i> set to true.
+ <li><p>Let <var>result</var> be the result of running <a>domain to ASCII</a> with <var>domain</var>
+ and true.
 
- <li><p>If <var>result</var> is a failure value, return failure.
+ <li><p>If <var>result</var> is failure, then return failure.
 
  <li><p>Set <var>result</var> to the result of running
  <a abstract-op lt=ToUnicode>Unicode ToUnicode</a> with
```
rmisev: You missed to specify CheckHyphens, CheckBidi and CheckJoiners for ToUnicode.

domenic:
@Sebmaster would you be up for helping validate this by making the appropriate changes to tr46.js and whatwg-url? I think you understand this area much better than I do.

TimothyGu:
Is there anything preventing this PR from being applied?

domenic:
We're working on validating the tests via jsdom/whatwg-url in https://github.com/jsdom/whatwg-url/pull/90, but kind of stuck since we don't implement CheckJoiners and CheckBidi. We'll probably go ahead and merge anyway, but would Node.js be able to create an implementation that passes the tests instead? That'd be even better.

TimothyGu:
> would Node.js be able to create an implementation that passes the tests instead?

Yes I think so. In fact, cases 3 and 4 [here](https://github.com/jsdom/whatwg-url/pull/90#issuecomment-305298171) (the two corresponding to CheckBidi) are already handled correctly in Node.js 8.0.0. CheckJoiners was not included in the release, since at the time the PR was merged there was not yet a resolution on CheckJoiners.

domenic:
Great. Would it make sense to put off merging this PR until there is a corresponding verification from Node.js that the tests and spec are implementable? That's been our general preference so far, using either whatwg-url or Node.js.

TimothyGu:
> until there is a corresponding verification from Node.js that the tests and spec are implementable?

And indeed there is: https://github.com/nodejs/node/pull/13362 I'll look into implementing the changes in tr46.js and whatwg-url as well.

Marginally off-topic, but just wanted to point out that there seems to be a circular dependency between the URL Standard, Unicode IDNA spec, whatwg-url, Node.js, and WPT; specifically:

- This PR depends on the proposed draft of Unicode IDNA
- This PR depends on independent verification of implementability (Node.js or whatwg-url)
- Node.js and whatwg-url PRs depend on this PR and WPT updates before they can be applied
- WPT update depends on this PR

domenic:
OK, looks like Node.js has a passing implementation, so let's merge this! BTW the dependency is not cyclic, as we don't need a merged PR, just an existing PR :).

PR #312, opened 2017-05-22
Specify window.webkitURL as an alias to URL

### LINKED ISSUE #135: Let webkitURL be an alias of the URL constructor
Measuring usage of attributes on the global object is hard because it's tainted by enumeration, but this is what we have:
https://www.chromestatus.com/metrics/feature/timeline/popularity/283

(Compare to https://www.chromestatus.com/metrics/feature/timeline/popularity/356 which is likely used very little but is also at ~0.1%.)

`window.webkitURL` has been deprecated in Blink for a long time, and I need to either remove it or get rid of the deprecation message. I'm leaning towards just spec'ing it, because Edge has `window.webkitURL` and it's trivial to support if you also have the infrastructure for other constructor aliases like `WebKitAnimationEvent` or, perhaps, `HTMLDocument` as an alias of `Document` (not yet implemented).

zcorpan:
Fixes #135.

***
[Preview](https://url.spec.whatwg.org/branch-snapshots/zcorpan/webkiturl/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/url/20c3257...62a0abf.html)

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -2532,7 +2532,8 @@ takes a string <var>input</var>, <a>UTF-8 encodes</a> it, and then returns the r
 
 <pre class=idl>
 [Constructor(USVString url, optional USVString base),
- Exposed=(Window,Worker)]
+ Exposed=(Window,Worker),
+ LegacyWindowAlias=webkitURL]
 interface URL {
   stringifier attribute USVString href;
   readonly attribute USVString origin;


zcorpan:
Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1366734

zcorpan:
Bug for Edge
https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/12091311/

domenic:
> This might also be something that Node.js doesn't want to copy. I think we can leave that up to them to decide though.

I guess technically Web IDL mandates [LegacyWindowAlias] to apply to the "primary global interface". Maybe we could tweak it to apply to Window only somehow.

But this is minor as Node doesn't implement the global-ness of URL anyway (yet?).

annevk:
@domenic aaah right, that'll be fixed with https://github.com/heycam/webidl/issues/365 so I think we're totally in the clear then. (Except that at some point Node might want Exposed entries for its global...)

zcorpan:
Relevant issue for chromium https://bugs.chromium.org/p/chromium/issues/detail?id=348985

achristensen07:
Is there a way that chrome stats could measure the uses of webkitURL without URL as a fallback?

zcorpan:
I don't know how that could reasonably be implemented, since JS execution doesn't reach the fallback, there's no way of knowing what the fallback is.

But if there's still interest to remove it in WebKit, Chromium, and EdgeHTML, it seems plausible that it would be successful since Gecko doesn't have it and has no bugs reported.

PR #328, opened 2017-06-27
Normalize port after updating scheme

### LINKED ISSUE #327: Setting scheme should set port to null if it matches the default port
When setting port state, https://url.spec.whatwg.org/#port-state says:

> 3. Set url’s port to null, if port is url’s scheme’s default port, and to port otherwise

However no corresponding verbiage is included in the process for setting a url scheme.  This leads to inconsistent URL states, as illustrated in node.js's current implementation of this standard:

```
> u1 = new URL('https://foo.com:80')
URL {...}
> u2 = new URL('http://foo.com:443')
URL {...}
> u1.protocol = 'http:'
'http:'
> u2.port = 80
80
> String(u1)
'http://foo.com:80/'
> String(u2)
'http://foo.com/'
```

`u1` and `u2` both represent the same URL (scheme, host, port), but u1's port is set to 80, while u2's port is null.

The spec should dictate that when the scheme state is set, the port should be set to null if it matches the default port for that scheme.

TimothyGu:
Closes #327.

Tests: https://github.com/w3c/web-platform-tests/pull/6346

***
[Preview](https://s3.amazonaws.com/pr-preview/TimothyGu/url/scheme-port.html) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/url/b128ba9...TimothyGu:156b207.html)

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -1400,6 +1400,9 @@ string <var>input</var>, optionally with a <a>base URL</a> <var>base</var>, opti
 
        <li><p>Set <var>url</var>'s <a for=url>scheme</a> to <var>buffer</var>.
 
+       <li><p>If <var>url</var>'s <a for=url>port</a> is <var>url</var>'s <a for=url>scheme</a>'s
+       <a>default port</a>, then set <var>url</var>'s <a for=url>port</a> to null.
+
        <li><p>Set <var>buffer</var> to the empty string.
 
        <li><p>If <var>state override</var> is given, then return.
--- a/url.bs
+++ b/url.bs
@@ -1400,12 +1400,17 @@ string <var>input</var>, optionally with a <a>base URL</a> <var>base</var>, opti
 
        <li><p>Set <var>url</var>'s <a for=url>scheme</a> to <var>buffer</var>.
 
-       <li><p>If <var>url</var>'s <a for=url>port</a> is <var>url</var>'s <a for=url>scheme</a>'s
-       <a>default port</a>, then set <var>url</var>'s <a for=url>port</a> to null.
+       <li>
+         <p>If <var>state override</var> is given, then:
+
+         <ol>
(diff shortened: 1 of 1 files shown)


domenic:
Can do.

annevk:
It seems like we should also file a bug against Edge. @TimothyGu are you up for that or shall I do it?

url.bs
```diff
@@ -1400,6 +1400,9 @@ string <var>input</var>, optionally with a <a>base URL</a> <var>base</var>, opti
 
        <li><p>Set <var>url</var>'s <a for=url>scheme</a> to <var>buffer</var>.
 
+       <li><p>If <var>url</var>'s <a for=url>port</a> is <var>url</var>'s <a for=url>scheme</a>'s
+       <a>default port</a>, then set <var>url</var>'s <a for=url>port</a> to null.
```
rmisev: This step is required for the protocol setter only, so it's better to place it after "4. If state override is given"

2. Set url ’s scheme to buffer
3. Set buffer to the empty string
4. If state override is given, then:
   1. **If url ’s port is url’s scheme’s default port , then set url’s port to null**.
   2. Return.

GPHemsley: Ah, yes, I think I agree with this.

And while we're in the area, is there a reason resetting _buffer_ (3) is necessary before returning from a state override (4)?

TimothyGu: @rmisev It's sort of odd to see "Return" on its own line, which was why I refrained from this style initally, but that's fine with me.

@GPHemsley not particularly. Changed.

TimothyGu:
@annevk Issued filed: https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/12555073/

domenic:
whatwg-url confirms spec and test change are in harmony and don't break anything else; merge time!

PR #334, opened 2017-07-10
Editorial: URLSearchParams always has an init parameter

### LINKED ISSUE #333: New URLSearchParams algorithm does not check if optional init variable is set
The algorithm definition begins:
> To create a **new** URLSearchParams object, optionally using _init_, run these steps: 

However, the algorithm itself doesn't check to see if _init_ is given before operating on it, going so far as identify its default case as being a string:
> 4. Otherwise, _init_ is a string, then set _query_’s list to the result of parsing _init_.

Steps 2, 3, and 4 should be enclosed in a check for whether _init_ is actually given before attempting to operate on it.

annevk:
Fixes #333.

***
[Preview](https://url.spec.whatwg.org/branch-snapshots/annevk/urlsearchparams/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/url/54fb58a...4f4e98f.html)

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -2897,7 +2897,7 @@ which is initially empty.
 initially null.
 
 <p>To create a <dfn export for=URLSearchParams id=concept-urlsearchparams-new>new</dfn>
-{{URLSearchParams}} object, optionally using <var>init</var>, run these steps:
+{{URLSearchParams}} object using <var>init</var>, run these steps:
 
 <ol>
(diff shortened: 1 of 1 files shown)


GPHemsley:
Is this waiting on something? (If it's me, 👍 )

annevk:
Just someone to review it, thanks!

PR #338, opened 2017-07-20
Editorial: fix various nits

### LINKED ISSUE #315: Clarify language around scheme state with state override in basic URL parser
In 11.(scheme state).2.1:
> 1. If _url_’s scheme is a special scheme and _buffer_ is not, then return.
> 1. If _url_’s scheme is not a special scheme and _buffer_ is, then return.

This is not unambiguously clear that _buffer_ is to be checked for being a special scheme.

cc @domenic

### LINKED ISSUE #322: Check if "state override" is given before checking its value
From 11(host state).2.5:
> If _state override_ is hostname state, then return.

Since _state override_ need not be given, it would be good to check whether it was given first. (This is the only occurrence where its value matters; every other check is only whether or not it was given.)

### LINKED ISSUE #324: URL's port setter missing context object
Step 2 of the port setter says:
> If the given value is the empty string, then set url’s port to null.

All other mentions of "url" in getters and setters refer to "url" as "context object’s url", including in the other steps of the port setter.

annevk:
Fixes #315, fixes #322, and fixes #324.

***
[Preview](https://url.spec.whatwg.org/branch-snapshots/annevk/nits/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/url/4315826...20c2cb9.html)

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -1386,10 +1386,10 @@ string <var>input</var>, optionally with a <a>base URL</a> <var>base</var>, opti
 
         <ol>
          <li><p>If <var>url</var>'s <a for=url>scheme</a> is a <a>special scheme</a> and
-         <var>buffer</var> is not, then return.
+         <var>buffer</var> is not a <a>special scheme</a>, then return.
 
          <li><p>If <var>url</var>'s <a for=url>scheme</a> is not a <a>special scheme</a> and
-         <var>buffer</var> is, then return.
+         <var>buffer</var> is a <a>special scheme</a>, then return.
 
          <li><p>If <var>url</var> <a>includes credentials</a> or has a non-null <a for=url>port</a>,
          and <var>buffer</var> is "<code>file</code>", then return.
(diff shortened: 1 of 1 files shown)


url.bs
```diff
@@ -2761,8 +2762,8 @@ the setter to always "reset" both.
  <li><p>If <a>context object</a>'s <a for=URL>url</a> <a>cannot have a username/password/port</a>,
  then return.
 
- <li><p>If the given value is the empty string, then set <a for=URL>url</a>'s <a for=url>port</a> to
- null.</p></li>
+ <li><p>If the given value is the empty string, then set <a>context object</a>a>'s
```
rmisev: `... </a>a>'s` - the redundant `a>`?

annevk: Thanks, pushed a fix. Should probably change some settings in the editor...

GPHemsley:
😁

PR #340, opened 2017-08-14
Define percent decoding of strings

annevk:
This is useful for https://github.com/whatwg/fetch/issues/234 and also simplifies the host parser a bit.

***
[Preview](https://url.spec.whatwg.org/branch-snapshots/annevk/string-percent-decode/) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/url/19e0ffa...a772a9e.html)


annevk:
I discovered the HTML Standard definitely needs this, so might as well merge this since it'll be useful irrespective of whether it ends up being used for `data:` URLs.

annevk:
Thanks for the reviews!

Issue #291, opened 2017-04-05
Unescaped '>' should probably not be allowed in URLs


bzbarsky:
The standard way, going back to at least the mid-90s, to mark up URLs in text is `<url>`.  This, of course, relies on unescaped `>` not being allowed in URLs.  This is clearly stated, with exactly this rationale, in RFC 1738 section 2.2.   The URL standard should have similar provisions.

I don't know what that should mean for URL _parsing_, but in terms of serialization '>' should always be escaped in URLs, imo.

I just tested browser behavior, and:

* Firefox consistently escapes '>' in path, userinfo, query, fragment.  '>' in host or port cause parsing failure.
* Safari escapes '>' in path, userinfo, query.  It allows '>' unchanged in host and fragment.  '>' in port causes parsing failure.
* Chrome escapes '>' in path, userinfo, query, host.  It allows '>' unchanged in fragment.  '>' in port causes parsing failure.
* Edge escapes '>' in path and host.  It allows '>' unchanged in fragment and query.  '>' in port causes parsing failure.  Presence of userinfo causes parsing failure no matter what.

Testcase used:

    <pre><script>
      var strs = [
        "http://test>test/foo\\bar",
        "http://a>b@test/foo\\bar",
        "http://test/foo\\bar/#a>b",
        "http://test/foo\\bar/?a=c>d",
        "http://test:2>3/foo\\bar",
        "http://test/foo>bar\\baz",
      ];
      for (var str of strs) {
        var a = document.createElement("a");
        a.setAttribute("href", str);
        var href;
        try {
          href = a.href;
        } catch(e) {
          href = "href getter threw";
        }
        var url;
        try {
          url = (new URL(str).href);
        } catch(e) {
          url = "constructor threw";
        }
        document.writeln(str, " -- ", href, " -- ", url);
      }
    </script>

with the `\\` bits in there a way to tell whether parsing failed in the href case.

bzbarsky:
Note also that there are various other standards (e.g. the one for the Link HTTP header) that rely on being able to put `<>` around a URL to delimit it.

annevk:
Apart from host it seems this is in order: https://jsdom.github.io/whatwg-url/#url=aHR0cHM6Ly9leGFtcGxlLmNvbS88Pj88PiM8Pg==&base=YWJvdXQ6Ymxhbms=. Probably due to #347.

Host is tracked by #458.

PR #347, opened 2017-10-09
Percent-encode additional characters in "fragment state".

### LINKED ISSUE #344: Consider percent-encoding more characters in "fragment state"
Currently, https://url.spec.whatwg.org/#fragment-state requires that we percent-encode characters in a URL's fragment using the [C0 control percent-encode set](https://url.spec.whatwg.org/#c0-control-percent-encode-set). Firefox currently encodes more than that, which I think actually makes a good deal of sense.

How would y'all feel about shifting fragment state encoding to match query state? That is, encoding "less than 0x21 (!), greater than 0x7E (~), or is 0x22 ("), 0x23 (#), 0x3C (<), or 0x3E (>)"?

mikewest:
Currently, we percent-encode characters in "fragment state" using the C0
control percent-encode set. Firefox encodes more than that, and it seems
reasonable to align around that behavior for reasons spelled out in #291
and the comments of #344.

This patch adds a new "fragment percent-encode set" which contains the
C0 control percent-encode set, along with:

* 0x22 (")
* 0x3C (<)
* 0x3E (>)
* 0x60 (`)

Closes #344.

***
[Preview](https://s3.amazonaws.com/pr-preview/mikewest/url/fragment-percent-encode.html) | [Diff](https://s3.amazonaws.com/pr-preview/whatwg/url/20c3257...mikewest:6c80db9.html)

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -170,7 +170,7 @@ contains bytes that are not <a>ASCII bytes</a> might be insecure and is not reco
 and all <a>code points</a> greater than U+007E (~).
 
 <p>The <dfn>fragment percent-encode set</dfn> is the <a>C0 control percent-encode set</a> and
-U+0020 SPACE, U+0022 ("), U+003C(&lt;), U+003E (&gt;), and U+0060 (`).
+U+0020 SPACE, U+0022 ("), U+003C (&lt;), U+003E (&gt;), and U+0060 (`).
 
 <p>The <dfn oldids=default-encode-set>path percent-encode set</dfn> is the
 <a>fragment percent-encode set</a> and U+0023 (#), U+003F (?), U+007B ({), and U+007D (}).
--- a/url.bs
+++ b/url.bs
@@ -169,16 +169,15 @@ contains bytes that are not <a>ASCII bytes</a> might be insecure and is not reco
 <p>The <dfn oldids=simple-encode-set>C0 control percent-encode set</dfn> are the <a>C0 controls</a>
 and all <a>code points</a> greater than U+007E (~).
 
+<p>The <dfn>fragment percent-encode set</dfn> is the <a>C0 control percent-encode set</a> and
+U+0022 SPACE, U+0022 ("), U+003C(&lt;), U+003E (&gt;), and U+0060 (`).
+
 <p>The <dfn oldids=default-encode-set>path percent-encode set</dfn> is the
-<a>C0 control percent-encode set</a> and U+0020 SPACE, U+0022 ("), U+0023 (#), U+003C (&lt;),
-U+003E (>), U+003F (?), U+0060 (`), U+007B ({), and U+007D (}).
+<a>fragment percent-encode set</a> and U+0023 (#), U+003F (?), U+007B ({), and U+007D (}).
 
 <p>The <dfn oldids=userinfo-encode-set>userinfo percent-encode set</dfn> is the
 <a>path percent-encode set</a> and U+002F (/), U+003A (:), U+003B (;), U+003D (=), U+0040 (@),
 U+005B ([), U+005C (\), U+005D (]), U+005E (^), and U+007C (|).
-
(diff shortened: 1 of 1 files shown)


mikewest:
I'll put tests together shortly.

url.bs
```diff
@@ -177,6 +177,9 @@ U+003E (>), U+003F (?), U+0060 (`), U+007B ({), and U+007D (}).
 <a>path percent-encode set</a> and U+002F (/), U+003A (:), U+003B (;), U+003D (=), U+0040 (@),
 U+005B ([), U+005C (\), U+005D (]), U+005E (^), and U+007C (|).
 
+<p>The <dfn>fragment percent-encode set</dfn> is the <a>C0 control percent-encode set</a> and
```
TimothyGu: How about we try to define path percent-encode set as a superset of fragment percent-encode set, for parity with userinfo->path.

mikewest: Makes sense to me. Poked at it in https://github.com/whatwg/url/pull/347/commits/f258f4dd4dfba66c7eed17534752065eed8a748a. WDYT?

url.bs
```diff
@@ -169,9 +169,11 @@ contains bytes that are not <a>ASCII bytes</a> might be insecure and is not reco
 <p>The <dfn oldids=simple-encode-set>C0 control percent-encode set</dfn> are the <a>C0 controls</a>
 and all <a>code points</a> greater than U+007E (~).
 
+<p>The <dfn>fragment percent-encode set</dfn> is the <a>C0 control percent-encode set</a> and
+U+0022 SPACE, U+0022 ("), U+003C(&lt;), U+003E (&gt;), and U+0060 (`).
```
annevk: U+0020 SPACE.

mikewest: Yup. That was super-dumb. Fixed in https://github.com/whatwg/url/pull/347/commits/d1d85b24579f29555a1b327ad56d85cc77bdfe9a.

url.bs
```diff
@@ -169,9 +169,11 @@ contains bytes that are not <a>ASCII bytes</a> might be insecure and is not reco
 <p>The <dfn oldids=simple-encode-set>C0 control percent-encode set</dfn> are the <a>C0 controls</a>
 and all <a>code points</a> greater than U+007E (~).
 
+<p>The <dfn>fragment percent-encode set</dfn> is the <a>C0 control percent-encode set</a> and
+U+0020 SPACE, U+0022 ("), U+003C(&lt;), U+003E (&gt;), and U+0060 (`).
```
annevk: Missing space after U+003C.

domenic:
The OP/commit message omit U+0020 space

annevk:
Tests moved to https://github.com/w3c/web-platform-tests/pull/7776 though setter tests are not updated. Seems like jsdom doesn't run those.

mikewest:
Snapping back to this after @noncombatant pinged the internal bug. Sorry I lost the thread.

@domenic, @annevk: Does the current set of tests on https://github.com/w3c/web-platform-tests/pull/7776 make sense to you? If so, I think we're done here. :) If not, I'll fix them.

annevk:
Looks good to me, but I'll let @domenic verify with the jsdom impl.

domenic:
Tests are still not quite right, so it's worth straightening that out first.

domenic:
Tests good, this (and its tests) can be merged :)

annevk:
@mikewest could you file implementation bugs? I'm guessing just for Edge/Safari?

mikewest:
WebKit: https://bugs.webkit.org/show_bug.cgi?id=180396
Edge: https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/14951215/ (looks like the latter has some escaping issues with URLs. :( )

annevk:
Thanks, I emailed some folks at Microsoft to look into that, since it makes reporting bugs against Edge much harder than it should be.

RByers:
For the record, chromium bug here: https://bugs.chromium.org/p/chromium/issues/detail?id=803103

PR #351, opened 2017-10-31
Meta: use the common deploy.sh

foolip:
The only irregular thing about url compared to other specs is that
.htaccess is checked in and copied by the old deploy.sh script. It's
removed here, but will be left on the old server (75.119.197.251) since
rsync --delete isn't used.


foolip:
I'm working on a separate PR for misc-server to host url.spec.whatwg.org. With both merged and the DNS updated, all should be well.

foolip:
> Shouldn't we use rsync --delete?

Yes, I think we should, but in the transition period that's been a risky thing. Maybe once the old deploy server is gone from deploy.sh entirely.

foolip:
https://github.com/whatwg/misc-server/pull/32 is that other PR if you'd like to review.

foolip:
Oops, I forgot to merge this :/

PR #360, opened 2017-12-18
Meta: update copyright license and Bikeshed boilerplate

domenic:
See https://blog.whatwg.org/copyright-license-change for the former and
https://github.com/whatwg/meta/issues/23 plus
https://github.com/tabatkins/bikeshed/pull/1155 for the latter. The new
boilerplate includes the new copyright license, so this commit just
removes it from the source file, as it does many other things that are
now deduplicated into the Bikeshed boilerplate.

***
<a href="https://whatpr.org/url/360.html" title="Last updated on Jan 5, 2018, 3:29 PM GMT (a1e7f5a)">Preview</a> | <a href="https://whatpr.org/url/360/7a3c69f...a1e7f5a.html" title="Last updated on Jan 5, 2018, 3:29 PM GMT (a1e7f5a)">Diff</a>


domenic:
Hmm, this doesn't update the LICENSE file. "do not merge yet" until both the Bikeshed change gets merged and I fix that.

domenic:
Blocked on https://github.com/whatwg/sg/issues/51

url.bs
```diff
@@ -1,26 +1,12 @@
...
-Logo: https://resources.whatwg.org/logo-url.svg
-Boilerplate: omit feedback-header, omit conformance
-!Participate: <a href=https://github.com/whatwg/url>GitHub whatwg/url</a> (<a href=https://github.com/whatwg/url/issues/new>new issue</a>, <a href="https://github.com/whatwg/url/issues">open issues</a>)
-!Participate: <a href="https://wiki.whatwg.org/wiki/IRC">IRC: #whatwg on Freenode</a>
-!Commits: <a href="https://github.com/whatwg/url/commits">https://github.com/whatwg/url/commits</a>
-!Commits: [SNAPSHOT-LINK]
-!Commits: <a href="https://twitter.com/urlstandard">@urlstandard</a>
-!Tests: <a href=https://github.com/w3c/web-platform-tests/tree/master/url>web-platform-tests url/</a> (<a href=https://github.com/w3c/web-platform-tests/labels/url>ongoing work</a>)
-!Translation (non-normative): <span title=Japanese><a href=https://triple-underscore.github.io/URL-ja.html lang=ja hreflang=ja rel=alternate>日本語</a></span>
+Translation: ja https://triple-underscore.github.io/URL-ja.html
```
annevk: I'm pretty sure this does not generate the desired markup. That's why I use `!Translation (non-normative)` throughout.

annevk:
See also https://github.com/whatwg/meta/issues/56 btw.

url.bs
domenic: Check out the result; it seemed identical to me. But we can also fix Bikeshed while we're working on this.

annevk: I think we should at least change the message to just say "non-normative". "Likely out-of-date" is not friendly to the translators. Is the generated output available somewhere? https://whatpr.org/url/360.html has a very different heading, though I suspect that's due to a Bikeshed mismatch.

domenic: In https://github.com/tabatkins/bikeshed/pull/1155 I changed it to "potentially". Is that good or do you prefer just removing it?

domenic: Here is the output:

domenic: Here is the output:

domenic:
Here is the output:

```html
<dt>Translations <small>(non-normative and potentially out-of-date)</small>:</dt>
<dd><span title="Japanese"><a href="https://triple-underscore.github.io/URL-ja.html" hreflang="ja" rel="alternate" lang="ja">日本語</a></span></dd>
```

url.bs
annevk: I prefer just removing it.

domenic: OK, I'll update the Bikeshed PR; hopefully @tabatkins will be OK with it since I think it's mostly (all?) WHATWG specs that use this.

annevk: I think I was waiting on https://github.com/tabatkins/bikeshed/issues/1058 to be fixed, but then forgot about it or maybe I was waiting for us to use our own templating since it didn't quite work out, even if that were fixed.

annevk:
This change looks reasonable. Can we generate an example copy to preview the standard with the Bikeshed changes? Would be nice to see what changed and how.

domenic:
https://url-egupsydbxo.now.sh

url.bs
```diff
@@ -3,7 +3,7 @@ Group: WHATWG
 H1: URL
 Shortname: url
 Text Macro: TWITTER urlstandard
-Abstract: The URL Standard defines URLs, domains, IP addresses, the <code title>application/x-www-form-urlencoded</code> format, and their API.
+Abstract: The URL Standard defines URLs, domains, IP addresses, the <code><a>application/x-www-form-urlencoded</a></code> format, and their API.
```
annevk: I don't think we want links in abstracts.

domenic: Why not? I assume that's what the `title` was supposed to be about.

annevk: The empty title was there to prevent a link.

annevk: The idea is that abstracts stand on their own. If you need to click on links it has kinda failed. Perhaps that means this abstract is not super great though.

domenic: Happy to remove it, but I think links in abstracts are nice, just like they are anywhere else.

PR #366, opened 2018-01-09
Editorial: fix port value in the origin's example

rmisev:
The URLs port in the tuple origin must be null, because port is not
specified in the input:
`blob:https://whatwg.org/d0360e2f-caee-469f-9a2f-87d5b0456f6f`

According to specification, if port is not specified or is default for
scheme, then URLs port is set to null:
https://url.spec.whatwg.org/#concept-url-port
https://url.spec.whatwg.org/#ref-for-default-port%E2%91%A0

***
<a href="https://whatpr.org/url/366.html" title="Last updated on Jan 9, 2018, 2:31 PM GMT (35d6bd5)">Preview</a> | <a href="https://whatpr.org/url/366/0d15b3f...35d6bd5.html" title="Last updated on Jan 9, 2018, 2:31 PM GMT (35d6bd5)">Diff</a>

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -2306,7 +2306,7 @@ background information. [[!HTML]]
 
   <p class="example no-backref" id=example-43b5cea5>The <a for=url>origin</a> of
   <code>blob:https://whatwg.org/d0360e2f-caee-469f-9a2f-87d5b0456f6f</code> is the tuple
-  (<code>https</code>, <code>whatwg.org</code>, <code>443</code>, null).
+  (<code>https</code>, <code>whatwg.org</code>, null, null).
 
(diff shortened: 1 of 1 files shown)


annevk:
Thanks, it might take a while before this is up on url.spec.whatwg.org due to the build script failing for unrelated reasons. Looking into it.

annevk:
(I also checked HTML and it seems we're not making this mistake there as I was afraid of.)

domenic:
@rmisev will you or your company be able to sign https://participate.whatwg.org/agreement ? It'll be mandatory starting on the 11th.

annevk:
(Build problem issues resolved.)

PR #371, opened 2018-02-02
Use definitions in FileAPI spec to resolve blob URLs and their origins.

### LINKED ISSUE #127: Origin of blob URLs doesn't match what implementations do
As mentioned in [Origin of Blob URLs](https://w3c.github.io/FileAPI/#originOfBlobURL), the origin of a blob URL should be the origin of the incumbent settings object when the URL was created. The simple "parse the origin from the url in the first string of the path of the blob url" algorithm the URL spec uses doesn't accomplish that for opaque origins.

The way at least chrome implements this is by maintaining a separate blob url -> origin mapping, and using that to lookup the origin of a blob url. I assume firefox does someting similar (see also whatwg/html#1322 for an example where this behavior matters).

Should the URL spec be updated to actually match the File API spec about blob URLs, and to match what implementations seem to do?

### LINKED ISSUE #290: there can be more than one Blob URL Store
[URL parsing](https://url.spec.whatwg.org/#url-parsing) refers to looking up a path in '[the blob URL store](https://w3c.github.io/FileAPI/#BlobURLStore)', but the File API spec defines one Blob URL Store per global object, so there are as many BUSes as there are global objects. You'd have to attempt lookup in all of them one by one, with the current spec language.

mkruisselbrink:
This fixes #290, fixes #127, fixes w3c/FileAPI#20 and fixes w3c/FileAPI#63.

The only observable change here is the origin for blob URLs created in unique origins (which is only observable indirectly by trying to fetch such URLs), and tests for that already exist in https://w3c-test.org/FileAPI/url/sandboxed-iframe.html.

***
<a href="https://whatpr.org/url/371.html" title="Last updated on Jan 11, 2019, 9:19 AM UTC (8b426c4)">Preview</a> | <a href="https://whatpr.org/url/371/07ecb3b...8b426c4.html" title="Last updated on Jan 11, 2019, 9:19 AM UTC (8b426c4)">Diff</a>

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -1120,8 +1120,9 @@ resource the <a for=/>URL</a>'s other components identify. It is initially null.
 <p id=non-relative-flag>A <a for=/>URL</a> also has an associated
 <dfn export for=url>cannot-be-a-base-URL flag</dfn>. It is initially unset.
 
-<p>A <a for=/>URL</a> also has an associated <dfn export for=url id=concept-url-blob-entry>blob URL entry</dfn>
-that is either null or a <a for=/>blob URL entry</a>. It is initially null.
+<p>A <a for=/>URL</a> also has an associated
+<dfn export for=url id=concept-url-blob-entry>blob URL entry</dfn> that is either null or a
+<a for=/>blob URL entry</a>. It is initially null.
 
 <p class="note no-backref">This is used to support caching the object a "<code>blob</code>" URL
 refers to as well as its origin. It is important that these are cached as the <a for=/>URL</a> might
@@ -1428,8 +1429,8 @@ different document encoding. Using the <a>UTF-8</a> encoding everywhere solves t
  "<code>blob</code>", return <var>url</var>.
 
  <li><p>Set <var>url</var>'s <a for=url>blob URL entry</a> to the result of
- <a for="blob URL" lt="resolve">resolving the blob URL</a> <var>url</var>,
- if that did not return failure, and null otherwise.
+ <a for="blob URL" lt="resolve">resolving the blob URL</a> <var>url</var>, if that did not return
+ failure, and null otherwise.
 
  <li><p>Return <var>url</var>.
 </ol>
@@ -2432,16 +2433,18 @@ background information. [[!HTML]]
 <dl class=switch>
  <dt>"<code>blob</code>"
  <dd>
-  <p>If <a for=/>URL</a>'s <a for=url>blob URL entry</a> is not null, return <a for=/>URL</a>'s
-  <a for=url>blob URL entry</a>'s <a for="blob URL entry">environment</a>'s
-  <a for="environment settings object">origin</a>.
+  <ol>
+   <li><p>If <a for=/>URL</a>'s <a for=url>blob URL entry</a> is non-null, then return
+   <a for=/>URL</a>'s <a for=url>blob URL entry</a>'s <a for="blob URL entry">environment</a>'s
+   <a for="environment settings object">origin</a>.
 
-  <p>Let <var>url</var> be the result of <a lt="basic URL parser">parsing</a> <a for=/>URL</a>'s
-  <a for=url>path</a>[0].
+   <li><p>Let <var>url</var> be the result of <a lt="basic URL parser">parsing</a>
+   <a for=/>URL</a>'s <a for=url>path</a>[0].
 
-  <p>Return a new <a>opaque origin</a>, if <var>url</var> is failure, and <var>url</var>'s
-  <a for=url>origin</a> otherwise.
-  <!-- Did you mean: recursion -->
+   <li><p>Return a new <a>opaque origin</a>, if <var>url</var> is failure, and <var>url</var>'s
+   <a for=url>origin</a> otherwise.
+   <!-- Did you mean: recursion -->
+  </ol>
 
   <p class="example no-backref" id=example-43b5cea5>The <a for=url>origin</a> of
   <code>blob:https://whatwg.org/d0360e2f-caee-469f-9a2f-87d5b0456f6f</code> is the tuple
--- a/url.bs
+++ b/url.bs
@@ -1129,6 +1129,12 @@ that is null, a {{Blob}} object, or a {{MediaSource}} object. It is initially nu
 "<code>blob</code>" <a for=/>URLs</a>, but others can be added going forward, hence
 "object".
 
+<p>A <a for=/>URL</a> also has an associated
+<dfn for=url>cached origin</dfn> that is either null or a <a>ASCII string</a>.
+It is initially null.
+
+<p class="note">This is used to support caching the origin of a "<code>blob</code>"
+<a for=/>URL</a> at the time it is parsed.
 
 <h3 id=url-miscellaneous>URL miscellaneous</h3>
 
@@ -1429,12 +1435,11 @@ different document encoding. Using the <a>UTF-8</a> encoding everywhere solves t
  <li><p>If <var>url</var>'s <a for=url>scheme</a> is not
  "<code>blob</code>", return <var>url</var>.
 
- <li><p>If <var>url</var>'s <a for=url>path</a> <a for=list>is empty</a> or <var>url</var>'s
- <a for=url>path</a>[0] is not in the <a>Blob URL Store</a>, then return <var>url</var>.
- [[!FILEAPI]]
+ <li><p>Set <var>url</var>'s <a for=url>cached origin</a> to the result of
+ <a for="blob URL" lt="resolve the origin">resolving the origin of the blob URL</a> <var>url</var>.
 
- <li><p>Set <var>url</var>'s <a for=url>object</a> to the entry in the <a>Blob URL Store</a>
- corresponding to <var>url</var>'s <a for=url>path</a>[0].
+ <li><p>Set <var>url</var>'s <a for=url>object</a> to the result of
+ <a for="blob URL" lt="resolve">resolving the blob URL</a> <var>url</var>.
 
  <li><p>Return <var>url</var>.
 </ol>
@@ -2437,11 +2442,10 @@ background information. [[!HTML]]
 <dl class=switch>
  <dt>"<code>blob</code>"
  <dd>
-  <p>Let <var>url</var> be the result of <a lt="basic URL parser">parsing</a> <a for=/>URL</a>'s
-  <a for=url>path</a>[0].
+  If <a for=/>URL</a>'s <a for=url>cached origin</a> is not null, return
+  <a for=/>URL</a>'s <a for=url>cached origin</a>.
 
-  <p>Return a new <a>opaque origin</a>, if <var>url</var> is failure, and <var>url</var>'s
-  <a for=url>origin</a> otherwise.
+  <p>Return the result of <a for="blob URL" lt="resolve the origin">resolving the origin of the blob URL</a>.
   <!-- Did you mean: recursion -->
 
   <p class="example no-backref" id=example-43b5cea5>The <a for=url>origin</a> of
--- a/url.bs
+++ b/url.bs
@@ -1120,8 +1120,8 @@ resource the <a for=/>URL</a>'s other components identify. It is initially null.
 <p id=non-relative-flag>A <a for=/>URL</a> also has an associated
 <dfn export for=url>cannot-be-a-base-URL flag</dfn>. It is initially unset.
 
-<p>A <a for=/>URL</a> also has an associated <dfn export for=url id=concept-url-blob-entry>blob entry</dfn>
-that is either null or a <a>blob URL entry</a>. It is initially null.
+<p>A <a for=/>URL</a> also has an associated <dfn export for=url id=concept-url-blob-entry>blob URL entry</dfn>
+that is either null or a <a for=/>blob URL entry</a>. It is initially null.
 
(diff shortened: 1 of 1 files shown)


url.bs
```diff
@@ -1026,6 +1026,12 @@ resource the <a for=/>URL</a>'s other components identify. It is initially null.
 "<code>blob</code>" <a for=/>URLs</a>, but others can be added going forward, hence
 "object".
 
+<p>A <a for=/>URL</a> also has an associated
+<dfn for=url>cached origin</dfn> that is either null or a <a>ASCII string</a>.
+It is initially null.
+
+<p class="note">This is used to support caching the origin of a "<code>blob</code>"
+<a for=/>URL</a> at the time it is parsed.
```
annevk: Shall we call it object origin then? Also, is this observable? If not, maybe we shouldn't make it a thing to avoid folks making it observable?

url.bs
```diff
@@ -2311,11 +2316,10 @@ background information. [[!HTML]]
...
  <dd>
-  <p>Let <var>url</var> be the result of <a lt="basic URL parser">parsing</a> <a for=/>URL</a>'s
-  <a for=url>path</a>[0].
+  If <a for=/>URL</a>'s <a for=url>cached origin</a> is not null, return
+  <a for=/>URL</a>'s <a for=url>cached origin</a>.
 
-  <p>Return a new <a>opaque origin</a>, if <var>url</var> is failure, and <var>url</var>'s
-  <a for=url>origin</a> otherwise.
+  <p>Return the result of <a for="blob URL" lt="resolve the origin">resolving the origin of the blob URL</a>.
   <!-- Did you mean: recursion -->
```
annevk: There's no more recursion in your algorithm so this can be removed.

url.bs
```diff
@@ -1026,6 +1026,12 @@ resource the <a for=/>URL</a>'s other components identify. It is initially null.
 "<code>blob</code>" <a for=/>URLs</a>, but others can be added going forward, hence
 "object".
 
+<p>A <a for=/>URL</a> also has an associated
+<dfn for=url>cached origin</dfn> that is either null or a <a>ASCII string</a>.
```
annevk: an ASCII*

url.bs
annevk: Hmm I see now that https://w3c.github.io/FileAPI/#blob-url-resolve-the-origin still does recursion. I guess the advantage of that is making it less observable when things are revoked and allowing for a fast path.

url.bs
```diff
@@ -1325,12 +1331,11 @@ different document encoding. Using the <a>UTF-8</a> encoding everywhere solves t
...
- <li><p>If <var>url</var>'s <a for=url>path</a> <a for=list>is empty</a> or <var>url</var>'s
- <a for=url>path</a>[0] is not in the <a>Blob URL Store</a>, then return <var>url</var>.
- [[!FILEAPI]]
+ <li><p>Set <var>url</var>'s <a for=url>cached origin</a> to the result of
+ <a for="blob URL" lt="resolve the origin">resolving the origin of the blob URL</a> <var>url</var>.
 
- <li><p>Set <var>url</var>'s <a for=url>object</a> to the entry in the <a>Blob URL Store</a>
- corresponding to <var>url</var>'s <a for=url>path</a>[0].
+ <li><p>Set <var>url</var>'s <a for=url>object</a> to the result of
+ <a for="blob URL" lt="resolve">resolving the blob URL</a> <var>url</var>.
```
rmisev: The "[resolving the blob URL](https://w3c.github.io/FileAPI/#blob-url-resolve)" algorithm can return *failure*. I guess the url’s `object` should be set to `null` then.

url.bs
mkruisselbrink: Yeah, object origin makes sense. It's only observable for fetches to/from unique origins. The string serialization of the origin isn't changing, just that identity comparison with unique origined blob URLs will now work correctly (especially if the blob URL was parsed/resolved before the URL was revoked, but fetched after being revoked. Without caching this origin same-origin checks would fail during fetching in unique origins).

url.bs
mkruisselbrink: Done

url.bs
mkruisselbrink: Yeah, making it less observable when things are revoked, and also making it work for URLs that were never created in the first place, or if some spec creates a URL record manually without going through the parsing steps.

url.bs
```diff
@@ -1026,6 +1026,12 @@ resource the <a for=/>URL</a>'s other components identify. It is initially null.
 "<code>blob</code>" <a for=/>URLs</a>, but others can be added going forward, hence
 "object".
 
+<p>A <a for=/>URL</a> also has an associated
+<dfn for=url>object origin</dfn> that is either null or an <a>ASCII string</a>.
+It is initially null.
+
+<p class="note">This is used to support caching the origin of a "<code>blob</code>"
+<a for=/>URL</a> at the time it is parsed.
```
annevk: I think it would be helpful to add here that this is important as it might be removed from the blob URL store after that time.

url.bs
```diff
@@ -1026,6 +1026,12 @@ resource the <a for=/>URL</a>'s other components identify. It is initially null.
 "<code>blob</code>" <a for=/>URLs</a>, but others can be added going forward, hence
 "object".
 
+<p>A <a for=/>URL</a> also has an associated
+<dfn for=url>object origin</dfn> that is either null or an <a>ASCII string</a>.
+It is initially null.
```
annevk: Actually, shouldn't this be an origin rather than a string?

url.bs
```diff
@@ -1325,12 +1331,12 @@ different document encoding. Using the <a>UTF-8</a> encoding everywhere solves t
  <li><p>If <var>url</var>'s <a for=url>scheme</a> is not
  "<code>blob</code>", return <var>url</var>.
 
- <li><p>If <var>url</var>'s <a for=url>path</a> <a for=list>is empty</a> or <var>url</var>'s
- <a for=url>path</a>[0] is not in the <a>Blob URL Store</a>, then return <var>url</var>.
- [[!FILEAPI]]
+ <li><p>Set <var>url</var>'s <a for=url>object origin</a> to the result of
+ <a for="blob URL" lt="resolve the origin">resolving the origin of the blob URL</a> <var>url</var>.
```
annevk: Should we only set this if url's object is non-null?

annevk:
What did you think about the alternate design I offered (rephrased to fix errors)? We'd move the definition of object origin into URL: return url's object's relevant global object's origin if url's object is non-null and otherwise keep the existing steps.

Yet another alternative would be that File API returns a https://infra.spec.whatwg.org/#struct or some such wherein both the object and origin are stored. Going into File API twice for the "same thing" seems bad.

mkruisselbrink:
> What did you think about the alternate design I offered (rephrased to fix errors)? We'd move the definition of object origin into URL: return url's object's relevant global object's origin if url's object is non-null and otherwise keep the existing steps.
> 
> Yet another alternative would be that File API returns a https://infra.spec.whatwg.org/#struct or some such wherein both the object and origin are stored. Going into File API twice for the "same thing" seems bad.

Oh sorry, missed this comment. I don't think the first option works (the url's object's relevant global object is not necesarily the same as the current settings object at the time createObjectURL was called), but something like the second option could definitely work. We can just store the https://w3c.github.io/FileAPI/#blob-url-entry and extract the actual blob and origin from there. Changing the URL's object from being the blob to being a struct containing the blob would have the possibly unfortunate side effect of requiring every existing reference to that in other specs to be updated as well (which would mean at least the fetch spec, not sure if other specs refer to it). Probably wouldn't be too much work though.

mkruisselbrink:
Okay, got rid of object and replaced it with something for now called "blob entry" instead. Does this seem more in line with what you had in mind? I'll need to update the File API spec to export the relevant terms (and get rid of the no longer needed origin resolving algorithm there), but don't want to do that unless you're happy with this approach.

url.bs
```diff
@@ -1120,15 +1120,13 @@ resource the <a for=/>URL</a>'s other components identify. It is initially null.
...
-<p>A <a for=/>URL</a> also has an associated <dfn export for=url id=concept-url-object>object</dfn>
-that is null, a {{Blob}} object, or a {{MediaSource}} object. It is initially null.
-[[!FILEAPI]]
-[[!MEDIA-SOURCE]]
-
-<p class="note no-backref">At this point this is used primarily to support
-"<code>blob</code>" <a for=/>URLs</a>, but others can be added going forward, hence
-"object".
+<p>A <a for=/>URL</a> also has an associated <dfn export for=url id=concept-url-blob-entry>blob entry</dfn>
+that is either null or a <a>blob URL entry</a>. It is initially null.
```
annevk: This seems reasonable. Any reason for the name difference?

url.bs
```diff
@@ -1429,12 +1427,9 @@ different document encoding. Using the <a>UTF-8</a> encoding everywhere solves t
...
 
- <li><p>If <var>url</var>'s <a for=url>path</a> <a for=list>is empty</a> or <var>url</var>'s
- <a for=url>path</a>[0] is not in the <a>Blob URL Store</a>, then return <var>url</var>.
- [[!FILEAPI]]
-
- <li><p>Set <var>url</var>'s <a for=url>object</a> to the entry in the <a>Blob URL Store</a>
- corresponding to <var>url</var>'s <a for=url>path</a>[0].
+ <li><p>Set <var>url</var>'s <a for=url>blob entry</a> to the result of
+ <a for="blob URL" lt="resolve">resolving the blob URL</a> <var>url</var>,
+ or null if that returned failure.
```
annevk: if that did not return failure, and null otherwise.*

url.bs
```diff
@@ -2437,11 +2432,16 @@ background information. [[!HTML]]
 <dl class=switch>
  <dt>"<code>blob</code>"
  <dd>
+  If <a for=/>URL</a>'s <a for=url>blob entry</a> is not null, return <a for=/>URL</a>'s
+  <a for=url>blob entry</a>'s <a for="blob URL entry">environment</a>'s
+  <a for="environment settings object">origin</a>.
```
annevk: Presumably we'd return `<a for="blob URL entry">origin</a>`? Or do we need to store the environment on this entry as well?

url.bs
```diff
@@ -2437,11 +2432,16 @@ background information. [[!HTML]]
...
+
   <p>Let <var>url</var> be the result of <a lt="basic URL parser">parsing</a> <a for=/>URL</a>'s
   <a for=url>path</a>[0].
 
   <p>Return a new <a>opaque origin</a>, if <var>url</var> is failure, and <var>url</var>'s
   <a for=url>origin</a> otherwise.
+
```
annevk: This newline can now be dropped.

url.bs
mkruisselbrink: With the name change it's obvious that references to the old term need to be updated somehow. If I had just kept the old term it would be much harder to figure out where references to the old definition exist that now need to be updated because it is a different type. Also it seemed weird to be talking about a URL's object's object, but that's pretty minor of course.

url.bs
mkruisselbrink: The FileAPI spec at least has to store the environment for an entry (to be able to revoke/remove entries when the environment that created them is destroyed). Although of course I could change how the bookkeeping is spec'ed; this just seemed to simplest way from the FileAPI side of things at the time.

I could also change the type/tuple that is returned by resolving a blob URL to something other than what is stored in the blob URL store, but not sure I see benefits to adding that complication (other than separating internal state from what is exposed to other specs).

url.bs
annevk: Sorry, I meant between "blob entry" and "blob URL entry".

url.bs
annevk: That's fair.

url.bs
mkruisselbrink: Ah, no, no particular reason. Changed it to "blob URL entry".

mkruisselbrink:
> > Great, will you make the changes to Fetch too?
> 
> Yep, will prepare a PR for that today.

https://github.com/whatwg/fetch/pull/857, although obviously that won't build until this PR lands first

mkruisselbrink:
> What I should have asked about sooner is whether we have any tests around this already and if any implementation bugs need to be filed? Sorry for the additional delay, but I'd like to be sure we don't have to revisit this anytime soon.

As mentioned in the initial pull request message:

> The only observable change here is the origin for blob URLs created in unique origins (which is only observable indirectly by trying to fetch such URLs), and tests for that already exist in https://w3c-test.org/FileAPI/url/sandboxed-iframe.html.

And from looking at https://wpt.fyi/results/FileAPI/url/sandboxed-iframe.html?label=master&label=stable&aligned there are probably some implementation bugs that should be filed if they haven't been filed already, although not clear how much that is related to this PR (which really was just trying to codify what browsers are already doing today). I.e. tests like "Blob URLs can be used in XHR and "Blob URLs can be used in fetch" passing in all browsers show that at least in some cases they use the correct origin. But then "Blob URLs can be used in iframes, and are treated same origin" fails in all browsers, so I'll have to double check what's going on there.

So the test situation is a bit tricky since it is only possible to indirectly observe the origin of a opaque origin'ed blob URL, but I think for this PR we're fine, all implementations at least do the right thing for fetch/xhr. There's definitely more work to be done here though, for example as part of whatwg/fetch#666, which will help address when exactly same/cross origin blob URLs should actually load or not load. As part of that I expect I'll be writing more tests, and probably filing implementation bugs. Does that sound reasonable?

annevk:
Sounds good to me, appreciate it!

PR #382, opened 2018-04-25
Remove _charset_ handling.

tkent-google:
We don't need to handle _charset_ in application/x-www-form-urlencoded
serializer because it should be done in "construct the form data set"
steps in the HTML specification.

This fixes https://github.com/whatwg/html/issues/3560.


tkent-google:
HTML spec: https://github.com/whatwg/html/pull/3645
WPT: https://github.com/w3c/web-platform-tests/pull/10623

annevk:
Thanks, I pushed a small fixup, but this looks good to me to land when the other PRs are ready.

Do you want to add yourself to the Acknowledgments section?

PR #384, opened 2018-05-01
Remove 'type' from from data entries.

tkent-google:
This fixes https://github.com/whatwg/html/issues/3648.

***
<a href="https://whatpr.org/url/384.html" title="Last updated on May 8, 2018, 9:05 AM GMT (eeea22e)">Preview</a> | <a href="https://whatpr.org/url/384/5c0d2ec...eeea22e.html" title="Last updated on May 8, 2018, 9:05 AM GMT (eeea22e)">Diff</a>

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -2510,8 +2510,8 @@ override</var>, and then runs these steps:
  <li>Return <var>output</var>.
 </ol>
 
-<p class="note no-backref">The <cite>HTML standard</cite> invokes this algorithm with name-value
-tuples. [[HTML]]
+<p class="note no-backref">The <cite>HTML standard</cite> invokes this algorithm with values that are files.
+[[HTML]]</p>
 
 
 <h3 id=urlencoded-hooks>Hooks</h3>
--- a/url.bs
+++ b/url.bs
@@ -2475,8 +2475,8 @@ takes a byte sequence <var>input</var> and then runs these steps:
 
 <p>The
 <dfn export id=concept-urlencoded-serializer lt='urlencoded serializer'><code>application/x-www-form-urlencoded</code> serializer</dfn>
-takes a list of name-value or name-value tuples <var>tuples</var>, optionally with an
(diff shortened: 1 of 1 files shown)


tkent-google:
> You'll also want to update this note:
> > The HTML standard invokes this algorithm with name-value-type tuples.

Done.

> And if we can indeed change HTML to always have File objects we should change the conditional a little further to check for that.

I'd like to focus on removing 'type' now.

url.bs
```diff
@@ -2475,7 +2475,7 @@ takes a byte sequence <var>input</var> and then runs these steps:
 
 <p>The
 <dfn export id=concept-urlencoded-serializer lt='urlencoded serializer'><code>application/x-www-form-urlencoded</code> serializer</dfn>
-takes a list of name-value or name-value-type tuples <var>tuples</var>, optionally with an
+takes a list of name-value or name-value tuples <var>tuples</var>, optionally with an
```
rmisev: name-value or name-value?

url.bs
```diff
@@ -2511,8 +2510,8 @@ takes a list of name-value or name-value-type tuples <var>tuples</var>, optional
  <li>Return <var>output</var>.
 </ol>
 
-<p class="note no-backref">The <cite>HTML standard</cite> invokes this algorithm with
-name-value-type tuples. [[HTML]]
+<p class="note no-backref">The <cite>HTML standard</cite> invokes this algorithm with name-value
+tuples. [[HTML]]
```
TimothyGu: You might want to instead say "The HTML standard invokes this algorithm with values that are files" or something similar.

url.bs
tkent-google: Oops!  Fixed.

url.bs
tkent-google: Done.

annevk:
Thanks all!

PR #386, opened 2018-05-09
Change query state slightly to better deal with non-UTF-8 encodings

annevk:
If the input to the URL parser contains code points outside the non-UTF-8 encoding's value space and the URL parser was invoked using a non-UTF-8 encoding, then those code points end up as &#...;.

The problem is that &, #, and ; are also URL separators, but the previous algorithm would only encode #. This ensures that & and ; are also encoded, as some browsers already do, but only if they came about as the result of the encode operation.

Tests: [we need to make a number of test changes for this]

***
<a href="https://whatpr.org/url/386.html" title="Last updated on May 23, 2018, 6:41 AM GMT (2518aa4)">Preview</a> | <a href="https://whatpr.org/url/386/acc2bf4...2518aa4.html" title="Last updated on May 23, 2018, 6:41 AM GMT (2518aa4)">Diff</a>

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -2116,34 +2116,12 @@ string <var>input</var>, optionally with a <a>base URL</a> <var>base</var>, opti
       <p>then set <var>encoding</var> to <a>UTF-8</a>.
       <!-- https://simon.html5.org/test/url/url-encoding.html -->
 
-     <li>
-      <p>If <a>c</a> is the <a>EOF code point</a>, or <var>state override</var> is not given and
-      <a>c</a> is U+0023 (#), then:
-
-      <ol>
-       <li><p>Set <var>buffer</var> to the result of <a lt=encode>encoding</a> <var>buffer</var>
-       using <var>encoding</var>.
-
-       <li>
-        <p>For each <var>byte</var> in <var>buffer</var>:
-
-        <ol>
-         <li><p>If <var>byte</var> is less than 0x21 (!), greater than 0x7E (~), or is 0x22 ("),
-         0x23 (#), 0x3C (&lt;), or 0x3E (>), append <var>byte</var>,
-         <a lt="percent encode">percent encoded</a>, to <var>url</var>'s <a for=url>query</a>.
-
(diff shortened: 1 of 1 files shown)


annevk:
This ends up fixing https://github.com/whatwg/encoding/issues/139. https://bugs.chromium.org/p/chromium/issues/detail?id=795733 can be closed.

cc @rakuco @hsivonen @valenting @domenic

annevk:
cc @inexorabletash

url.bs
```diff
@@ -2106,47 +2106,64 @@ string <var>input</var>, optionally with a <a>base URL</a> <var>base</var>, opti
...
+       <a>ASCII hex digits</a>, <a>validation error</a>.
 
-       <li><p>Set <var>buffer</var> to the result of <a lt=encode>encoding</a> <var>buffer</var>
-       using <var>encoding</var>.
+       <li><p>Let <var>bytes</var> be the result of <a lt=encode>encoding</a> <a>c</a> using
+       <var>encoding</var>.
 
        <li>
-        <p>For each <var>byte</var> in <var>buffer</var>:
+        <p>If <var>bytes</var> starts with 0x26 (&amp;) 0x23 (#) and ends with 0x3B (;), then:
```
annevk: Maybe this should be "starts with \``&#`\`" for consistency.

annevk:
I'm somewhat hoping this doesn't affect Node.js (I think it should always use UTF-8), but please verify @jasnell.

TimothyGu:
In Node.js URL parsing always uses UTF-8, that's correct.

annevk:
Thanks @rmisev, adopted your suggestion with some slight tweaks to the wording.

annevk:
Given that this now has had review, browser bugs are filed, and test changes have been reviewed as well, I plan on landing this tomorrow unless I hear concerns before that time.

It's blocking further test refactoring efforts somewhat so it'd be good to get this over with.

annevk:
Yeah, though I wouldn't mind non-UTF-8-performance penalties.

PR #387, opened 2018-05-11
Meta: Review Draft infrastructure

annevk:
See https://github.com/whatwg/whatwg.org/issues/197 for details.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@ This repository hosts the [URL Standard](https://url.spec.whatwg.org/).
 
 ### Code of conduct
 
-We are committed to providing a friendly, safe and welcoming environment for all. Please read and
+We are committed to providing a friendly, safe, and welcoming environment for all. Please read and
 respect the [WHATWG Code of Conduct](https://whatwg.org/code-of-conduct).
 
 ### Contribution opportunities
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -22,12 +22,7 @@ reading through the [WHATWG FAQ](https://whatwg.org/faq) if you are new here.
 Please add your name to the Acknowledgments section in your first pull request, even for trivial
 fixes. The names are sorted lexicographically.
 
-If you want to preview the spec locally, you can either use a locally installed copy of
-[Bikeshed](https://github.com/tabatkins/bikeshed) by running `make` or use the HTTP API version by
-running `make remote`.
-
-If you want to do a complete "local deploy" including commit and/or branch snapshots, run
-`make deploy`.
+If you want to verify your changes before submitting them to GitHub, run `make deploy`.
```
domenic: I feel not great about recommending people do full `make deploy` with its three curlbikesheds; I'd rather continue recommending `make remote` as a quicker flow for local iteration.

domenic:
I think we should add a pointer to whatwg/meta CONTRIBUTING.md that explains the build process in a bit more detail, including things like "where do I get Make on Windows" or "how can I build while I'm on an airplane" (by installing Bikeshed). I'll work on that and the corresponding whatwg/meta PR today.

annevk:
@foolip remarked that we should make `make remote` the default, which we seem to have agreement on, as long as `make local` remains available.

PR #401, opened 2018-07-14
Add official IANA assigned FTPS (FTP over TLS) port

ExE-Boss:
**FTPS (FTP over TLS)** was specified in [RFC 4217](https://tools.ietf.org/html/rfc4217) by IETF, and officially assigned the ports `990` (command/​control) and `989` (data) by IANA soon after.

Because of the TLS everything process that is going on among browser vendors, I feel like adding FTPS to the URL specification is a good addition.

***
<a href="https://whatpr.org/whatwg-url/401.html" title="Last updated on Jul 14, 2018, 5:26 AM GMT (3b3791a)">Preview</a> | <a href="https://whatpr.org/url/401/6ef17eb...3b3791a.html" title="Last updated on Jul 14, 2018, 5:26 AM GMT (3b3791a)">Diff</a>

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -1140,6 +1140,7 @@ corresponding <a for=url>port</a> and is listed in the second column on the same
  <tr><th><a for=url>scheme</a>
      <th><a for=url>port</a>
  <tr><td>"<code>ftp</code>"<td>21
+ <tr><td>"<code>ftps</code>"<td>990
  <tr><td>"<code>file</code>"<td>
  <tr><td>"<code>gopher</code>"<td>70
  <tr><td>"<code>http</code>"<td>80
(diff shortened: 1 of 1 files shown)


annevk:
Thanks for contributing this, but I don't think we want to make this change. As far as I know implementations do not parse `ftps:` URLs the same way as `ftp:` URLs. They're parsed per the non-special URL rules. Did you find differently?

annevk:
I'm pretty sure that's no longer the case. It's also not entirely clear to me we'd want to change the URL parser for that. It seems it could be supported as a non-special URL just as well.

annevk:
Closing this per the above comments. Happy to reconsider once implementers are on board.

Issue #24, opened 2015-05-27
Grammar specification for URLs


ericedens:
Any thoughts on providing a formal grammar? The current spec defines URLs in terms of algorithms - while that's helpful to quickly write an implementation in an arbitrary language, it makes it tough to write code generators that will update parsers as the spec evolves.

ericedens:
To clarify, this isn't a nagging feature request that's asking you to do more work -- I'm happy to help write a grammar :)

More specifically, I was surprised to not see a grammar, and I'm wondering whether that was an intentional omission, or whether it's just something that's low priority.

domenic:
I don't remember the details exactly but I believe there are several large parts of the spec that cannot be captured by any kind of grammar.

This spec https://specs.webplatform.org/url/webspecs/develop/#parsing-rules attempts to codify URLs as a grammar (via railroad diagrams) but almost every grammar production also needs some Turing-complete post-processing steps---every step that doesn't say "Parse input according to the above railroad diagram" is outside the grammar.

rubys:
Also note the 'note' at the end of the section that @domenic pointed.

> Note: extracting the railroad diagrams from this specification and interpreting them in isolation will produce incorrect results. In particular, the definitions provided in the prose modifies these parsing algorithms in important ways including returning failure and early termination.

annevk:
@ericedens here are some reasons to avoid grammar:

1. Encourages independently written code. This catches bugs and leads to a better overall ecosystem. (Note how many grammar-based RFCs have not resulted in many compliant implementations.)
2. It cannot actually be done.

Feel free to suggest a grammar, but I don't think we'll ever have one that's considered normative.

ericedens:
@annevk , a problem that I've seen with other specs is that they provide a single grammar. If the grammar is strict enough to produce valid URLs, then it won't accept real-world URLs. If it accepts real-world URLs, then it will produce non-compliant URLs.

A better approach would be to have two grammars. One that generates compliant URLs, and one that accepts real-world URLs.

A good example is  '\' vs '/'.  The strict grammar would only allow '/'; the looser grammar would allow either. A parser that used the loose grammer would be responsible for converting '\' to '/' after the parse tree has been created.

> 2. It cannot actually be done.

Sounds like a challenge to me :)

domenic:
Unless you come up with a Turing-complete grammar specification language,  then no matter how many grammars you have (two,  five,  whatever) it will not be enough. And we usually call such languages "programming languages".

[...]

PR #416, opened 2018-09-16
Restore formal grammar

sjamaan:
As per #24 I would really like to see a formal grammar in the URL spec.

This is an initial start on such a grammar. I've taken the ABNF from RFC3986.  As far as I can tell, the WHATWG URL spec has removed support for the `ipvfuture` production, so I've ripped that out.  I've also **not** added support for IPv6 zone identifiers from RFC6874, as the WHATWG spec explicitly states

> Support for `<zone_id>` is intentionally omitted.

I'm not sure how to best proceed to add error handling. I think it's probably best to first ensure that the syntax described by the ABNF matches the syntax accepted by the algorithm (without any errors or warnings). Where the algorithm accepts a URL that is not supported by the ABNF, the ABNF could be updated if it the algorithm would not raise a validation error. If there are URLs that will be accepted by the ABNF, the algorithm probably needs to be updated, unless it that syntax is intentionally no longer supported.

***
<a href="https://whatpr.org/url/416.html" title="Last updated on Jan 15, 2021, 7:42 AM UTC (ccadabf)">Preview</a> | <a href="https://whatpr.org/url/416/85d96e2...ccadabf.html" title="Last updated on Jan 15, 2021, 7:42 AM UTC (ccadabf)">Diff</a>

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -58,6 +58,7 @@ might increase in scope somewhat.
  <li>Media Source Extensions [[!MEDIA-SOURCE]]
  <li>Unicode IDNA Compatibility Processing [[!UTS46]]
  <li>Web IDL [[!WEBIDL]]
+ <li>ABNF [[!STD68]]
 </ul>
 
 <hr>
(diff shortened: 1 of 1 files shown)


url.bs
```diff
@@ -2346,6 +2347,105 @@ string <var>input</var>, optionally with a <a>base URL</a> <var>base</var>, opti
...
 
+<h4 id=grammar>Formal grammar for minimally acceptable URLs</h4>
+
+<p>It is the intention that the above parsing algorithm accepts a
+superset of the language defined by the following grammar.  The
+grammar is only defined for UTF-8.  The <var>encoding override</var>
+option is not supported.  Resolving the URL against a <a>base URL</a>
+or raising an error if the URL is relative should be done after
+parsing.</p>
+
```
GPHemsley: All text within the specification is to be considered as normative. However, this sounds more like a note to me than a normative requirement.

url.bs
```diff
@@ -2346,6 +2347,105 @@ string <var>input</var>, optionally with a <a>base URL</a> <var>base</var>, opti
...
+<p>This specification uses the Augmented Backus-Naur Form (ABNF)
+notation of STD 68 [[!STD68]], including the following core ABNF
+syntax rules defined by that specification: ALPHA (letters), CR
+(carriage return), DIGIT (decimal digits), DQUOTE (double quote),
+HEXDIG (hexadecimal digits), LF (line feed), and SP (space).</p>
+
```
GPHemsley: I think it's generally preferred that terms defined in other standards are linked to their definitions in those other standards.

url.bs
```diff
@@ -2346,6 +2347,105 @@ string <var>input</var>, optionally with a <a>base URL</a> <var>base</var>, opti
...
+   pct-encoded   = "%" HEXDIG HEXDIG
+
+   unreserved    = ALPHA / DIGIT / "-" / "." / "_" / "~"
+   reserved      = gen-delims / sub-delims
+   gen-delims    = ":" / "/" / "?" / "#" / "[" / "]" / "@"
+   sub-delims    = "!" / "$" / "&" / "'" / "(" / ")"
+                 / "*" / "+" / "," / ";" / "="
+</pre>
+
```
GPHemsley: I think more needs to be done here to make this useful than to just drop a huge chunk of preformatted text into the spec. Also, I think wpt would need to be updated to somehow test this grammar as distinct from the parser algorithm.

url.bs
sjamaan: How should I reword it to make it normative?

url.bs
sjamaan: Regarding the preformatted text, I wouldn't mind polishing it to make it formatted in a way that makes it more readable. Would that involve adding BNF tags to bikeshed, or are there suitable options already?

Testing a grammar would be kind of tricky. Usually one would pick a parser generator and make the necessary transformations to make the BNF fit that particular parser. I have several different implementations of the RFC using various parser generators in [my project's repository](https://bugs.call-cc.org/browser/project/release/5/uri-generic/trunk/alternatives) but if you inspect them they all involve some modifications. Funny enough, the [one based on SRE](https://bugs.call-cc.org/browser/project/release/5/uri-generic/trunk/alternatives/uri-generic.irregex.scm#L503) (a composable form of regular expressions) is the only one without *any* changes from the BNF, IIRC.

We could include that SRE-based implementation in WPT. If only JavaScript is accepted as an implementation language I can take a look at the ecosystem to find a good parser generator and try my hand at transcoding the BNF to it. I'm not a good JavaScript coder, though!

url.bs
sjamaan: I copied this from the RFC. I did link to the ABNF STD. How would you recommend I change this?

sjamaan:
hm, I just clicked "resolve conversation". I'm not sure that's the correct usage of that button. I thought that would actually submit the comments but it looks like it just closes the "subthread". Sorry about that!

PR #428, opened 2019-02-07
Be clearer that URL serialization returns an ASCII string.

jyasskin:
I added this fact in the wrong place in #409: it ought to be on the definition of the [URL serializer](https://url.spec.whatwg.org/#concept-url-serializer) and not just [one of the uses](https://url.spec.whatwg.org/#ref-for-concept-url-serializer①). (I care because I just wasted a bit of time double-checking that it was true.)

I don't feel strongly about the `Assert` at the bottom of the algorithm.

***
<a href="https://whatpr.org/url/428.html" title="Last updated on Feb 7, 2019, 9:05 PM UTC (4d5ca88)">Preview</a> | <a href="https://whatpr.org/url/428/d2ef633...4d5ca88.html" title="Last updated on Feb 7, 2019, 9:05 PM UTC (4d5ca88)">Diff</a>

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -2347,7 +2347,7 @@ string <var>input</var>, optionally with a <a>base URL</a> <var>base</var>, opti
 
 <p>The <dfn export id=concept-url-serializer lt="URL serializer">URL serializer</dfn> takes a
 <a for=/>URL</a> <var>url</var>, an optional <i title>exclude fragment flag</i>, and
-then runs these steps:
+then runs these steps, returning an <a>ASCII string</a>:
 
 <ol>
  <li><p>Let <var>output</var> be <var>url</var>'s <a for=url>scheme</a> and U+003A (:) concatenated.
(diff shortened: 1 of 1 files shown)


annevk:
So we clarified domain to be an ASCII string which means all pieces of a URL are now either an ASCII string or an object of some kind.

This is somewhat different however. We generally don't list return types and we generally don't assert their types before returning them. This might be a case where it's especially useful though.

I wonder if @domenic has any Infra insights/thoughts on this.

domenic:
This seems pretty reasonable to me. I agree we don't usually list return types, and I might lean toward removing that and just keeping the assertion, but I don't mind much either way. I think any assertion that adds clarity is welcome; I wouldn't make a habit of asserting before every return, since it's often redundant, but I can see how this particular case is of interest and merits it.

The Infra connection would be https://github.com/whatwg/infra/issues/92.

jyasskin:
I think we ought to specify all return types that are produced by a sequence of steps, so that people calling algorithms can tell what they're getting out without needing to read the implementations. Doing this gets more important the longer the algorithm is, and the more sub-algorithms it calls that don't themselves describe their return types.

I think the assertion is less useful to readers than the up-top declaration, because when reading the assertion, I also have to check for any early returns that might return a different type. The assertion is still more useful than nothing.

Infra, if it defines the pattern for declaring return types, could say that every return statement implicitly includes an assertion that the types match the ones declared for the algorithm.

annevk:
That makes sense for pure functions, I think. I'm not sure I particularly like the style, but the only alternative I have is to give up on formulating it as a sentence and create some kind of structure for an algorithm header.

So, if you remove the assert I'll merge this. And we'll just evaluate over time how useful this kind of thing is and where to take it next.

jyasskin:
Done, thanks.

PR #434, opened 2019-03-15
Restructure URL rendering section and add additional guidance

estark37:
This commit expands the URL rendering section to contain some of the
non-Chrome-specific guidelines from Chromium's URL display guidelines
(https://chromium.googlesource.com/chromium/src/+/master/docs/security/url_display_guidelines/url_display_guidelines.md).
The guidance is geared towards URLs that are rendered primarily in a
context in which a user is making a security decision.

The advice is separated into 3 sections: simplifying URLs to prevent
spoofing and confusion, eliding in space-constrained displays, and
i18n/bidi/special characters.

***
<a href="https://whatpr.org/url/434.html" title="Last updated on Mar 27, 2019, 10:30 AM UTC (397288e)">Preview</a> | <a href="https://whatpr.org/url/434/c4f729a...397288e.html" title="Last updated on Mar 27, 2019, 10:30 AM UTC (397288e)">Diff</a>


estark37:
@annevk As we briefly discussed over email, this is an attempt to incorporate some of Chromium's general URL display guidelines into the URL rendering section. I'd be happy to open an issue to discuss first if that's preferred -- just thought it might be easier to have a specific proposal to work from.

cc @mikewest @emilyschechter

url.bs
```diff
@@ -2476,39 +2476,83 @@ background information. [[!HTML]]
...
+ <code>https://examplecorp.com@attacker.example/</code>).
 
- <li><p>Other parts of the <a for=/>URL</a> should have their sequences of
- <a>percent-encoded bytes</a> replaced with code points resulting from
- <a>percent decoding</a> those sequences converted to bytes, unless that renders those
- sequences invisible.
+ <li><p>A URL can be rendered without its <a for=url>scheme</a> if the display surface only ever
+ permits a single scheme (such as a browser feature that omits <code>https://</code> because it is
+ only enabled for secure origins). Otherwise, the scheme may be replaced or supplemented with a
+ human-readable string (e.g., "Not secure") and/or a security indicator icon.
```
annevk: Replacing "and/or" with a comma and tacking ", or both" at the end seems slightly preferable (and I suspect the current wording doesn't get past our tooling).

url.bs
```diff
@@ -2476,39 +2476,83 @@ background information. [[!HTML]]
...
+
+ <li><p>A leading <code>www</code> domain label can be omitted to simplify the host, except when it
+ is part of the <a for=host>registrable domain</a>.
```
annevk: This one led to some issues right? It seems this can also be problematic if the site isn't reachable without it.

url.bs
```diff
@@ -2476,39 +2476,83 @@ background information. [[!HTML]]
...
+
+ <li><p>Omit the port if it is a <a>default port</a>.
```
annevk: For those schemes the URL's port will be null in those cases, so I guess we want to say to not serialize the port when it's null.

url.bs
```diff
@@ -2476,39 +2476,83 @@ background information. [[!HTML]]
...
+making a security decision:
+
+<ul class=brief>
+ <li><p>Ensure that at least the <a for=host>registrable domain</a> can be shown when the URL is
+ rendered (to avoid showing, e.g., <code>...examplecorp.com</code> when loading
+ <code>https://not-really-examplecorp.com/</code>).
+
+ <li><p>When the full <a for=url>host</a> cannot be rendered, elide domain labels starting from the
+ front. (Note that bidirectional text means that the front of the string may not appear at the
+ left.)
```
annevk: I think this might benefit from one or two examples, in particular because how it's different from the previous point is a little subtle.

url.bs
```diff
@@ -2476,39 +2476,83 @@ background information. [[!HTML]]
...
+ <p class="note no-backref">Unfortunately, as rendered <a for=/>URLs</a> are strings and can appear
+ anywhere, a specific bidirectional algorithm for rendered <a for=/>URLs</a> would not see wide
+ adoption. Bidirectional text interacts with the parts of a <a for=/>URL</a> in ways that can cause
+ the rendering to be different from the model. Users of bidirectional languages are thus cautioned
+ that this is to be expected, particularly in plain text environments.
+
+ <li><p>Other parts of the <a for=/>URL</a> should have their sequences of
+ <a>percent-encoded bytes</a> replaced with code points resulting from
+ <a>percent decoding</a> those sequences converted to bytes, unless that renders those
+ sequences invisible.
```
annevk: This is primarily about the path right? So maybe we should combine this with the bit two points above whereby we encourage only to render the host?

url.bs
```diff
@@ -2476,39 +2476,83 @@ background information. [[!HTML]]
...
+modifications described below when the primary purpose of displaying a URL is to have the user make
+a security decision (e.g., users are expected to make trust decisions based on a URL rendered in the
+browser address bar).
+
+<h4 id=url-rendering-simplification>Simplify non-human-readable or irrelevant components</h4>
+
+<p>Remove components that may provide opportunities for spoofing or distract from security-relevant
+information:
 
 <ul class=brief>
```
domenic: Removing the `class=brief` from these `<ul>`s would make the result much prettier, I expect.

url.bs
estark37: Done

url.bs
estark37: The main issue was a bug in the implementation, which has now been fixed. We are planning on re-landing the change in Chrome and so it would be great to have it documented here.

Chrome's implementation unelides and shows the full domain when you copy/select/edit the URL, which is intended to minimize confusion if the site is not reachable without 'www'. We also did an analysis of top sites and found that it's uncommon to serve different content at www vs bare domain, or have one be reachable but not the other. (Only 2% of Alexa top 500 hosts serve different content at www vs bare domain.)

Finally, I think this guidance is also somewhat in line with what Safari does: they don't elide www specifically, but they show the registrable domain only. The underlying reasoning is the same -- it's okay to show a subset of the actual host in service of helping users understand the site identity better. (In fact, one possibility might be to fold this bullet in with the first bullet and phrase it more as "Browsers are encouraged to render a URL's host only. Browsers may further consider rendering only the most relevant parts of a host, such as omitting a leading 'www' domain label or rendering the registrable domain only.") WDYT?

url.bs
estark37: Done

url.bs
estark37: Done

url.bs
estark37: Done

url.bs
estark37: Done

estark37:
Thanks for the review @annevk and @domenic! Pushed a new commit to address your comments.

url.bs
```diff
@@ -2476,39 +2476,81 @@ background information. [[!HTML]]
...
 
-<p>Due to the confusion that can arise between a <a for=/>URL</a>'s <a for=url>host</a>
-and <a for=url>path</a> with bidirectional text, browsers are encouraged to only render a
-<a for=/>URL</a>'s <a for=url>host</a> in places where it is important for users to
-distinguish between the two. E.g., users are expected to make trust decisions based on a
-<a for=/>URL</a>'s <a for=url>host</a> rendered in the address bar.
+ <li><p>URLs are particularly prone to confusion between host and path when they contain
+ bidirectional text, so in this case it is particularly advisable to only render a URL’s <a
+ for=url>host</a>. Other parts of the <a for=/>URL</a>, if rendered, should have their sequences of
+ <a>percent-encoded bytes</a> replaced with code points resulting from <a>percent decoding</a> those
```
jyasskin: I probably haven't thought about this enough, but I've lost track of how replacing %-encoded bytes with their decodings help disambiguate between the host and path. An example might help?

url.bs
```diff
@@ -2476,39 +2476,81 @@ background information. [[!HTML]]
...
- <li><p>A <a for=/>URL</a>'s <a for=url>username</a> and <a for=url>password</a> should
- not be rendered as they can be mistaken for a <a for=/>URL</a>'s <a for=url>host</a>.
- E.g., consider <code>https://examplecorp.com@attacker.example/</code>.
+<h4 id=url-rendering-simplification>Simplify non-human-readable or irrelevant components</h4>
+
+<p>Remove components that may provide opportunities for spoofing or distract from security-relevant
+information:
+
+<ul>
+ <li><p>Browsers are encouraged to only render a URL’s <a for=url>host</a> in places where it is
```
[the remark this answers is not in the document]
estark37: Done

url.bs
```diff
@@ -2476,39 +2476,83 @@ background information. [[!HTML]]
 <h3 id=url-rendering>URL rendering</h3>
 <!-- See https://www.w3.org/Bugs/Public/show_bug.cgi?id=27641 for context -->
 
-<p>A <a for=/>URL</a> should be rendered in its <a lt="URL serializer">serialized</a>
-form, with these modifications:
+<p>A <a for=/>URL</a> should be rendered in its <a lt="URL serializer">serialized</a> form, with
+modifications described below when the primary purpose of displaying a URL is to have the user make
+a security decision (e.g., users are expected to make trust decisions based on a URL rendered in the
```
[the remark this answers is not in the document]
estark37: Done

url.bs
```diff
@@ -2476,39 +2476,81 @@ background information. [[!HTML]]
...
 </ul>
 
-<p>For the purposes of bidirectional text it should be rendered as if it were in a
-left-to-right embedding. [[!BIDI]]
+<h4 id=url-rendering-i18n>Internationalization and special characters</h4>
+
+<p>International domain names (IDNs), special characters, and bidirectional text should be handled
+with care to prevent spoofing:
+
+<ul>
```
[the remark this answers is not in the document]
estark37: Done

url.bs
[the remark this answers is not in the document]
estark37: Done

url.bs
```diff
@@ -2476,39 +2476,81 @@ background information. [[!HTML]]
...
+ important for users to distinguish between the host and other parts of the URL such as the <a
+ for=url>path</a>. Browsers may further consider rendering only the URL’s host's <a
+ for=host>registrable domain</a> to remove spoofing opportunities posed by subdomains (e.g.,
+ <code>https://examplecorp.attacker.com/</code>).
+
+ <li><p>A <a for=/>URL</a>'s <a for=url>username</a> and <a for=url>password</a> should not be
+ rendered as they can be mistaken for a <a for=/>URL</a>'s <a for=url>host</a> (as in, e.g.,
+ <code>https://examplecorp.com@attacker.example/</code>).
+
+ <li><p>A URL can be rendered without its <a for=url>scheme</a> if the display surface only ever
```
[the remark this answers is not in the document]
estark37: Done

url.bs
```diff
@@ -2476,39 +2476,81 @@ background information. [[!HTML]]
...
+<p>In a space-constrained display, URLs should be elided carefully to avoid misleading the user when
+making a security decision:
+
+<ul>
+ <li><p>Ensure that at least the <a for=host>registrable domain</a> can be shown when the URL is
+ rendered (to avoid showing, e.g., <code>...examplecorp.com</code> when loading
+ <code>https://not-really-examplecorp.com/</code>).
+
+ <li><p>When the full <a for=url>host</a> cannot be rendered, elide domain labels starting from the
+ front. For example, <code>examplecorp.com.evil.com</code> should be elided as
```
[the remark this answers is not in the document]
estark37: I changed this to "lowest-level label", do you think that helps? I'm not sure what the standard term is, if any...

url.bs
estark37: This bit was in the original section, so I'm not sure what the original motivation was. My motivation for keeping it is that we should avoid having URLs that are so syntax-y that users don't even bother trying to understand them. I prepended "For readability," to this sentence to try to make that clearer. Do you think that helps? (I also added a note that some characters may not be decoded if they present spoofing risks, e.g. Unicode lock.)

estark37:
Thanks @jyasskin! Comments addressed.

url.bs
```diff
@@ -2476,39 +2476,84 @@ background information. [[!HTML]]
 <h3 id=url-rendering>URL rendering</h3>
 <!-- See https://www.w3.org/Bugs/Public/show_bug.cgi?id=27641 for context -->
 
-<p>A <a for=/>URL</a> should be rendered in its <a lt="URL serializer">serialized</a>
-form, with these modifications:
+<p>A <a for=/>URL</a> should be rendered in its <a lt="URL serializer">serialized</a> form, with
+modifications described below when the primary purpose of displaying a URL is to have the user make
```
sideshowbarker: Adding a comma after “below” in this sentence would improve readability

url.bs
```diff
@@ -2476,39 +2476,84 @@ background information. [[!HTML]]
...
- <li><p>A <a for=/>URL</a>'s <a for=url>host</a> should be rendered using
- <a>domain to Unicode</a>.
+<p>Remove components that may provide opportunities for spoofing or distract from security-relevant
+information:
+
+<ul>
+ <li><p>Browsers may render only a URL’s <a for=url>host</a> in places where it is important
+ for users to distinguish between the host and other parts of the URL such as the <a
+ for=url>path</a>. Browsers may further consider rendering only the URL’s host's <a
+ for=host>registrable domain</a> to remove spoofing opportunities posed by subdomains (e.g.,
```
sideshowbarker: Suggest comma after “registrable domain”

url.bs
```diff
@@ -2476,39 +2476,84 @@ background information. [[!HTML]]
...
+ bidirectional text means that the lowest-level label may not appear at the left.)
+</ul>
 
+<h4 id=url-rendering-i18n>Internationalization and special characters</h4>
 
+<p>International domain names (IDNs), special characters, and bidirectional text should be handled
+with care to prevent spoofing:
+
+<ul>
+ <li><p>Browser should render a <a for=/>URL</a>'s <a for=url>host</a> using
```
sideshowbarker: Should be “Browsers” (plural), for consistency

url.bs
```diff
@@ -2476,39 +2476,84 @@ background information. [[!HTML]]
...
 
-<ul class=brief>
- <li><p>A <a for=/>URL</a>'s <a for=url>username</a> and <a for=url>password</a> should
- not be rendered as they can be mistaken for a <a for=/>URL</a>'s <a for=url>host</a>.
- E.g., consider <code>https://examplecorp.com@attacker.example/</code>.
+<h4 id=url-rendering-simplification>Simplify non-human-readable or irrelevant components</h4>
 
- <li><p>A <a for=/>URL</a>'s <a for=url>host</a> should be rendered using
- <a>domain to Unicode</a>.
+<p>Remove components that may provide opportunities for spoofing or distract from security-relevant
```
sideshowbarker: Suggest “components that can provide opportunities” (“can” instead “may”), for consistency with WHATWG style that, which avoids use of “may” in any non-normative sense.

url.bs
annevk: I cannot reproduce Safari only showing the registrable domain. E.g., for https://a.x.testsuite.org/ I see `a.x.testsuite.org`.

estark37: Ah interesting, I thought they were but it looks like it's more complicated than that... they at least omit `www` (see screenshot on `www.google.com`).

image
image

Maybe @johnwilander knows the specifics.

estark37: Seems they omit `m` too:

image
image

url.bs
annevk: I guess it depends on how much we expect the path/query/fragment to show up going forward. It's definitely a little fragile to decode as the browser will later have to encode it again when going to the network. Differences there could result in issues (and have).

url.bs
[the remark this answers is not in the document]
annevk: https://tools.ietf.org/html/rfc1034#section-3.1 has some wording. The URL Standard hadn't really introduced any wording around labels as thus far it wasn't needed for anything normative. Perhaps it should.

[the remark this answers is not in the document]
sleevi: https://tools.ietf.org/html/rfc8499 is meant to be the informative "all terms go here" doc. I suspect it's something like "beginning with the labels furthest from the DNS root" or something h'awkward like that ;)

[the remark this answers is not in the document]
annevk: Skimming both a bit I guess strictly speaking this means we need to define something ourselves, as we don't talk about byte-based labels in the URL Standard.

https://tools.ietf.org/html/rfc5890#section-2.3.2.1 might have some building blocks, though nothing great. Anyway, this seems like a follow-up.

url.bs
estark37: Done

url.bs
estark37: Done

url.bs
estark37: Done

url.bs
estark37: @annevk are you okay leaving this in here given that Safari does it and Chrome is planning too as well?

url.bs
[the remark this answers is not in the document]
estark37: Alright, I'm happy to file an issue and follow up separately. Anyone have strong feelings about the options currently on the table ("front", "lowest-level label", "label furthest from the DNS root") in the meantime?

url.bs
estark37: FWIW I don't expect path/query/fragment from all URL display surfaces. They'll still be displayed when editing URLs in the address bar, and in secondary UI like bookmarks and history, etc. It probably makes sense to continue advising to decode in contexts like those.

jyasskin: It might help to put "Other parts of the URL …" in a separate `<li>` since it's a good idea but doesn't help with de-confusing the host and path. I don't feel strongly about this.

url.bs
annevk: Yeah, I wonder if we should generalize it given the `m` case you discovered.

estark37: Ok, that seems reasonable. I collapsed it into the first bullet and made it about simplifying the host generally, with omitting `www` or `m` or displaying registrable domain only listed as examples.

url.bs
```diff
@@ -2476,39 +2476,82 @@ background information. [[!HTML]]
 <h3 id=url-rendering>URL rendering</h3>
 <!-- See https://www.w3.org/Bugs/Public/show_bug.cgi?id=27641 for context -->
 
-<p>A <a for=/>URL</a> should be rendered in its <a lt="URL serializer">serialized</a>
-form, with these modifications:
+<p>A <a for=/>URL</a> should be rendered in its <a lt="URL serializer">serialized</a> form, with
+modifications described below, when the primary purpose of displaying a URL is to have the user make
+a security decision. For example, users are expected to make trust decisions based on a URL rendered
```
annevk: Should we call it a security or trust decision? Or maybe instead we could say "of displaying a URL is to assist the user in making decisions".

url.bs
```diff
@@ -2476,39 +2476,82 @@ background information. [[!HTML]]
...
+ <li><p>Browsers should not render a <a for=/>URL</a>'s <a for=url>username</a> and <a
+ for=url>password</a>, as they can be mistaken for a <a for=/>URL</a>'s <a for=url>host</a> (as in,
+ e.g., <code>https://examplecorp.com@attacker.example/</code>).
+
+ <li><p>Browsers may render a URL without its <a for=url>scheme</a> if the display surface only ever
+ permits a single scheme (such as a browser feature that omits <code>https://</code> because it is
+ only enabled for secure origins). Otherwise, the scheme may be replaced or supplemented with a
+ human-readable string (e.g., "Not secure"), a security indicator icon, or both.
+
+ <li><p>As described in <a>URL serializer</a>, browsers should not serialize null ports.
```
annevk: Having rephrased it, do you still think this is worth keeping? It seems redundant to me at this point.

url.bs
```diff
@@ -2476,39 +2476,82 @@ background information. [[!HTML]]
...
+
+<ul>
+ <li><p>Browsers should ensure that at least the <a for=host>registrable domain</a> can be shown
+ when the URL is rendered (to avoid showing, e.g., <code>...examplecorp.com</code> when loading
+ <code>https://not-really-examplecorp.com/</code>).
+
+ <li><p>When the full <a for=url>host</a> cannot be rendered, browsers should elide domain labels
+ starting from the lowest-level domain label. For example, <code>examplecorp.com.evil.com</code>
+ should be elided as <code>...com.evil.com</code>, not <code>examplecorp.com...</code>. (Note that
+ bidirectional text means that the lowest-level label may not appear at the left.)
```
annevk: I'm not a native speaker, but "on the left" sounds more natural to me. Or "at the left side" perhaps.

url.bs
```diff
@@ -2476,39 +2476,82 @@ background information. [[!HTML]]
...
-the rendering to be different from the model. Users of bidirectional languages are thus cautioned
-that this is to be expected, particularly in plain text environments.
+<h4 id=url-rendering-i18n>Internationalization and special characters</h4>
 
-<p>Due to the confusion that can arise between a <a for=/>URL</a>'s <a for=url>host</a>
-and <a for=url>path</a> with bidirectional text, browsers are encouraged to only render a
-<a for=/>URL</a>'s <a for=url>host</a> in places where it is important for users to
-distinguish between the two. E.g., users are expected to make trust decisions based on a
-<a for=/>URL</a>'s <a for=url>host</a> rendered in the address bar.
+<p>International domain names (IDNs), special characters, and bidirectional text should be handled
```
annevk: Editorial (I can fix this before merging): "Internationalized domain name" seems to be the canonical expansion of this abbreviation.

url.bs
```diff
@@ -2476,39 +2476,82 @@ background information. [[!HTML]]
...
+with care to prevent spoofing:
 
+<ul>
+ <li><p>Browsers should render a <a for=/>URL</a>'s <a for=url>host</a> using
```
annevk: Editorial (I can fix this before merging): as this `<li>` contains multiple elements those need to be on their own lines.

url.bs
```diff
@@ -2476,39 +2476,82 @@ background information. [[!HTML]]
...
+
+ <li><p>URLs are particularly prone to confusion between host and path when they contain
+ bidirectional text, so in this case it is particularly advisable to only render a URL’s <a
+ for=url>host</a>. For readability, other parts of the <a for=/>URL</a>, if rendered, should have
+ their sequences of <a>percent-encoded bytes</a> replaced with code points resulting from <a>percent
+ decoding</a> those sequences converted to bytes, unless that renders those sequences
+ invisible. Browsers may choose to not decode certain sequences that present spoofing risks (e.g.,
+ the Unicode LOCK character U+1F512).
+
+ <li><p>Browsers should render bidirectional text as if it were in a left-to-right embedding. [[!BIDI]]
```
annevk: Editorial (I can fix this before merging): as this `<li>` contains multiple elements those need to be on their own lines.

url.bs
```diff
@@ -2476,39 +2476,82 @@ background information. [[!HTML]]
...
+ <li><p>Browsers should render a <a for=/>URL</a>'s <a for=url>host</a> using
+ <a>domain to Unicode</a>.
 
+ <p class="note no-backref">Note that non-ASCII characters can be used in <a
+ href="http://unicode.org/faq/idn.html#26">homograph</a> spoofing attacks. Consider detecting <a
+ href="http://www.unicode.org/reports/tr39/#Confusable_Detection">confusable characters</a> or
+ warning when they are in use.
+
+ <li><p>URLs are particularly prone to confusion between host and path when they contain
+ bidirectional text, so in this case it is particularly advisable to only render a URL’s <a
```
annevk: Editorial (I can fix this before merging): we use normal single quotes in the source.

url.bs
```diff
@@ -2476,39 +2476,82 @@ background information. [[!HTML]]
...
-distinguish between the two. E.g., users are expected to make trust decisions based on a
-<a for=/>URL</a>'s <a for=url>host</a> rendered in the address bar.
+<p>International domain names (IDNs), special characters, and bidirectional text should be handled
+with care to prevent spoofing:
 
+<ul>
+ <li><p>Browsers should render a <a for=/>URL</a>'s <a for=url>host</a> using
+ <a>domain to Unicode</a>.
 
+ <p class="note no-backref">Note that non-ASCII characters can be used in <a
```
annevk: Editorial (I can fix this before merging): no newlines inside inline elements. (Happens a few times.)

url.bs
```diff
@@ -2476,39 +2476,82 @@ background information. [[!HTML]]
...
+ invisible. Browsers may choose to not decode certain sequences that present spoofing risks (e.g.,
+ the Unicode LOCK character U+1F512).
+
+ <li><p>Browsers should render bidirectional text as if it were in a left-to-right embedding. [[!BIDI]]
+
+ <p class="note no-backref">Unfortunately, as rendered <a for=/>URLs</a> are strings and can appear
+ anywhere, a specific bidirectional algorithm for rendered <a for=/>URLs</a> would not see wide
+ adoption. Bidirectional text interacts with the parts of a <a for=/>URL</a> in ways that can cause
+ the rendering to be different from the model. Users of bidirectional languages are thus cautioned
+ that this is to be expected, particularly in plain text environments.
```
annevk: I wonder if we should rephrase this a bit as I doubt users read this document or take advice from it. Perhaps state it more as a matter of fact, that users will come to expect this or are expecting this.

url.bs
```diff
@@ -2476,39 +2476,82 @@ background information. [[!HTML]]
...
+ href="http://www.unicode.org/reports/tr39/#Confusable_Detection">confusable characters</a> or
+ warning when they are in use.
+
+ <li><p>URLs are particularly prone to confusion between host and path when they contain
+ bidirectional text, so in this case it is particularly advisable to only render a URL’s <a
+ for=url>host</a>. For readability, other parts of the <a for=/>URL</a>, if rendered, should have
+ their sequences of <a>percent-encoded bytes</a> replaced with code points resulting from <a>percent
+ decoding</a> those sequences converted to bytes, unless that renders those sequences
+ invisible. Browsers may choose to not decode certain sequences that present spoofing risks (e.g.,
+ the Unicode LOCK character U+1F512).
```
annevk: Editorial (I can fix this before merging): write this as "U+1F512 (🔒)" per Infra.

url.bs
```diff
@@ -2476,39 +2476,82 @@ background information. [[!HTML]]
...
+<p>International domain names (IDNs), special characters, and bidirectional text should be handled
+with care to prevent spoofing:
 
+<ul>
+ <li><p>Browsers should render a <a for=/>URL</a>'s <a for=url>host</a> using
+ <a>domain to Unicode</a>.
 
+ <p class="note no-backref">Note that non-ASCII characters can be used in <a
+ href="http://unicode.org/faq/idn.html#26">homograph</a> spoofing attacks. Consider detecting <a
+ href="http://www.unicode.org/reports/tr39/#Confusable_Detection">confusable characters</a> or
```
TimothyGu: Editorial: https for both of the links ;) We should also cite the links so that they show up in the Bibliography section, but doesn't look like either of them is in [Specref](https://www.specref.org/) yet…

sideshowbarker: > Editorial: https for both of the links ;) We should also cite the links so that they show up in the Bibliography section, but doesn't look like either of them is in [Specref](https://www.specref.org/) yet…

I’ve now added IDNFAQ and UNICODE-SECURITY-MECHANISMS refs to Specref

sideshowbarker: > I’ve now added IDNFAQ and UNICODE-SECURITY-MECHANISMS refs to Specref

I’ve also added UTR39 as an alias for UNICODE-SECURITY-MECHANISMS

gijsk: Should this be "and" rather than "or"? Just 'detecting' confusables does not imply anything about how that would affect the rendering of the URL. Chrome shows punycode in some of the cases where confusables show up, which arguably is also not really a "warning". Maybe "... and altering the URL rendering to avoid user confusion." ?

(To be clear, I don't want perfect to be the enemy of good here, or overly focus on the IDN issues, so just noting this problem and a loosely phrased suggestion seems fine, certainly as a first step.)

url.bs
annevk: I missed that this singles out non-ASCII. I don't think we should do that as it's Western-centric. E.g., formula1 vs formulaI vs formulal, kerning vs keming, or some such.

url.bs
estark37: Thank you! Done.

url.bs
```diff
@@ -2531,10 +2529,9 @@ with care to prevent spoofing:
  <li><p>Browsers should render a <a for=/>URL</a>'s <a for=url>host</a> using
  <a>domain to Unicode</a>.
 
- <p class="note no-backref">Note that non-ASCII characters can be used in <a
- href="http://unicode.org/faq/idn.html#26">homograph</a> spoofing attacks. Consider detecting <a
- href="http://www.unicode.org/reports/tr39/#Confusable_Detection">confusable characters</a> or
- warning when they are in use.
+ <p class="note no-backref">Note that various characters can be used in homograph [[IDNFAQ]]
+ spoofing attacks. Consider detecting confusable characters [[UTR39]] and warning when they are in
```
sideshowbarker: The reference here should now be to `[[UTS39]]` instead of `[[UTR39]]`. (I should’ve added it to Specref as  UTS39 to begin with, but anyway https://github.com/tobie/specref/pull/522 from @TimothyGu subsequently revamped all the Specref handling for Unicode references, including that fix.)

estark37: Fixed, thanks!

url.bs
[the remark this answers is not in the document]
annevk: Filed #435.

annevk:
I pushed a commit with my final nits and filed a follow-up issue on domain labels. From my perspective I'm happy to merge this, but since lots of people have helped shaped this (thanks!) I'll give it a bit more time before doing so.

estark37:
Friendly ping @annevk

estark37:
(To be clear, that ping was about whether that can be merged now that there's been a chance for people to look it over again)

annevk:
Ah, thanks! I had both forgotten a bit due to travel, but I was also hoping you'd mention you're okay with my edits, which I'll assume you are now. 😊

Really appreciate the work you put into this and everyone that helped refine it here!

Issue #448, opened 2019-09-11
Consider introducing a "same-site" concept that includes scheme.


mikewest:
As discussed in https://github.com/w3c/webappsec-fetch-metadata/issues/34, we're slowly tightening the various places that evaluate URLs'/Origins' same-siteness to look beyond the host and require scheme matches as well.

That is, `http://example.com/` and `https://sub.example.com/` have _hosts_ which are considered 
"same-site", but many places in the platform have begun considering those contexts "cross-site" as the non-secure scheme creates risk for the secure origin.

To quote from that other thread:

> @mikewest 
> In fact, I wonder if we can/should change this more generally for URL's "same site" definition, as it's something we run into in ~all the "same-site" checks we're introducing these days.
>
> In fact, SameSite cookies might be the only place where we're not doing this kind of check. We made that decision because we wanted people to be able to upgrade their sites to HTTPS piecemeal, starting with their authentication systems, then spreading throughout the site. Perhaps we've reached enough of an inflection point with that migration that we can tighten the restriction generally.

> @annevk 
> I'm supportive. We could have schemeless same site or some such for the variant that only takes hosts. https://fetch.spec.whatwg.org/#cross-origin-resource-policy-check also needs schemeless btw, even though it ends up checking the scheme a bit.

annevk:
I guess to be useful we should define this for both origins and URLs.

mikewest:
I agree. The latter seems reasonably placed in URL. Would you prefer the former to live in HTML?

annevk:
Oh that's a good question, I guess so? And have the latter defined in terms of the former? It'll all be rather circular though, hope that's acceptable to people (I guess in practice almost nobody will click through to verify things make sense).

mikewest:
Indeed.

(Note that I haven't asked what RFC6265bis needs to depend on for `SameSite` (or whether HTML/URL should depend on it?), because that will be an exciting conversation that @mnot should probably be involved in.)

mikewest:
I'll try to send you PRs that we can chat about at TPAC.

mikewest:
(@annevk: Would you be willing to allow me to upload branches to this repository? It would save me the trouble of forking it again (which, admittedly, isn't much trouble. :)).

annevk:
You should have write access now.

PR #449, opened 2019-09-13
Tighten 'same site' checks to include 'scheme'.

mikewest:
This patch introduces 'schemelessly same site' on both 'host' and 'URL',
moves 'same site' to URL, and tightens it by requiring a scheme match in
addition to a registrable domain match. This, hopefully, will create
stronger security boundaries in places where we perform "same site"
checks, while giving us the flexibility to grandparent in existing
behavior we decide to keep.

Partially addresses whatwg/url#448.

***
<a href="https://whatpr.org/url/449.html" title="Last updated on Sep 13, 2019, 10:14 AM UTC (16677c3)">Preview</a> | <a href="https://whatpr.org/url/449/3730361...16677c3.html" title="Last updated on Sep 13, 2019, 10:14 AM UTC (16677c3)">Diff</a>

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -346,50 +346,6 @@ obtain <var>host</var>'s <a for=host>registrable domain</a>, run these steps:
  </table>
 </div>
 
-<p>Two <a for=/>hosts</a>, <var>A</var> and <var>B</var> are said to be
-<dfn for=host export>same site</dfn> with each other if either of the following statements are true:
-
-<ul class=brief>
- <li><p><var>A</var> <a for=host>equals</a> <var>B</var> and <var>A</var>'s
- <a for=host>registrable domain</a> is non-null.
-
- <li><p><var>A</var>'s <a for=host>registrable domain</a> is <var>B</var>'s
- <a for=host>registrable domain</a> and is non-null.
-</ul>
-
-<div class=example id=example-same-site>
- <p>Assuming that <code>suffix.example</code> is a <a for=host>public suffix</a> and that
- <code>example.com</code> is not:
-
- <ul>
-  <li><p><code>example.com</code>, <code>sub.example.com</code>, <code>other.example.com</code>,
-  <code>sub.sub.example.com</code>, and <code>sub.other.example.com</code> are all <a>same site</a>
-  with each other (and themselves), as their <a for=host>registrable domains</a> are
-  <code>example.com</code>.
-
-  <li><p><code>registrable.suffix.example</code>, <code>sub.registrable.suffix.example</code>,
-  <code>other.registrable.suffix.example</code>, <code>sub.sub.registrable.suffix.example</code>,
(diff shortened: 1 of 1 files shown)


mikewest:
(Nevermind, I apparently already had a fork of the repo lying around. :) )

url.bs
```diff
@@ -2473,6 +2429,77 @@ background information. [[!HTML]]
...
+ <a for=host>registrable domain</a> is non-null.
+
+ <li><p><var>A</var>'s <a for=host>registrable domain</a> is <var>B</var>'s
+ <a for=host>registrable domain</a> and is non-null.
+</ul>
+
+<p>Two <a for=/>URL</a>s, <var>A</var> and <var>B</var> are said to be
+<dfn for=url>schemelessly same site</dfn> with each other if <var>A</var>'s
+<a for=url>host</a> is <a for=host>schemelessly same site</a> with <var>B</var>'s
+<a for=url>host</a>.
```
annevk: A URL's host can be null. So probably needs to be about as complicated (or you need to assert a certain scheme).

domenic:
I ran into the same confusion today writing an explainer where I wanted to use "same site" and "site isolation" to mean the same thing. I'm happy to try to drive this check over the finish line, but I can't understand @annevk's outstanding review comment. (In particular the second sentence seems to be missing some words.)

annevk:
Sorry, I meant about as complicated as the comparison done above for hosts. In that you have to null check host before comparing and returning false if a host is null.

Also, reading the original issue again the idea was that we'd also define same-site for origins in HTML and defer the URL comparison to that. Meaning that for a URL you first compute the origin and then see if it's same-site.

Also, HTML has this "agent cluster key" concept. Should we define that as "site" and also define "schemeless site" so we can talk about these things concretely without referring to the comparisons?

domenic:
Those all seem like reasonable things to do. The definitions for "site" and "schemeless site" seem like they should be able to be done in the URL spec.

If we changed same-site (and probably schemelessly-same-site) to operate on origins I'm unsure whether it'd still be useful to have definitions for URLs in this spec.

annevk:
The usage in https://fetch.spec.whatwg.org/#cross-origin-resource-policy-check could be done on top of origins entirely.

HTML's https://html.spec.whatwg.org/multipage/origin.html#is-a-registrable-domain-suffix-of-or-is-equal-to is also used by WebAuthn, but uses registrable domain and public suffix directly (and has to).

I think HTML's algorithm could be used for cookies too, reading https://tools.ietf.org/html/draft-ietf-httpbis-rfc6265bis-03#section-5.1.3.

HTML's agent stuff would be a little cleaner if you could obtain a site from an origin (i.e., "obtain an agent cluster key" renamed).

No other uses come to mind, but it seems like you had at least one other and maybe Mike knows more.

PR #453, opened 2019-10-17
Remove gopher from list of special schemes

### LINKED ISSUE #342: Consider removing the `gopher` special-case.
`gopher` is called out as a special scheme with a default port of 70 (https://url.spec.whatwg.org/#url-miscellaneous). Browsers generally implement this bit of the spec. I'd suggest that it's time we stopped doing so.

So far as I can tell, Chrome's support for `gopher` is limited to URL parsing. The protocol itself isn't supported, and `gopher:` URLs in the address bar are treated as searches. Gecko looks like it has similar behavior.

Perhaps we could consider dropping the special-casing in URL, and just treating `gopher:` like any other scheme that the browser doesn't understand. That is, instead of parsing `gopher://hostname.goes.here` like we'd parse `http://hostname.goes.here`, we'd parse it like `not-a-scheme://hostname.goes.here`. This wouldn't preclude user agents from supporting the protocol with external handlers, of course, but it would allow them to stop pretending that they understand the scheme when they clearly don't.

WDYT, @annevk, @mcmanus, @sleevi, @travisleithead, @achristensen07?

achristensen07:
Fixes #342

***
<a href="https://whatpr.org/url/453.html" title="Last updated on Oct 17, 2019, 7:09 PM UTC (1dba2e1)">Preview</a> | <a href="https://whatpr.org/url/453/5f8576e...1dba2e1.html" title="Last updated on Oct 17, 2019, 7:09 PM UTC (1dba2e1)">Diff</a>

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -1141,19 +1141,11 @@ corresponding <a for=url>port</a> and is listed in the second column on the same
      <th><a for=url>port</a>
  <tr><td>"<code>ftp</code>"<td>21
  <tr><td>"<code>file</code>"<td>
- <tr><td>"<code>gopher</code>"<td>70
  <tr><td>"<code>http</code>"<td>80
  <tr><td>"<code>https</code>"<td>443
  <tr><td>"<code>ws</code>"<td>80
  <tr><td>"<code>wss</code>"<td>443
 </table>
-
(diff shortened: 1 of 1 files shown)


annevk:
Thanks @achristensen07!

PR #457, opened 2019-11-11
Let HTML handle the "same site" definition

domenic:
Closes #449. Partially addresses #448.

---

Marking "do not merge yet" because the build will fail until the HTML pull request is merged.

***
<a href="https://whatpr.org/url/457.html" title="Last updated on Nov 20, 2019, 5:39 PM UTC (51f8e2f)">Preview</a> | <a href="https://whatpr.org/url/457/7ae1c69...51f8e2f.html" title="Last updated on Nov 20, 2019, 5:39 PM UTC (51f8e2f)">Diff</a>

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -346,49 +346,13 @@ obtain <var>host</var>'s <a for=host>registrable domain</a>, run these steps:
  </table>
 </div>
 
-<p>Two <a for=/>hosts</a>, <var>A</var> and <var>B</var> are said to be
-<dfn for=host export>same site</dfn> with each other if either of the following statements are true:
-
-<ul class=brief>
- <li><p><var>A</var> <a for=host>equals</a> <var>B</var> and <var>A</var>'s
- <a for=host>registrable domain</a> is non-null.
-
- <li><p><var>A</var>'s <a for=host>registrable domain</a> is <var>B</var>'s
- <a for=host>registrable domain</a> and is non-null.
-</ul>
-
-<div class=example id=example-same-site>
- <p>Assuming that <code>suffix.example</code> is a <a for=host>public suffix</a> and that
- <code>example.com</code> is not:
-
- <ul>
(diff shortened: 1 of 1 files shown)


url.bs
```diff
@@ -346,49 +346,13 @@ obtain <var>host</var>'s <a for=host>registrable domain</a>, run these steps:
...
-suffix list will diverge from client to client. Specifications which ignore this advice are
-encouraged to carefully consider whether URLs' schemes ought to be incorporated into any decision
-made based upon whether or not two <a for=/>hosts</a> are <a>same site</a>. HTML's <a>same
-origin-domain</a> concept is a reasonable example of this consideration in practice.
+decisions. The notion of "<a for=host>public suffix</a>" and "<a for=host>registrable domain</a>"
+cannot be relied-upon to provide a hard security boundary, as the public suffix list will diverge
+from client to client. Specifications which ignore this advice are encouraged to carefully consider
+whether URLs' schemes ought to be incorporated into any decisions made, i.e. whether to use the
+<a>same site</a> or <a>schemelessly same site</a> concepts. HTML's <a>same origin-domain</a> concept
+is a reasonable example of this consideration in practice.
```
annevk: Should we give this an ID and reference it from HTML?

domenic: The note, you mean? I'm not sure what we would reference it for... personally I find the note pretty confusing, and was mostly trying to just preserve the meaning.

annevk: The point is that we don't really want people to use (schemelessly) same site if they can avoid it in favor of origins. And I'd like HTML to also suggest that to some extent.

domenic: Added an ID; will update HTML pull request now.

I was also thinking of removing the pointer to HTML's same-origin domain concept since I'm not sure I'd call it "reasonable". WDYT?

annevk: Yeah, let's remove that.

domenic:
This should be mergeable whenever the bikeshed link database updates. I'll set a reminder to do so tomorrow, if you give it the green checkmark.

PR #459, opened 2019-12-12
Forbid < and > in hosts

### LINKED ISSUE #458: Forbid < and > in hosts
https://url.spec.whatwg.org/#forbidden-host-code-point currently lists a number of code points as forbidden in hosts. However, some implementations have more:

- [Gecko](https://searchfox.org/mozilla-central/rev/62a130ba0ac80f75175e4b65536290b52391f116/netwerk/base/nsStandardURL.cpp#67-80) additionally has `*<>^|"` (`^` was [recently added](https://bugzilla.mozilla.org/show_bug.cgi?id=1548306))
- [WebKit](https://github.com/WebKit/webkit/blob/fc9fab113bf5016fc7cecc812760487ef615fd43/Source/WebCore/platform/URLParser.cpp#L154-L411) additionally has `<>` (added in [2018](https://github.com/WebKit/webkit/commit/eb0f458f05e0886de8881edf5fa64978c0e0439e))

Considering `<>` seem to be forbidden in two browsers, we should consider making them forbidden too. The list is testable through `new URL('http://</').href`.

See also: nodejs/node#30223

ZYSzys:
> Considering <> seem to be forbidden in two browsers, we should consider making them forbidden too. The list is testable through new URL('http://</').href.

This PR add `<` and `>` to `Host miscellaneous`, Close #458

***
<a href="https://whatpr.org/url/459.html" title="Last updated on May 7, 2020, 5:04 PM UTC (605aa0f)">Preview</a> | <a href="https://whatpr.org/url/459/3703f92...605aa0f.html" title="Last updated on May 7, 2020, 5:04 PM UTC (605aa0f)">Diff</a>

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -274,8 +274,8 @@ for further processing.
 <h3 id=host-miscellaneous>Host miscellaneous</h3>
 
 <p>A <dfn export>forbidden host code point</dfn> is U+0000 NULL, U+0009 TAB, U+000A LF, U+000D CR,
-U+0020 SPACE, U+0023 (#), U+0025 (%), U+002F (/), U+003A (:), U+003F (?), U+0040 (@), U+005B ([),
-U+005C (\), or U+005D (]).
+U+0020 SPACE, U+0023 (#), U+0025 (%), U+002F (/), U+003A (:), U+003C (<), or U+003E (>), U+003F (?),
+U+0040 (@), U+005B ([), U+005C (\), or U+005D (]).
 
 <p>A <a for=/>host</a>'s <dfn for=host export>public suffix</dfn> is the portion of a
 <a for=/>host</a> which is included on the <cite>Public Suffix List</cite>. To obtain
--- a/url.bs
+++ b/url.bs
@@ -3313,7 +3313,8 @@ Trevor Rowbotham,
 Valentin Gosu,
 Vyacheslav Matva,
 Wei Wang,
-山岸和利 (Yamagishi Kazutoshi), and
+山岸和利 (Yamagishi Kazutoshi),
+Yongsheng Zhang, and
 成瀬ゆい (Yui Naruse)
 for being awesome!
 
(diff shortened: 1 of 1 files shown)


url.bs
```diff
@@ -274,8 +274,8 @@ for further processing.
 <h3 id=host-miscellaneous>Host miscellaneous</h3>
 
 <p>A <dfn export>forbidden host code point</dfn> is U+0000 NULL, U+0009 TAB, U+000A LF, U+000D CR,
-U+0020 SPACE, U+0023 (#), U+0025 (%), U+002F (/), U+003A (:), U+003F (?), U+0040 (@), U+005B ([),
-U+005C (\), or U+005D (]).
+U+0020 SPACE, U+0023 (#), U+0025 (%), U+002F (/), U+003A (:), U+003C (<), U+003E (>), U+003F (?),
```
annevk, line 277:
```suggestion
U+0020 SPACE, U+0023 (#), U+0025 (%), U+002F (/), U+003A (:), U+003C (&lt;), U+003E (>), U+003F (?),
```

ZYSzys: If changing `<` to `$lt;` here, should I also change `>` to `&gt;`, to be consistent ?

annevk: No that's not needed. HTML only requires it for `<` to disambiguate from a start tag. I guess maybe bikeshed does something so that it's not always needed, but I prefer keeping the source of the standard mostly HTML-compatible.

domenic:
Sadly I'm not able to spare the creative energy for writing tests, but I can do the usual jsdom/whatwg-url testing and reviewing. In advance I will report that there is only one failure in the existing test suite, which is parsing `sc://!"$&'()*+,-.;<=>^_`{|}~/` against `about:blank`.

rmisev:
Added tests: https://github.com/web-platform-tests/wpt/pull/23572

annevk:
Thanks everyone!

I filed https://bugs.webkit.org/show_bug.cgi?id=211901. Firefox and Chrome already have a more general bug on implementing non-special URLs properly.

PR #482, opened 2020-04-26
Clarify double-dot path segment parsing

### LINKED ISSUE #295: "append the empty string to url’s path." is confusing
https://url.spec.whatwg.org/commit-snapshots/488c459d9e4245a3f6bf087e7dcd2c7e91487ac5/#url-parsing

> append the empty string to url’s path.

It would be nice if the text explained why we are doing this apparently pointless step.

annevk:
Fixes #295.

@GPHemsley @Hixie, what do you think?

***
<a href="https://whatpr.org/url/482.html" title="Last updated on Apr 28, 2020, 6:17 AM UTC (3f408d6)">Preview</a> | <a href="https://whatpr.org/url/482/980ba89...3f408d6.html" title="Last updated on Apr 28, 2020, 6:17 AM UTC (3f408d6)">Diff</a>

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -2110,7 +2110,8 @@ string <var>input</var>, optionally with a <a>base URL</a> <var>base</var>, opti
           U+005C (\), <a for=list>append</a> the empty string to <var>url</var>'s
           <a for=url>path</a>.
 
-          <p class="note no-backref">Usually <a>c</a> is U+002F (/) at this point.
+          <p class="note no-backref">This means that for input <code>/usr/..</code> the result is
+          <code>/</code> and not a lack of a path.
         </ol>
 
        <li><p>Otherwise, if <var>buffer</var> is a <a>single-dot path segment</a> and if neither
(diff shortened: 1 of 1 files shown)


url.bs
```diff
@@ -2099,10 +2099,19 @@ string <var>input</var>, optionally with a <a>base URL</a> <var>base</var>, opti
...
+
+        <ol>
+         <li><p><a>Shorten</a> <var>url</var>'s <a for=url>path</a>.
+
+         <li>
+          <p>If neither <a>c</a> is U+002F (/), nor <var>url</var> <a>is special</a> and <a>c</a> is
+          U+005C (\), <a for=list>append</a> the empty string to <var>url</var>'s
+          <a for=url>path</a>.
+
+          <p class="note no-backref">Usually <a>c</a> is U+002F (/) at this point.
```
domenic, line 2113: This doesn't really help much IMO. Addressing #482 would involve explaining the meaning of empty path segments, either here or somewhere else that this step could link to.

annevk: How are they different from other path segments?

domenic: Empty is often a special case, at least in human's minds. I agree for computers they are not different.

annevk: Okay, would something like "This means that for input `/foo/..` the result is `/foo/` and not `/foo`." work?

domenic: Yeah that makes a lot of sense to me.

PR #484, opened 2020-04-26
Clarify output of PSL is in A-Labels

### LINKED ISSUE #396: Clarify A-label vs U-label for public suffix / registrable domain
At the moment the Public Suffix List allows implementations to do either. The URL Standard clearly centers around A-label (ASCII) for everything normalized and expects the same from public suffix and registrable domain (due to the way it performs comparisons with them).

Filing this as a follow-up to #391 since we don't expose public suffix and registrable domain directly (yet).

I'd suggest we change

> Return the public suffix obtained by executing the algorithm defined by the Public Suffix List on host.

to

> Return the public suffix obtained by executing the algorithm defined by the Public Suffix List on host, using A-labels.

and do likewise for registrable domain.

Does that work @sleevi & @mikewest?

annevk:
Closes #396.

***
<a href="https://whatpr.org/url/484.html" title="Last updated on Apr 26, 2020, 3:58 PM UTC (695c968)">Preview</a> | <a href="https://whatpr.org/url/484/980ba89...695c968.html" title="Last updated on Apr 26, 2020, 3:58 PM UTC (695c968)">Diff</a>

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -288,7 +288,7 @@ U+005C (\), or U+005D (]).
  <a href="https://publicsuffix.org/list/">Public Suffix List algorithm</a> with <var>host</var> as
  domain. [[!PSL]]
 
- <li><p>Assert: <var>publicSuffix</var> contains only <a>ASCII code points</a>.
+ <li><p>Assert: <var>publicSuffix</var> is an <a>ASCII string</a>.
 
  <li><p>Return <var>publicSuffix</var>.
 </ol>
@@ -305,7 +305,7 @@ obtain <var>host</var>'s <a for=host>registrable domain</a>, run these steps:
  <a href="https://publicsuffix.org/list/">Public Suffix List algorithm</a> with <var>host</var> as
  domain. [[!PSL]]
 
- <li><p>Assert: <var>registrableDomain</var> contains only <a>ASCII code points</a>.
+ <li><p>Assert: <var>registrableDomain</var> is an <a>ASCII string</a>.
 
  <li><p>Return <var>registrableDomain</var>.
 </ol>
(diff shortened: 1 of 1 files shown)


url.bs
```diff
@@ -284,9 +284,13 @@ U+005C (\), or U+005D (]).
...
  <li><p>If <var>host</var> is not a <a>domain</a>, then return null.
 
- <li><p>Return the <a for=host>public suffix</a> obtained by executing the
- <a href="https://publicsuffix.org/list/">algorithm</a> defined by the Public Suffix List on
- <var>host</var>. [[!PSL]].
+ <li><p>Let <var>publicSuffix</var> be the public suffix determined by running the
+ <a href="https://publicsuffix.org/list/">Public Suffix List algorithm</a> with <var>host</var> as
+ domain. [[!PSL]]
+
+ <li><p>Assert: <var>publicSuffix</var> is encoded using A-Labels. [[!UTS46]]
```
TimothyGu, line 291: UTS 46 doesn't really define what an A-label is, as it only seeks to define a replacement for the processing algorithm. In fact, it defers the definition to [RFC 5890](https://tools.ietf.org/html/rfc5890) and doesn't use the word at all itself.

Additionally, the term "A-label" exclusively refers to validly encoded labels with `xn--` prefix. To refer to all ASCII-encoded labels, it may be better to use either of the following:

> Assert: _publicSuffix_ contains only ASCII characters.

or

> Assert: Labels in _publicSuffix_ must be either A-labels or NR-LDH labels. [RFC5890]

annevk: I wanted to use UTS46 as an intermediary, but as A-Label doesn't work and NR-LDH isn't mentioned there we'll just assert ASCII code points.

PR #485, opened 2020-04-26
Editorial: assert that init is a string

### LINKED ISSUE #439: "Otherwise, init is a string, then set ..." in list of steps for how to create a URLSearchParams object
In the [section 6.2](http://url.spec.whatwg.org/#interface-urlsearchparams), there is a list of steps an implementation must take as part of creating a `URLSearchParams` object. After defining some steps that apply depending on what kind of object `init` is, the 4th step is defined:

> 4. Otherwise, *init* is a string, then set *query*’s list to the result of parsing *init*.

How is one to interpret this sentence? Its first part seems... *assertive*, as opposed to being speculative, with regard to the type of object `init` is? Should there be an "if" between "Otherwise, " and "*init*"?

Or does the specification intentionally assert that if `init` isn't any kind of object as defined by earlier steps in the list, then it *is to be considered* a string (possibly co-ercing the value, which would support how Firefox and Chrome behave here) and the implementation is then to parse it?

Because there isn't otherwise any mention of coercion/cast of *init* to a string, or is this left up to implementations, by intent?

annevk:
Fixes #439.

***
<a href="https://whatpr.org/url/485.html" title="Last updated on Apr 26, 2020, 3:00 PM UTC (3ec3322)">Preview</a> | <a href="https://whatpr.org/url/485/980ba89...3ec3322.html" title="Last updated on Apr 26, 2020, 3:00 PM UTC (3ec3322)">Diff</a>

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -3077,9 +3077,15 @@ initially null.
  <var>name</var> and value is <var>value</var>, to <var>query</var>'s
  <a for=URLSearchParams>list</a>.
 
- <li><p>Otherwise, <var>init</var> is a string, then set <var>query</var>'s
- <a for=URLSearchParams>list</a> to the result of
- <a lt='urlencoded string parser'>parsing</a> <var>init</var>.
+ <li>
+  <p>Otherwise:
+
+  <ol>
+   <li><p>Assert: <var>init</var> is a string.
+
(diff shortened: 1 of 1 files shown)


annevk:
Thanks @TimothyGu!

PR #486, opened 2020-04-26
Remove U+0000 case in the fragment state

### LINKED ISSUE #440: Percent encode NULLs in fragments
Currently URL parser removes NULL characters from fragments. But is this really needed?
My tests show that NULLs are percent encoded in most browsers (I tested URLs "`https://example.com/#abc\u0000xyz`" and "`non-spec://example.com/#abc\u0000xyz`"):

Browser  | Special URL's hash | Non-special URL's hash
---------|--------------------|----------------------
Chrome   | `#abcxyz`          | `#abc%00xyz`
Edge, IE | `#abc`             | `#abc`
Firefox  | `#abc%00xyz`       | `#abc%00xyz`
Safari   | `#abc%00xyz`       | `#abc%00xyz`

So NULLs are removed in Chrome in special URLs only. In the Edge and IE a NULL character
denotes the end of string.

One reason, why Chrome removes NULLs, I found in the source code:
https://chromium.googlesource.com/chromium/src/+/refs/tags/76.0.3803.1/url/url_canon_etc.cc#304
```c++
  for (int i = ref.begin; i < end; i++) {
    if (spec[i] == 0) {
      // IE just strips NULLs, so we do too.
      continue;
    }
```

But as we see NULLs in Chrome and IE (and Edge) are handled differently.

I suggest to follow Firefox, Safari and percent encode NULLs in fragments.

annevk:
Tests: ...

Fixes #440.

(Note that I think we should make all states an explicit ordered list for clarity. Because of that I added it here, despite it only containing one step.)

   * Chrome
   * Firefox
   * Safari
   * https://github.com/web-platform-tests/wpt/pull/23256
   * Chrome: https://bugs.chromium.org/p/chromium/issues/detail?id=1075492
   * Firefox: (probably not needed, but need tests to confirm)
   * Safari: (probably not needed, but need tests to confirm)

***
<a href="https://whatpr.org/url/486.html" title="Last updated on Apr 27, 2020, 9:40 AM UTC (5d63a5c)">Preview</a> | <a href="https://whatpr.org/url/486/980ba89...5d63a5c.html" title="Last updated on Apr 27, 2020, 9:40 AM UTC (5d63a5c)">Diff</a>

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -2256,16 +2256,10 @@ string <var>input</var>, optionally with a <a>base URL</a> <var>base</var>, opti
 
    <dt><dfn>fragment state</dfn>
    <dd>
-    <p>Switching on <a>c</a>:
-    <dl class=switch>
-     <dt>The <a>EOF code point</a>
-     <dd><p>Do nothing.
+    <ol>
+     <li>
+      <p>If <a>c</a> is not the <a>EOF code point</a>, then:
 
-     <dt>U+0000 NULL
-     <dd><p><a>Validation error</a>.
-
-     <dt>Otherwise
-     <dd>
       <ol>
        <li><p>If <a>c</a> is not a <a>URL code point</a> and not U+0025 (%),
        <a>validation error</a>.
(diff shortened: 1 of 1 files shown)


rmisev:
LGTM. Just added tests: https://github.com/web-platform-tests/wpt/pull/23256

PR #488, opened 2020-05-04
Editorial: improve the URL parser

### LINKED ISSUE #373: "Let pointer be a pointer to first code point in..."
https://url.spec.whatwg.org/commit-snapshots/5fc4c4c17f3077beda5e3647df73f4b1b0ba7084/#url-parsing

> Let pointer be a pointer to first code point in input\.
> 
> 

What happens if input is the empty string? See some confusion in https://github.com/whatwg/fetch/issues/669

### LINKED ISSUE #480: Path setter jumps off the leading edge of input
When the path setter is called, the basic parser is overridden to the *path start state*. But per either 1.3 or 4.2 for that state, the pointer can be decremented:

> If c is neither U+002F (/) nor U+005C (\), then decrease pointer by one.

> If c is not U+002F (/), then decrease pointer by one.

This is ill-defined when the pointer is to the start of the input.

annevk:
Closes #373 and closes #480.

@rmisev @domenic @alercah @mattto would appreciate your feedback on this clarification.

***
<a href="https://whatpr.org/url/488.html" title="Last updated on May 4, 2020, 1:35 PM UTC (c75037d)">Preview</a> | <a href="https://whatpr.org/url/488/d8fb1a4...c75037d.html" title="Last updated on May 4, 2020, 1:35 PM UTC (c75037d)">Diff</a>

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -88,19 +88,29 @@ valid input. User agents, especially conformance checkers, are encouraged to rep
 
 <h3 id=parsers>Parsers</h3>
 
-<p>The <dfn>EOF code point</dfn> is a conceptual code point that signifies the end of a
-string or code point stream.
+<p>The <dfn>EOF code point</dfn> is a conceptual code point that signifies the end of a string or
+code point stream.
 
-<p>Within a parser algorithm that uses a <var>pointer</var> variable, <dfn>c</dfn>
-references the code point the <var>pointer</var> variable points to.
+<p>A <dfn>pointer</dfn> for a <a for=/>string</a> <var>input</var> is an integer that points to a
+<a for=/>code point</a> within <var>input</var>. Initially it points to the start of
+<var>input</var>. If it is &minus;1 it points nowhere. If it is greater than or equal to
+<var>input</var>'s <a for=string>code point length</a>, it points to the <a>EOF code point</a>.
 
-<p>Within a string-based parser algorithm that uses a <var>pointer</var> variable,
-<dfn>remaining</dfn> references the substring after <var>pointer</var> in the string
-being processed.
+<p>When a <a>pointer</a> is defined, <dfn>c</dfn> references the <a for=/>code point</a>
+<a>pointer</a> points to as long as it does not point nowhere. When <a>pointer</a> points to nowhere
+<a>c</a> cannot be used.
+
+<p>When a <a>pointer</a> is defined, <dfn>remaining</dfn> references the substring after <a>c</a> as
+long as <a>c</a> is not the <a>EOF code point</a>. When <a>c</a> is the <a>EOF code point</a>
+<a>remaining</a> cannot be used.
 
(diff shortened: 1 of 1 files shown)


url.bs
```diff
@@ -88,20 +88,30 @@ valid input. User agents, especially conformance checkers, are encouraged to rep
...
-references the code point the <var>pointer</var> variable points to.
+<p>A <dfn>pointer</dfn> for a <a for=/>string</a> <var>input</var> is an integer that points to a
+<a for=/>code point</a> within <var>input</var>. Initially it points to the start of
+<var>input</var>. If it is &minus;1 it points nowhere. If it is greater than or equal to
+<var>input</var>'s <a for=string>code point length</a>, it points to the <a>EOF code point</a>.
 
-<p>Within a string-based parser algorithm that uses a <var>pointer</var> variable,
-<dfn>remaining</dfn> references the substring after <var>pointer</var> in the string
-being processed.
+<p>When a <a>pointer</a> is defined, <dfn>c</dfn> references the <a for=/>code point</a>
```
domenic, line 99: What does "defined" mean? My first guess was "not -1", but reading the rest of the paragraph that doesn't seem right...

Maybe this is trying to say something like "each pointer has a c, which is a reference to the..."?

url.bs
```diff
@@ -88,20 +88,30 @@ valid input. User agents, especially conformance checkers, are encouraged to rep
...
+<a>pointer</a> points to as long as it does not point nowhere. When <a>pointer</a> points to nowhere
+<a>c</a> cannot be used.
+
+<p>When a <a>pointer</a> is defined, <dfn>remaining</dfn> references the substring after <a>c</a> as
+long as <a>c</a> is not the <a>EOF code point</a>. When <a>c</a> is the <a>EOF code point</a>
+<a>remaining</a> cannot be used.
 
 <p class=example id=example-12672b6a>If "<code>mailto:username@example</code>" is a <a>string</a>
-being processed and <var>pointer</var> points to @, <a>c</a> is U+0040 (@) and <a>remaining</a> is
+being processed and <a>pointer</a> points to @, <a>c</a> is U+0040 (@) and <a>remaining</a> is
```
domenic, line 108: "pointer points to" seems a bit strange; should it be "a pointer" or "the pointer"?

url.bs
annevk: Would "When a pointer is used" work?

url.bs
annevk: I can make it "a pointer"

url.bs
domenic: Yeah, that makes sense I think.

annevk:
To summarize the other changes:

1. Use 1 instead of one when decreasing or increasing.
2. Remove a comma preceding and if there are only two things.
3. Give all states a nested list.
4. Clone lists (with clone being defined) rather than copy.

alercah:
Looks like it addresses my comments in #480. Thanks!

Issue #337, opened 2017-07-17
Add a neato informative table of various URL pieces


domenic:
Basically copy the bottom half of this: https://nodejs.org/api/url.html#url_url_strings_and_url_objects

(We could presumably SVG-ize it so it's a little prettier.)   

Via the thread at https://twitter.com/wa7son/status/886982643463708673

TimothyGu:
Quick WIP with Inkscape:

![drawing svg](https://user-images.githubusercontent.com/1538624/30042227-a0f3af64-9222-11e7-96a4-39c0cf11d279.png)

Would something like this work?

annevk:
I think so, although maybe a table is better as that would be more accessible I suspect. Note also that `?` is not part of the `pathname` getter. The other thing that might be interesting is to illustrate a couple different URLs. In particular different schemes. You also omitted the origin field although that's rather hard given that it needs to skip user/pass somehow.

TimothyGu:
> Note also that `?` is not part of the `pathname` getter.

Oops, an off-by-one.

> In particular different schemes

That does indeed sound like a nice idea. 

> You also omitted the origin field

I did so intentionally, as it's not a concept intrinsically related to URL parsing, but rather more about Web apps/security. And because it's hard.

----

I did a table version with a few variants. The first is a straight translation of the SVG graph. The second is closer to the version in the Node.js doc. The third is the same as the second but has `origin`, mainly there to show how ugly it is. The forth is a URN for fun. I do like the fact that you can link to the spec for that exact attribute, but with some coloring I still think the SVG one is a bit prettier.

![screenshot from 2017-09-05 16-07-49](https://user-images.githubusercontent.com/1538624/30051381-646ac596-9254-11e7-9e24-4b9e53fd4fd7.png)

annevk:
Alright, I'm game. We should be able to get the links to work with SVG too. I'm not really sure if we can make all of it equally accessible though.

domenic:
Personally I like the first table one, possibly with additional text-align center.

I also think it might be interesting to have a counterpart table that is about the URL record terms, instead of the API? (E.g. scheme instead of protocol, query instead of search, fragment instead of hash.) Maybe that wouldn't be that helpful though.

annevk:
It's probably useful as there are some interesting differences between the two. Bit unclear where the table should be located at that point, but maybe we could put it in an Appendix?

EnnexMB:
[...]

PR #489, opened 2020-05-04
Clarify how URL strings map to URL components via table

annevk:
Helps with #337.

Next is taking @TimothyGu's work and give the API a similar table...

***
<a href="https://whatpr.org/url/489.html" title="Last updated on May 5, 2020, 8:08 AM UTC (4e48520)">Preview</a> | <a href="https://whatpr.org/url/489/d8fb1a4...4e48520.html" title="Last updated on May 5, 2020, 8:08 AM UTC (4e48520)">Diff</a>

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -1114,6 +1114,50 @@ refers to as well as its origin. It is important that these are cached as the <a
 be removed from the <a>blob URL store</a> between parsing and fetching, while fetching will still
 need to succeed.
 
+<div class=example id=example-url-components>
+ <p>The following table lists how <a>valid URL strings</a>, when <a lt="URL parser">parsed</a>, map
+ to a <a for=/>URL</a>'s components. Username and password are left as exercise for the reader.
+
+ <table>
+  <tr>
+   <th>Input
+   <th>Scheme
+   <th>Host
+   <th>Port
+   <th>Path
+   <th>Query
+   <th>Fragment
+   <th>Cannot-be-base-URL flag
+  <tr>
+   <td><code>https://example.com/</code>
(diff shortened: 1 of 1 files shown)


url.bs
```diff
@@ -1114,6 +1114,50 @@ refers to as well as its origin. It is important that these are cached as the <a
 be removed from the <a>blob URL store</a> between parsing and fetching, while fetching will still
 need to succeed.
 
+<div class=example id=example-url-components>
+ <p>The following table lists how <a>valid URL strings</a>, when <a lt="URL parser">parsed</a>, map
+ to a <a for=/>URL</a>'s components. Username and password are left as exercise for the reader.
```
domenic, line 1119: "left as an exercise for the reader" is a bit misleading because it implies that the reader might have to do some work. Instead I'd state this as something like

> Username and password are omitted from the table; in all the examples given they are the empty string. Similarly for blob URL entry, which is null in all the examples given.

url.bs
```diff
@@ -1114,6 +1114,50 @@ refers to as well as its origin. It is important that these are cached as the <a
...
+  <tr>
+   <td><code>urn:isbn:9780307476463</code>
+   <td>"<code>urn</code>"
+   <td>null
+   <td>null
+   <td>« "<code>isbn:9780307476463</code>" »
+   <td>null
+   <td>null
+   <td>set
+ </table>
```
domenic, line 1158: A file: URL might be helpful for people to see, and could also allow illustrating multiple path segments.

annevk: Great idea! Done.

PR #490, opened 2020-05-04
Clarify how URL strings map to URL getters via table

annevk:
Helps with #337.

***
<a href="https://whatpr.org/url/490.html" title="Last updated on Jan 15, 2021, 7:42 AM UTC (1cff012)">Preview</a> | <a href="https://whatpr.org/url/490/d8fb1a4...1cff012.html" title="Last updated on Jan 15, 2021, 7:42 AM UTC (1cff012)">Diff</a>

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -2723,18 +2723,46 @@ interface URL {
 };
 </pre>
 
-<!-- XXX Ideas:
-  boolean isEqual(URL, optional URLEqualOptions options)
-           attribute URLPath segments;
+<div class=example id=example-url-components>
+ <p>The following table lists how <a>valid URL strings</a>, when <a lt="URL parser">parsed</a>, map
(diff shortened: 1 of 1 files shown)


domenic:
I think omitting origin and host removes most of the value of such a table, unfortunately. Those are the confusing cases where an example is most helpful.

annevk:
I suppose I could see how wide the table gets. Where would you put them? Or are you thinking about the multiple rows approach? (I didn't really like how that ended up looking.)

domenic:
I was thinking of the multiple rows approach. It's pretty common, e.g. something like it (in ASCII form) appears even in the RFCs.

annevk:
https://tools.ietf.org/html/rfc3986#section-3 seems to be a table with the heading row in the middle, but doesn't have multiple heading rows, or is there another one I missed?

domenic:
You're right, that's the one that I was thinking of.

Issue #282, opened 2017-03-28
Align with Infra


annevk:
- [ ] Reword any tuple language
- [ ] Rephrase URL records and hosts to be structs
- [ ] Investigate removing labels from the IPv6 parser
- ...

domenic:
Eh, they seem like structs already; it seems like you'd just call them a struct and be pretty much done?

annevk:
Yeah okay, I guess I'm wondering if that's something we want to do. Should all our concepts bottom out in these types? Should request and response be structs too?

PR #493, opened 2020-05-05
IPv4 parser cleanup

annevk:
Helps with #282.

***
<a href="https://whatpr.org/url/493.html" title="Last updated on May 5, 2020, 3:29 PM UTC (c806d34)">Preview</a> | <a href="https://whatpr.org/url/493/909218b...c806d34.html" title="Last updated on May 5, 2020, 3:29 PM UTC (c806d34)">Diff</a>

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -510,10 +510,11 @@ string <var>input</var> with an optional boolean <var>isNotSpecial</var>, and th
  <li><p>Return <var>asciiDomain</var>.
 </ol>
 
-<p>The <dfn>IPv4 number parser</dfn> takes a string <var>input</var> and a
-<var>validationErrorFlag</var> pointer, and then runs these steps:
+<p>The <dfn>IPv4 number parser</dfn> takes a string <var>input</var> and then runs these steps:
 
 <ol>
+ <li><p>Let <var>validationError</var> be false.
+
  <li><p>Let <var>R</var> be 10.
 
  <li>
@@ -521,7 +522,7 @@ string <var>input</var> with an optional boolean <var>isNotSpecial</var>, and th
   "<code>0x</code>" or "<code>0X</code>", then:
 
   <ol>
-   <li><p>Set <var>validationErrorFlag</var>.
+   <li><p>Set <var>validationError</var> to true.
 
    <li><p>Remove the first two code points from <var>input</var>.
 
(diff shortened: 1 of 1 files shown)


domenic:
It's not clear to me why the validationError variable exists in the IPv4 parser algorithm. It seems like the net effect is delaying the validation error until later, and "hiding" multiple validation errors behind a single validation error. Would it be bad to just report the validation errors when they occur, in steps 3.1 and 6.4?

Another issue: step 6.2/6.3 assumes that the tuple will be of the form (failure, validationError) in the case of failure. But the IPv4 number parser algorithm step 6 just returns raw "failure".

annevk:
Since https://url.spec.whatwg.org/#concept-host-parser almost always invokes the IPv4 parser that tries to avoid reporting errors when it tries to parse a "non-IPv4 address". I should fix that bug though.

domenic:
Hmm, I think I see. So if we made it to step 7, then none of the calls to IPv4 number parser returned "failure", so we're confident that we're parsing as an IPv4 address? If so I think calling that out explicitly might be helpful; I'll try to do one of those fancy multi-line suggestion things.

url.bs
```diff
@@ -510,114 +510,77 @@ string <var>input</var> with an optional boolean <var>isNotSpecial</var>, and th
...
 
-   <li><p>If <var>n</var> is failure, return <var>input</var>.
+   <li><p>If <var>result</var>[1] is true, then set <var>validationError</var> to true.
 
-   <li><p>Append <var>n</var> to <var>numbers</var>.
+   <li><p><a for=list>Append</a> <var>result</var>[0] to <var>numbers</var>.
   </ol>
 
- <li><p>If <var>validationErrorFlag</var> is set, <a>validation error</a>.
+ <li><p>If <var>validationError</var> is true, <a>validation error</a>.
```
domenic, line 565:
```suggestion
 <li>
  <p class="note">If we've reached this point, then each part was parsed into a number, so we're
  committed to treating <var>input</var> as an IPv4 address and can resume reporting validation
  errors.</p>

  <p>If <var>validationError</var> is true, <a>validation error</a>.
 </li>
```

annevk:
I don't really like the aesthetic of a step starting with a note so I'm gonna tweak it a bit more...

PR #494, opened 2020-05-05
Fix validity for opaque hosts

### LINKED ISSUE #437: Definition of "valid opaque-host string" contradicts parser
Per the definitions in https://url.spec.whatwg.org/#urls, I believe that the following is a [*valid URL string*](https://url.spec.whatwg.org/#valid-url-string)

```
qwe://foobar/baz
```

with a [*valid opaque-host string*](https://url.spec.whatwg.org/#valid-opaque-host-string) of `foobar`, whereas

```
qwe://foo^bar/baz
```

is *not* a valid URL string because `^` is not a ["URL unit"](https://url.spec.whatwg.org/#url-units) and therefore `foo^bar` is not a valid opaque-host string (which is defined to be made up of URL units).

However, the parser disagrees. If I step through the [parsing algorithm](https://url.spec.whatwg.org/#concept-basic-url-parser) with `qwe://foo^bar/baz` as input, then `foo^baz` ultimately gets passed to the [opaque-host parser](https://url.spec.whatwg.org/#concept-opaque-host-parser) without any validation error being triggered, and then the opaque-host parser in turn returns `foo^bar` without a validation error.

I believe this is a spec bug; as I understand https://url.spec.whatwg.org/#urls, a string is meant to be a "valid URL string" if and only if it does not trigger an error from the URL parser. But `qwe://foo^bar/baz`, which is not a "valid URL string" per the definition at https://url.spec.whatwg.org/#valid-url-string, does not trigger an error.

annevk:
Closes #437.

(It does strike me that there are some patterns here that might warrant abstraction.)

***
<a href="https://whatpr.org/url/494.html" title="Last updated on May 5, 2020, 3:41 PM UTC (fd7a344)">Preview</a> | <a href="https://whatpr.org/url/494/1bb45ce...fd7a344.html" title="Last updated on May 5, 2020, 3:41 PM UTC (fd7a344)">Diff</a>

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -457,9 +457,12 @@ up to three <a>ASCII digits</a> per sequence, each representing a decimal number
 
      XXX should we define the format inline instead just like STD 66? -->
 
-<p>A <dfn export>valid opaque-host string</dfn> must be one or more <a>URL units</a>, excluding
-<a>forbidden host code points</a>, or: U+005B ([), followed by a <a>valid IPv6-address string</a>,
-followed by U+005D (]).
+<p>A <dfn export>valid opaque-host string</dfn> must be one of the following:
+
+<ul class=brief>
+ <li><p>one or more <a>URL units</a> excluding <a>forbidden host code points</a>
+ <li><p>U+005B ([), followed by a <a>valid IPv6-address string</a>, followed by U+005D (]).
+</ul>
 
 <p class="note no-backref">This is not part of the definition of <a>valid host string</a> as it
 requires context to be distinguished.
@@ -788,7 +791,7 @@ then runs these steps:
 <var>input</var>, and then runs these steps:
 
 <ol>
- <li><p>If <var>input</var> contains a <a>forbidden host code point</a>, excluding U+0025 (%),
+ <li><p>If <var>input</var> contains a <a>forbidden host code point</a> excluding U+0025 (%),
  <a>validation error</a>, return failure.
 
  <li><p>If <var>input</var> contains a <a>code point</a> that is not a <a>URL code point</a> and not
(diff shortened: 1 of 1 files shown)


url.bs
```diff
@@ -457,8 +457,9 @@ up to three <a>ASCII digits</a> per sequence, each representing a decimal number
 
      XXX should we define the format inline instead just like STD 66? -->
 
-<p>A <dfn export>valid opaque-host string</dfn> must be one or more <a>URL units</a> or: U+005B ([),
-followed by a <a>valid IPv6-address string</a>, followed by U+005D (]).
+<p>A <dfn export>valid opaque-host string</dfn> must be one or more <a>URL units</a>, excluding
```
domenic, line 460: I think this would be clearer as a bulleted list (one bullet for each branch of the or)

url.bs
```diff
@@ -787,9 +788,15 @@ then runs these steps:
 <var>input</var>, and then runs these steps:
 
 <ol>
- <li><p>If <var>input</var> contains a <a>forbidden host code point</a> excluding U+0025 (%),
+ <li><p>If <var>input</var> contains a <a>forbidden host code point</a>, excluding U+0025 (%),
```
domenic, line 791: IMO this comma addition slightly decreases readability.

annevk: So one usage of excluding could also omit it, but another is after inclusive and therefore cannot omit it. I guess I'm okay with omitting when possible.

annevk:
It's been over a year since @ExplodingCabbage reported the issue, but if they're up for taking a look that'd be great. Otherwise I'll probably merge this as this is a lot less wrong as far as I can tell.

Issue #491, opened 2020-05-04
Introduce a variant of url.searchParams that operates according to URL rules instead of <form> rules


domenic:
### Problem

See background in #18 and #478.

`URLSearchParams` was designed, not to hold URL query data, but instead to hold `application/x-www-form-urlencoded` data, i.e. the data that is sent to a server when submitting a HTML `<form>`.

Unfortunately, it was misnamed `URLSearchParams` instead of `ApplicationXWWWFormURLEncodedParams`. And, even more unfortunately, a property named `searchParams` was added to the `URL` class, which is an instance of the `URLSearchParams` class. Any attempts to use the `searchParams` class will give misleading information about the URL. And any attempts to manipulate it will change the contents of your URL's query string in unintended ways, converting values from a query string serialization (of the type produced by the URL parser) into an `application/x-www-form-urlencoded` serialization.

Some examples of how `url.searchParams` does not allow faithful introspection into the URL record:

```js
let a = 'http://localhost:9999/segment?foo=bar/baz? boo';
let b = 'http://localhost:9999/segment?foo=bar%2Fbaz%3F%20boo';

let urlA = new URL(a);
let urlB = new URL(b);

// Not equal:
console.log(urlA.href); // "http://localhost:9999/segment?foo=bar/baz?%20boo"
console.log(urlB.href); // "http://localhost:9999/segment?foo=bar%2Fbaz%3F%20boo"

console.log(urlA.search); // "?foo=bar/baz?%20boo"
console.log(urlB.search); // "?foo=bar%2Fbaz%3F%20boo"

// Equal:
console.log(urlA.searchParams.get("foo")); // "bar/baz? boo"
console.log(urlB.searchParams.get("foo")); // "bar/baz? boo"

// Equal, but both different from search:
console.log(urlA.searchParams.toString()); // "foo=bar%2Fbaz%3F+boo"
console.log(urlB.searchParams.toString()); // "foo=bar%2Fbaz%3F+boo"
```

Some examples of how using `url.searchParams` for mutation will cause unintended changes to your URL record:

```js
const url = new URL('http://httpbin.org/anything?a=~');

console.log(url.href);   // "http://httpbin.org/anything?a=~"
console.log(url.search); // "?a=~"

// This should be a no-op, but it is not:
url.searchParams.set("a", url.searchParams.get("a"));

console.log(url.href); // "http://httpbin.org/anything?a=%7E"
```

### Solution

[...]

PR #495, opened 2020-05-05
Enumerate differences between URLSearchParams and search

### LINKED ISSUE #18: Investigate aligning application/x-www-form-urlencoded with URL's query
There's already some differences with respect to escaping and #17 will enlarge that gap. Ideally they are the same I think.

annevk:
Closes #18. Follow-up: #491.

I should probably add some editor warnings in case we ever change one of these four enumerations.

***
<a href="https://whatpr.org/url/495.html" title="Last updated on May 6, 2020, 10:20 AM UTC (c7ada96)">Preview</a> | <a href="https://whatpr.org/url/495/96ead3f...c7ada96.html" title="Last updated on May 6, 2020, 10:20 AM UTC (c7ada96)">Diff</a>

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -3106,11 +3106,24 @@ params.toString() // "key=730d67"</code></pre>
 </div>
 
 <div class=note>
- <p>As {{URLSearchParams}} objects use the <a><code>application/x-www-form-urlencoded</code></a>
- format underneath there are some difference with how it encodes certain code points compared to
- {{URL/search}}.
+ <p>As a {{URLSearchParams}} object uses the <a><code>application/x-www-form-urlencoded</code></a>
+ format underneath there are some difference with how it encodes certain code points compared to a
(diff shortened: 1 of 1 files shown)


url.bs
```diff
@@ -3102,6 +3102,23 @@ let params = new URLSearchParams({key: "730d67"})
 params.toString() // "key=730d67"</code></pre>
 </div>
 
+<div class=note>
+ <p>As {{URLSearchParams}} objects use the <a><code>application/x-www-form-urlencoded</code></a>
+ format underneath there are some difference with how it encodes certain code points compared to
+ {{URL/search}}.
```
domenic, line 3108: and href

annevk: maybe just say URL then?

domenic: True, although I think explicitly comparing the `searchParams` API to the `href` and `search` APIs is helpful. But if there are examples then it's less important.

Issue #438, opened 2019-05-02
can't parse urls starting with xn--


Janpot:
Can't seem to parse urls like `http://xn--abc.com`. This seems to work in browsers though.
I've been digging through the code and specs a bit.

It looks like `tr46.toASCII` returns an error. Digging further, it looks like it should implement this spec: https://www.unicode.org/reports/tr46/#Processing. But that seems to say:
> Even if an error occurs, the conversion of the string is performed as much as is possible.

And it says

> If the label starts with “xn--”:
> Attempt to convert the rest of the label to Unicode according to Punycode [RFC3492]. If that conversion fails, record that there was an error, and continue with the next label. Otherwise replace the original label in the string by the results of the conversion.

The url spec seems to dictate (https://url.spec.whatwg.org/#idna)

> If result is a failure value, validation error, return failure.

I feel like this should be possible though, tr46 seems quite ambiguous as to what's recoverable and what not. 

I came across an example that renders and parses in the browser but seems to fail the parsing algorithm: http://xn--12cr4aua8bifvs3aljr6edb1al1vlg1a.blogspot.com (disclaimer: I am in no way connected to this url or the content of the site, it just passed by our systems)

In any case, I'm not super experienced in reading these specs, so take the previous with an appropriate grain of salt. It just seems strange to me that urls can render in a browser, but fail parsing them according to the spec.

EDIT:

forgot to mention, when I say "I've been digging through the code" I'm talking about https://github.com/jsdom/whatwg-url. FWIW, the node.js native url parser seems to behave the same way.

annevk:
Thank you for reporting this. Unfortunately, I suspect you're correct and this is not something we have adequately tested for thus far.

Interesting cases (Live URL Viewer links for comparison purposes):

* [https://xn--.com/](https://jsdom.github.io/whatwg-url/#url=aHR0cHM6Ly94bi0tLmNvbS8=&base=YWJvdXQ6Ymxhbms=) (jsdom has interesting behavior here)
* [https://xn--a.com/](https://jsdom.github.io/whatwg-url/#url=aHR0cHM6Ly94bi0tYS5jb20v&base=YWJvdXQ6Ymxhbms=)
* [https://xn--ß.com/](https://jsdom.github.io/whatwg-url/#url=aHR0cHM6Ly94bi0tw58uY29tLw==&base=YWJvdXQ6Ymxhbms=) (this does seem to fail everywhere)

[...]

PR #497, opened 2020-05-06
Verify domain is not empty after "domain to ASCII"

rmisev:
The empty host is not allowed in special non-"file" URLs, but
`domain to ASCII` result can be the empty string, even if the
input is not empty, for example:

* the input consists entirely of IDNA ignored code points;
* the input is `xn--`.

The `domain to ASCII` do not fail when it is called from `host parser`,
because `VerifyDnsLength` is false.
So we need additional check after it.

An interesting case is with file URLS, as empty host is allowed. But most
browsers report failure as well. I tested URLs with the `U+00AD`
percent encoded in host:

| URL               | Chrome    | Firefox     | Safari    |
|-------------------|-----------|-------------|-----------|
| `file://%C2%AD/p` | TypeError | `file:///p` | TypeError |
| `http://%C2%AD/p` | TypeError | TypeError   | TypeError |

Note. To find more IDNA ignored code points refer to IDNA Mapping Table:
https://unicode.org/reports/tr46/#IDNA_Mapping_Table

   * …
   * …
   * https://github.com/web-platform-tests/wpt/pull/23432
- [ ] [Implementation bugs](https://github.com/whatwg/meta/blob/master/MAINTAINERS.md#handling-pull-requests) are filed:
   * Chrome: …
   * Firefox: …
   * Safari: …

***
<a href="https://whatpr.org/url/497.html" title="Last updated on May 6, 2020, 6:24 PM UTC (62c5129)">Preview</a> | <a href="https://whatpr.org/url/497/5d17e7f...62c5129.html" title="Last updated on May 6, 2020, 6:24 PM UTC (62c5129)">Diff</a>

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -397,6 +397,8 @@ decisions made, i.e. whether to use the <a>same site</a> or <a>schemelessly same
 
  <li><p>If <var>result</var> is a failure value, <a>validation error</a>, return failure.
 
+ <li><p>If <var>result</var> is the empty string, <a>validation error</a>, return failure.
+
  <li><p>Return <var>result</var>.
 </ol>
 
@@ -503,8 +505,6 @@ string <var>input</var> with an optional boolean <var>isNotSpecial</var>, and th
  <a>domain to ASCII</a> on <var>domain</var>.
 
  <li><p>If <var>asciiDomain</var> is failure, <a>validation error</a>, return failure.
-
- <li><p>If <var>asciiDomain</var> is the empty string, <a>validation error</a>, return failure.
 
  <li><p>If <var>asciiDomain</var> contains a <a>forbidden host code point</a>,
  <a>validation error</a>, return failure.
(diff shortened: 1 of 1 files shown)


annevk:
See also #438.

I think we should add this check in https://url.spec.whatwg.org/#concept-domain-to-ascii so that a domain also cannot be valid if it ends up empty.

rmisev:
Hmm, in the valid domain check https://whatpr.org/url/497.html#valid-domain `domain to ASCII` is called with `beStrict = true`, i. e. `VerifyDnsLength = true`, so will fail anyway on empty result.

annevk:
Ah fair, I guess I'd still like to include this check there, but I'd be okay with making it conditional upon beStrict. At some point we have to revise the general ToASCII operation and it would be good if all the weirdness was consolidated.

rmisev:
OK, changed.

Issue #296, opened 2017-04-21
"UTF-8 percent encode c using the path percent-e..."


Hixie:
https://url.spec.whatwg.org/commit-snapshots/488c459d9e4245a3f6bf087e7dcd2c7e91487ac5/#url-parsing

> UTF-8 percent encode c using the path percent\-encode set, and append the result to buffer.

It's not at all clear when reading the parser how a path segment consisting of "%62[" should end up. If I'm reading it right, it should end up as "%25%36%32%5B", which doesn't seem right.

TimothyGu:
None of `%`, `6`, `2`, `[` are part of the [path percent-encode set](https://url.spec.whatwg.org/commit-snapshots/488c459d9e4245a3f6bf087e7dcd2c7e91487ac5/#path-percent-encode-set), so they will get appended to _buffer_ verbatim.

zcorpan:
Would it be clearer if the step

> If codePoint is not in percentEncodeSet, then return codePoint.

was taken out of "UTF-8 percent encode" algorithm? And the places that call UTF-8 percent encode first check if *c* is in the relevant encode set?

Hixie:
Oh wow yeah that's really unclear. The "UTF-8 percent encode" algorithm, in normal operation, generally does not encode? Very confusing. :-)

GPHemsley:
The "UTF-8 percent encode" algorithm also appears to be unclear as to whether it's operating on code points or bytes. What's its return type?

>To **UTF-8 percent encode** a _codePoint_, using a _percentEncodeSet_, run these steps:
> 1. If _codePoint_ is not in _percentEncodeSet_, then return _codePoint_.
> 1. Let _bytes_ be the result of running **UTF-8 encode** on _codePoint_.
> 1. **Percent encode** each byte in _bytes_, and then return the results concatenated, in the same order.

TimothyGu:
Well *codePoint* should be self-evidently a code point. **UTF-8 encode** converts the code point into a byte sequence *bytes*. **Percent encode** then converts every *byte* into a [scalar value string](https://infra.spec.whatwg.org/#scalar-value-string) (in fact a [percent-encoded byte](https://url.spec.whatwg.org/#percent-encoded-byte), a special type of string). So the return type of **UTF-8 percent encode** is a string, while it takes in a code point.

GPHemsley:
[...]

PR #500, opened 2020-05-07
Editorial: minor tweaks to percent-encoded bytes

annevk:
Helps with #296.

***
<a href="https://whatpr.org/url/500.html" title="Last updated on May 8, 2020, 6:23 AM UTC (254b3c7)">Preview</a> | <a href="https://whatpr.org/url/500/c76c15f...254b3c7.html" title="Last updated on May 8, 2020, 6:23 AM UTC (254b3c7)">Diff</a>

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -116,17 +116,17 @@ error.
 <h3 id=percent-encoded-bytes>Percent-encoded bytes</h3>
 
 <p>A <dfn>percent-encoded byte</dfn> is U+0025 (%), followed by two <a>ASCII hex digits</a>.
-Sequences of <a lt="percent-encoded byte">percent-encoded bytes</a>, after conversion to bytes,
+Sequences of <a lt="percent-encoded byte">percent-encoded bytes</a>, <a>isomorphic encoded</a>,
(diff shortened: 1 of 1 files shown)


url.bs
```diff
@@ -116,17 +116,17 @@ error.
 <h3 id=percent-encoded-bytes>Percent-encoded bytes</h3>
 
 <p>A <dfn>percent-encoded byte</dfn> is U+0025 (%), followed by two <a>ASCII hex digits</a>.
-Sequences of <a lt="percent-encoded byte">percent-encoded bytes</a>, after conversion to bytes,
+Sequences of <a lt="percent-encoded byte">percent-encoded bytes</a>, <a>isomorphic encoded</a>,
 should not cause <a>UTF-8 decode without BOM or fail</a> to return failure.
```
domenic, line 120: This sentence is pretty unclear to me both before and after the change, so I'm unsure how to evaluate whether "isomorphic encoded" was what is meant. The usage of "should" also seems off.

annevk: Thanks, that was indeed wrong.

annevk: I'll file a follow-up to investigate further clarification.

Issue #406, opened 2018-07-27
Named validation errors


TRowbotham:
It would be neat if the URL parser had named validation errors like what was done for the HTML Tokenizer in whatwg/html#2701.

annevk:
FWIW, I'd be supportive of someone adding this to the standard and will happily help out with review.

TRowbotham:
I have this mostly done. I'll try to finish it up and post something for review.

alercah:
My 2c: This is an excellent example of an issue for which my opinion changes depending on whether this is in the context of a spec for a data format (the URL string) or for an API (the web API exposed by browsers).

In any particular API, I think named errors are excellent, and I would support adding them to the web API. By contrast, it's not immediately clear to me that there would be value in trying to define these errors on the data format that is a URL string, as the exact manner of error handling is something typically left to an implementation decision, particularly given that error handling is often a context specific and language specific thing. For instance, error design for human consumption and for programmatic use often differ, and I think that the URL format should be agnostic to such concerns.

PR #502, opened 2020-05-08
Named validation errors

TRowbotham:
This addresses #406. 

By my count, there are 54 different places where a `validation error` can occur with the majority of them being unique. This adds a table with descriptive error codes for each unique validation error, which also has an associated description of the error. 

- Would there by value in adding a column indicating whether the validation error can cause the parser to fail?
- I am considering adding example input that would trigger each validation error. Is this something you think would have value?
- The `unexpected-windows-drive-letter-host` needs an error description still.
- I wasn't sure how best to link the validation errors in each spot, so I just ended up doing `<a><unique error name></a> <a>validation error</a>` everywhere. I had considered a prefix or suffix of `validation-error` to every code, but felt it was overly verbose, though I'm not opposed to it if someone feels its more appropriate.
- "Error Message" should probably be renamed to "Error Description".
- Some of the error descriptions could probably use a little more detail, but I wanted to get this initial draft posted.

***
<a href="https://whatpr.org/url/502.html" title="Last updated on Feb 17, 2023, 11:08 AM UTC (1fddbe3)">Preview</a> | <a href="https://whatpr.org/url/502/8653878...1fddbe3.html" title="Last updated on Feb 17, 2023, 11:08 AM UTC (1fddbe3)">Diff</a>

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -100,28 +100,15 @@ valid input. User agents, especially conformance checkers, are encouraged to rep
   <tr>
    <td><dfn id=validation-error-domain-to-ascii>domain-to-ASCII</dfn>
    <td>
-    <p>The result of <a abstract-op lt=ToASCII>Unicode ToASCII</a> records an error while processing
-    the input domain.
-    <p class=note>[[!UTS46]] conformance does not require the reporting of precise errors, only that
-    an error has occurred. If the [[!UTS46]] implementation reports precise error codes, user agents
-    are encouraged pass those codes along.
-   <td>Yes
-  <tr>
-   <td><dfn>domain-to-ASCII-empty</dfn>
-   <td>
-    <p>The result of <a abstract-op lt=ToASCII>Unicode ToASCII</a> returns an empty string. This
-    could have been caused by:
-    <ul>
-     <li>Input consists of all ignorable code points.
-     <li>Input is the string "<code>xn--</code>".
-     <li>Input is the empty string and the <i>VerifyDnsLength</i> parameter is false.
-    </ul>
+    <p><a abstract-op lt=ToASCII>Unicode ToASCII</a> records an error or returns the empty string.
+    [[UTS46]]
+    <p class=note>If details about <a abstract-op lt=ToASCII>Unicode ToASCII</a> errors are
+    recorded, user agents are encouraged to pass those along.
    <td>Yes
   <tr>
    <td><dfn>domain-to-Unicode</dfn>
    <td>
-    <p>The result of <a abstract-op lt=ToUnicode>Unicode ToUnicode</a> returns an error while
-    processing the input domain.
+    <p><a abstract-op lt=ToUnicode>Unicode ToUnicode</a> records an error. [[UTS46]]
     <p class=note>The same considerations as with <a>domain-to-ASCII</a> apply.
    <td>·
  <tbody>
@@ -951,8 +938,8 @@ concepts.
  <li><p>If <var>result</var> is a failure value, <a>domain-to-ASCII</a> <a>validation error</a>,
  return failure.
 
- <li><p>If <var>result</var> is the empty string, <a>domain-to-ASCII-empty</a>
- <a>validation error</a>, return failure.
+ <li><p>If <var>result</var> is the empty string, <a>domain-to-ASCII</a> <a>validation error</a>,
+ return failure.
 
  <li><p>Return <var>result</var>.
 </ol>
--- a/url.bs
+++ b/url.bs
@@ -96,6 +96,179 @@ valid input. User agents, especially conformance checkers, are encouraged to rep
    <th>Failure
  <tbody>
   <tr>
+   <th colspan=3 scope=rowgroup><a href=#idna>IDNA</a>
+  <tr>
+   <td><dfn id=validation-error-domain-to-ascii>domain-to-ASCII</dfn>
+   <td>
+    <p>The result of <a abstract-op lt=ToASCII>Unicode toASCII</a> records an error while processing
+    the input domain.
+    <p class=note>[[!UTS46]] conformance does not require the reporting of precise errors, only that
+    an error has occurred. If the [[!UTS46]] implementation reports precise error codes, user agents
+    are encouraged pass those codes along.
+   <td>✅
+  <tr>
+   <td><dfn>domain-to-ASCII-empty</dfn>
+   <td>
+    <p>The result of <a abstract-op lt=ToASCII>Unicode toASCII</a> returns an empty string. This
+    could have been caused by:
+    <ul>
+     <li>Input consists of all ignorable code points.
+     <li>Input is the string "<code>xn--</code>".
+     <li>Input is the empty string and the <i>VerifyDnsLength</i> parameter is false.
+    </ul>
+   <td>✅
+  <tr>
+   <td><dfn>domain-to-Unicode</dfn>
+   <td>
+    <p>The result of <a abstract-op lt=ToUnicode>Unicode toUnicode</a> returns an error while
+    processing the input domain.
+    <p class=note>The same considerations as with <a>domain-to-ASCII</a> apply.
+   <td>❌
+ <tbody>
+  <tr>
+   <th colspan=3 scope=rowgroup><a href=#host-parsing>Host parsing</a>
+  <!-- host parser -->
+  <tr>
+   <td><dfn>unclosed-IPv6-address</dfn>
+   <td>
+    <p>An IPv6 address is missing the closing U+005D (]).
+    <p class=example id=example-unclosed-ipv6-address>"<code>https://[::1</code>"
+   <td>✅
+  <tr>
+   <td><dfn id=validation-error-forbidden-domain-code-point>forbidden-domain-code-point</dfn>
+   <td>
+    <p>The input's host contains a <a>forbidden domain code point</a>.
+    <div class=example id=example-forbidden-domain-code-point>
+     <p>Hosts are <a for=string>percent-decoded</a> before being processed when the URL
+     <a>is special</a>, which would result in the following host portion becoming
+     "<code>exa#mple.org</code>".
+     <p>"<code>https://exa%23mple.org</code>"
+    </div>
+   <td>✅
+  <!-- IPv4 parser -->
+  <tr>
+   <td><dfn>empty-IPv4-part</dfn>
+   <td>
+    <p>Input to the <a>IPv4 parser</a> ends with a U+002E.
+    <p class=example id=example-empty-ipv4-part>"<code>https://127.0.0.1./</code>"
+   <td>❌
+  <tr>
+   <td><dfn>too-many-IPv4-parts</dfn>
+   <td>
+    <p>An <a for=/>IPv4 address</a> consists of 4 parts.
+    <p class=example id=example-too-many-ipv4-parts>"<code>https://1.2.3.4.5/</code>"
+   <td>✅
+  <tr>
+   <td><dfn>non-numeric-IPv4-part</dfn>
+   <td>
+    <p>An IPv4 part was not numeric.
+    <p class=example id=example-non-numeric-ipv4-part>"<code>https://test.42</code>"
+   <td>✅
+  <tr>
+   <td><dfn>non-decimal-IPv4-part</dfn>
+   <td>
+    <p>The IPv4 address contains numbers expressed using hexadecimal or octal digits.
+    <p class=example id=example-non-decimal-ipv4-part>"<code>https://127.0.0x0.1</code>"
+   <td>❌
+  <tr>
+   <td><dfn>IPv4-part-out-of-range</dfn>
+   <td>
+    <p>An IPv4 part exceeds 255.
+    <p class=example id=example-ipv4-part-out-of-range>"<code>https://255.255.4000.1</code>"
+   <td>✅ (only if applicable to the last IPv4 part)
+  <!-- IPv6 parser -->
+  <tr>
+   <td><dfn>invalid-compressed-IPv6-address</dfn>
+   <td>
+    <p>An IPv6 address begins with improper compression.
+    <p class=example id=example-invalid-compressed-ipv6-address>"<code>https://[:1]</code>"
+   <td>✅
+  <tr>
+   <td><dfn>IPv6-too-many-pieces</dfn>
+   <td>
+    <p>An IPv6 address contains more than 8 pieces.
+    <p class=example id=example-ipv6-too-many-pieces>"<code>https://[1:2:3:4:5:6:7:8:9]</code>"
+   <td>✅
+  <tr>
+   <td><dfn>IPv6-multiple-compression</dfn>
+   <td>
+    <p>An IPv6 address is compressed in more than one spot.
+    <p class=example id=example-ipv6-multiple-compression>"<code>https://[1::1::1]</code>"
+   <td>✅
+  <tr>
+   <td><dfn>IPv4-in-IPv6-empty-part</dfn>
+   <td>
+    <p>An IPv6 address that contains an IPv4 address has an empty part in the IPv4 address.
+    <p class=example id=example-ipv4-in-ipv6-empty-part>"<code>https://[ffff::.0.0.1]</code>"
+   <td>✅
+  <tr>
+   <td><dfn>IPv4-in-IPv6-too-many-pieces</dfn>
+   <td>
+    <p>An IPv6 address contains an IPv4 address and the IPv6 address has more than 6 pieces.
+    <p class=example id=example-ipv4-in-ipv6-too-many-pieces>"<code>https://[1:1:1:1:1:1:1:127.0.0.1]</code>"
+   <td>✅
+  <tr>
+   <td><dfn>IPv4-in-IPv6-too-many-parts</dfn>
+   <td>
+    <p>An IPv6 address contains an IPv4 address and the IPv4 address has more than 4 parts.
+    <p class=example id=example-ipv4-in-ipv6-too-many-parts>"<code>https://[ffff::127.0.0.1.2]</code>"
+   <td>✅
+  <tr>
+   <td><dfn>IPv4-in-IPv6-unexpected-code-point</dfn>
+   <td>
+    <p>An IPv6 address contains an IPv4 address of which a part contain a code point other than an
+    <a>ASCII digits</a> or is the empty string.
+    <p class=example id=example-ipv4-in-ipv6-unexpected-code-point>"<code>https://[ffff::127.0.xyz.1]</code>"
+   <td>✅
+  <tr>
+   <td><dfn>IPv4-in-IPv6-invalid-first-part</dfn>
+   <td>
+    <p>The first part of an IPv4 address that is contained within an IPv6 address is 0.
+    <p class=example id=example-ipv4-in-ipv6-invalid-first-part>"<code>https://[ffff::0.0.0.1]</code>"
+   <td>✅
+  <tr>
+   <td><dfn>IPv4-in-IPv6-part-out-of-range</dfn>
+   <td>
+    <p>An IPv4 address contained within an IPv6 address contains a part that exceeds 255.
+    <p class=example id=example-ipv4-in-ipv6-part-out-of-range>"<code>https://[ffff::127.0.0.4000]</code>"
+   <td>✅
+  <tr>
+   <td><dfn>IPv4-in-IPv6-too-few-parts</dfn>
+   <td>
+    <p>An IPv4 address contained within an IPv6 address does not contain exactly 4 parts.
+    <p class=example id=example-ipv4-in-ipv6-too-few-parts>"<code>https://[ffff::127.0.0]</code>"
+   <td>✅
+  <tr>
+   <td><dfn>IPv6-unexpected-EOF</dfn>
+   <td>
+    <p>An IPv6 address unexpectedly ends.
+    <p class=example id=example-ipv6-unexpected-eof>"<code>https://[1:2:3:]</code>"
+   <td>✅
+  <tr>
+   <td><dfn>IPv6-unexpected-delimiter</dfn>
+   <td>
+    <p>An IPv6 address contains a code point that is neither an <a>ASCII hex digit</a> nor a
+    U+003A (:).
+    <p class=example id=example-ipv6-unexpected-delimiter>"<code>https://[1:2:3!:4]</code>"
+   <td>✅
+  <tr>
+   <td><dfn>IPv6-too-few-pieces</dfn>
+   <td>
+    <p>An uncompressed IPv6 address contains fewer than 8 pieces.
+    <p class=example id=example-ipv6-too-few-pieces>"<code>https://[1:2:3]</code>"
+   <td>✅
+  <!-- opaque-host parser -->
+  <tr>
+   <td><dfn id=validation-error-forbidden-host-code-point>forbidden-host-code-point</dfn>
+   <td>
+    <p>When an <a>opaque host</a> (in a URL that <a>is not special</a>) contains a
+    <a>forbidden host code point</a>.
+   <p class=example id=example-forbidden-host-code-point>"<code>foo://exa[mple.org</code>"
+   <td>✅
+ <tbody>
+  <tr>
+   <th colspan=3 scope=rowgroup><a href=#url-parsing>URL parsing</a>
+  <tr>
    <td><dfn>unexpected-C0-control-or-space</dfn>
    <td>
     <p>The input to the <a>URL parser</a> contains a leading or trailing <a>C0 control or space</a>. The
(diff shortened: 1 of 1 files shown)


domenic:
This looks amazing! Giving my opinions on some of your questions, but you should probably wait for @annevk before acting on them:

> Would there by value in adding a column indicating whether the validation error can cause the parser to fail?

IMO yes. I'm thinking a "Fatal" column which is either empty or one of the unicode checkmarks.

> I am considering adding example input that would trigger each validation error. Is this something you think would have value?

IMO yes. Doing so in a way that makes things non-overwhelming might be a bit tricky. E.g. I think another column would probably not work well. Maybe a second row (making each error code `rowspan="2") with `class="example"`, similar to https://html.spec.whatwg.org/#usage-summary-2 ? Or maybe just putting in `<p class="example">` wouldn't be so bad.

Relatedly, although I imagine you are following the pattern from HTML, maybe the descriptions could be tightened up by removing "This error occurs when".

> I wasn't sure how best to link the validation errors in each spot,

I like what you did. It provides a convenient link to the concept of validation error, which is important as people can be confused by parse error vs. validation error.

url.bs
```diff
@@ -85,6 +85,170 @@ valid input. User agents, especially conformance checkers, are encouraged to rep
...
 
+<table>
+ <thead>
+  <tr>
+   <th>Error Code
+   <th>Error Message
+ <tbody>
+  <tr>
+   <td><dfn>unexpected-c0-control-or-space</dfn>
+   <td>This error occurs when the input to the URL parser contains leading or trailing
```
annevk, line 96: If we make the header "Description" as you suggest we could make this "The input to the URL parser ..." and not have the repeating introductory phrase.

url.bs
```diff
@@ -85,6 +85,170 @@ valid input. User agents, especially conformance checkers, are encouraged to rep
...
+   <a>C0 control or space</a> code points. The URL parser subsequently strips any matching code
+   points.
+  <tr>
+   <td><dfn>unexpected-ascii-tab-or-newline</dfn>
```
annevk, line 100: I would prefer uppercasing ASCII here. Or is there a reason for all lowercase names? Similar for IPv4 and such.

url.bs
```diff
@@ -85,6 +85,170 @@ valid input. User agents, especially conformance checkers, are encouraged to rep
...
+  <tr>
+   <td><dfn>wrong-solidus-special-scheme</dfn>
+   <td>This error occurs when a URL with a <a>special scheme</a> uses the U+005C (\) reverse
+   solidus code point instead of the U+002F (/) code point.
+  <tr>
+   <td><dfn>missing-solidus-before-authority</dfn>
+   <td>This error occurs if the URL <a>includes credentials</a> and they are not preceeded by
+   two "<code>//</code>" U+002F (/) code points.
+  <tr>
+   <td><dfn>unexpected-@-code-point</dfn>
```
annevk, line 133: I think this should be commercial-at if we're not using / for solidus.

url.bs
```diff
@@ -85,6 +85,170 @@ valid input. User agents, especially conformance checkers, are encouraged to rep
...
+   <td><dfn>unexpected-host-file-scheme</dfn>
+   <td>This error occurs when the URL's <a>scheme</a> is changed to the "<code>file</code>"
+   <a>scheme</a> and the existing URL has a <a for=url>host</a>.
+  <tr>
+   <td><dfn>unexpected-empty-path</dfn>
+   <td>This error occurs when a URL with a "<code>file</code>" <a>scheme</a> contains an empty
+   <a>path</a>.
+  <tr>
+   <td><dfn>invalid-url-code-point</dfn>
+   <td>This error occurs when a code point that is not a <a>url code point</a> or U+0025 (%) in the
```
annevk, line 176: URL code point*

url.bs
```diff
@@ -85,6 +85,170 @@ valid input. User agents, especially conformance checkers, are encouraged to rep
...
+  <tr>
+   <td><dfn>unescaped-percent-sign</dfn>
+   <td>This error occurs when a U+0025 (%) code point is found that is not followed by two
+   <a>ASCII hex digits</a>, representing a <a>percent-encoded byte</a>, in the URL's <a>path</a>,
+   <a>query</a>, or <a for=url>fragment</a>.
+  <tr>
+   <td><dfn>unclosed-ipv6-address</dfn>
+   <td>This error occurs when an IPv6 address is missing the closing U+005D (]) code point.
+  <tr>
+   <td><dfn>ascii-idn-failure</dfn>
```
annevk, line 187: There are more errors here now so you'll have to split this (and rebase). I suggest calling this after IDNA, ToASCII, or domain-to-ASCII and not add a new term (IDN).

url.bs
```diff
@@ -85,6 +85,170 @@ valid input. User agents, especially conformance checkers, are encouraged to rep
...
+   <td>This error occurs when transforming the domain to ASCII using IDN records an error.
+  <tr>
+   <td><dfn>host-forbidden-code-point</dfn>
+   <td>This error occurs when the host contains a <a>forbidden host code point</a>.
+  <tr>
+   <td><dfn>unexpected-non-decimal-number</dfn>
+   <td>This occurs when the IPv4 address contains numbers expressed using hexadecimal or octal
+   digits.
+  <tr>
+   <td><dfn>ipv4-octet-out-of-range</dfn>
```
annevk, line 197: Let's use part and not octet for IPv4 parts (throughout). This matches the parser variable naming.

TRowbotham:
Thanks for the feedback. I'll work on adding a new "Fatal" column as domenic suggested and rewording the descriptions to try not to repeat the "This error occurs when...". 

I agree that an additional column probably isn't the best approach for the example input. The HTML [parse errors](https://html.spec.whatwg.org/multipage/parsing.html#parse-errors) table appears to put notes and examples right under the error description text, so I'll experiment with doing that and hopefully things don't get too cluttered.

url.bs
TRowbotham: No particular reason. Was mostly following what was done for the HTML [parse errors](https://html.spec.whatwg.org/multipage/parsing.html#parse-errors) since they were all lowercase and didn't use uppercase for terms like "HTML" or "EOF".

url.bs
TRowbotham: Makes sense.

url.bs
annevk: Thanks, @domenic what do you think? Should we adjust HTML a bit there?

domenic: I like the all-lowercase style personally.

In general really unclear what these identifiers are to be used for and what conventions they should follow, so we're just making up our own rules. Other systems might use numeric error codes, or SHOUTY_CASE variants which could be used as variable names in most programming languages. Or, we could ask, why are we using dashes instead of spaces?

Similar questions have come up in the unrelated context of Web IDL enumerations. There we've mostly settled on lower-kebab-case, I think.

domenic:
> The HTML parse errors table appears to put notes and examples right under the error description text, so I'll experiment with doing that and hopefully things don't get too cluttered.

One thing that will declutter things a bit is adding `<p>` around each description, which will introduce some nicer margins.

url.bs
annevk: I see, but these are non-string enumerations. I guess I can go along with that.

annevk:
@TRowbotham if this needs rebasing at some point due to other ongoing changes I'm happy to do that work for you.

url.bs
TRowbotham: Is the consensus to stick with all lowercase?

annevk: Yeah.

TRowbotham:
I think I broke this while rebasing :(

TRowbotham:
@annevk, I think this is ready now. It just needs a rebase, which I'll leave up to you so I don't botch it again.

annevk:
@TRowbotham how are the rows ordered?

TRowbotham:
They are grouped by section in the spec. I started at section 4.4 URL Parsing and worked my way backwards through the sections.

annevk:
For port-out-of-range a problem is that URL-port string should define the upper limit. If it defines the upper limit, do we still want a separate enum value? I guess we do, but the description would have to change as it would match invalid-port otherwise.

karwa:
I think the `missing-credentials` error is incorrect. That error occurs if there is an `@` but no host. `https://@test.com` should not cause the parser to fail, but `https://username:password@` should (and does, in the Live URL viewer in Safari 13).

Perhaps it should be replaced by something like 'unexpected credentials without host' (similar to 'unexpected port without host')?

annevk:
The IPv4 parser still has two unclassified validation errors.

Otherwise this has been rebased. I adjusted casing to my liking. I addressed @karwa's observation. I want to give it another pass and address those missing cases, but it's getting late.

annevk:
I want #739 to land first and rebase this on top.

I also want to divide the table into 3 row groups. Tentatively titled "Hosts", "URLs", and "Opaque hosts and URLs". And then put the errors in order of appearance.

annevk:
This is now ready for review. All validation errors should be covered. Please don't hold back on nits!

TRowbotham:
Thank you for driving this to completion!

Overall, I think this looks pretty good. I think the IPv4 and IPv6 validation error names could use a normalization pass. Currently, there are a few with similar names, but slightly different word order such as:
* too-many-IPv4-parts 
* IPv6-too-many-pieces
* IPv4-in-IPv6-too-many-parts

url.bs
```diff
@@ -88,6 +88,334 @@ valid input. User agents, especially conformance checkers, are encouraged to rep
...
+  <tr>
+   <td><dfn>IPv4-in-IPv6-too-many-parts</dfn>
+   <td>
+    <p>An IPv6 address contains an IPv4 address and the IPv4 address has more than 4 parts.
+    <p class=example id=example-ipv4-in-ipv6-too-many-parts>"<code>https://[ffff::127.0.0.1.2]</code>"
+   <td>✅
+  <tr>
+   <td><dfn>IPv4-in-IPv6-unexpected-code-point</dfn>
+   <td>
+    <p>An IPv6 address contains an IPv4 address of which a part contain a code point other than an
```
TRowbotham, line 219: Should probably read ", of which, a part contains code points other than ASCII digits or is an empty string." or feel free to completely reword it.

url.bs
```diff
@@ -849,21 +1180,22 @@ return value of the <a for=/>host parser</a> is an <a for=/>IPv4 address</a>.
...
 
- <li><p>If any item in <var>numbers</var> is greater than 255, <a>validation error</a>.
+ <li><p>If any item in <var>numbers</var> is greater than 255, <a>IPv4-part-out-of-range</a>
+ <a>validation error</a>.
 
  <li><p>If any but the last <a for=list>item</a> in <var>numbers</var> is greater than 255, then
  return failure.
 
  <li><p>If the last <a for=list>item</a> in <var>numbers</var> is greater than or equal to
- 256<sup>(5 &minus; <var>numbers</var>'s <a for=list>size</a>)</sup>, <a>validation error</a>,
```
TRowbotham, line 865: Is the validation error that was removed here always caught by Step 6 above, where it checks for any number > 255?

annevk: Yeah, numbers cannot contain more than 4 items.

url.bs
```diff
@@ -1016,10 +1352,11 @@ actually doing that with the editors of this document first.
          <li><p>If <a>c</a> is a U+002E (.) and <var>numbersSeen</var> is less than 4, then increase
          <var>pointer</var> by 1.
 
-         <li>Otherwise, <a>validation error</a>, return failure.
+         <li>Otherwise, <a>IPv4-in-IPv6-too-many-parts</a> <a>validation error</a>, return failure.
```
annevk, line 1355: This can also happen when c is not `.`. E.g., `https://[ffff::12xyz]`. Should we perhaps reuse IPv4-in-IPv6-unexpected-code-point here and just call out all the different cases in examples?

url.bs
```diff
@@ -1031,13 +1368,14 @@ actually doing that with the editors of this document first.
          <li>
           <p>If <var>ipv4Piece</var> is null, then set <var>ipv4Piece</var> to <var>number</var>.
 
-          <p>Otherwise, if <var>ipv4Piece</var> is 0, <a>validation error</a>, return failure.
+          <p>Otherwise, if <var>ipv4Piece</var> is 0, <a>IPv4-in-IPv6-invalid-first-part</a>
```
annevk, line 1371: This name is also wrong. The error here is the leading zero. This can happen in any part, not just the first.

TRowbotham:
I stood up a small [demo](http://url-demo.trowbotham.test) implementation, similar to the jsdom version, you can use it to quickly check validation errors if that helps with the review. 

The git branch of the url parser the demo is using isn't published at the moment, but I can publish the branch if you would find the source code useful. The error highlighting could probably use some refinement, but it should do the job for reporting the validation error.

url.bs
TRowbotham: `https://[ffff::12xyz]`[(demo)](https://url-demo.trowbotham.com/?url=aAB0AHQAcABzADoALwAvAFsAZgBmAGYAZgA6ADoAMQAyAHgAeQB6AF0A) reports IPv6-unexpected-delimiter for me, but modifying it slightly to `https://[ffff::12.12xyz]`[(demo)](https://url-demo.trowbotham.com/?url=aAB0AHQAcABzADoALwAvAFsAZgBmAGYAZgA6ADoAMQAyAC4AMQAyAHgAeQB6AF0A) shows the problem as the parser needs to encounter at least one `.` before taking that code path.

Your suggestion of using IPv4-in-IPv6-unexpected-code-point seems reasonable to me. I'd be interested to hear other peoples opinions.

url.bs
TRowbotham: Good catch. Perhaps changing the name to something like "IPv4-in-IPv6-leading-zero" or "leading-zero-in-IPv4-part-in-IPv6"

url.bs
```diff
@@ -88,6 +88,334 @@ valid input. User agents, especially conformance checkers, are encouraged to rep
...
+   <th>Error type
+   <th>Error description
+   <th>Failure
+ <tbody>
+  <tr>
+   <th colspan=3 scope=rowgroup><a href=#idna>IDNA</a>
+  <tr>
+   <td><dfn id=validation-error-domain-to-ascii>domain-to-ASCII</dfn>
+   <td>
+    <p>The result of <a abstract-op lt=ToASCII>Unicode toASCII</a> records an error while processing
```
domenic, line 103:
```suggestion
    <p>The result of <a abstract-op lt=ToASCII>Unicode ToASCII</a> records an error while processing
```

domenic, line 103:
```suggestion
    <p>The result of <a abstract-op lt=ToASCII>Unicode ToASCII</a> records an error while processing
```

url.bs
```diff
@@ -88,6 +88,334 @@ valid input. User agents, especially conformance checkers, are encouraged to rep
...
+    the input domain.
+    <p class=note>[[!UTS46]] conformance does not require the reporting of precise errors, only that
+    an error has occurred. If the [[!UTS46]] implementation reports precise error codes, user agents
+    are encouraged pass those codes along.
+   <td>✅
+  <tr>
+   <td><dfn>domain-to-ASCII-empty</dfn>
+   <td>
+    <p>The result of <a abstract-op lt=ToASCII>Unicode toASCII</a> returns an empty string. This
```
domenic, line 112:
```suggestion
    <p>The result of <a abstract-op lt=ToASCII>Unicode ToASCII</a> returns an empty string. This
```

url.bs
```diff
@@ -88,6 +88,334 @@ valid input. User agents, especially conformance checkers, are encouraged to rep
...
+    <ul>
+     <li>Input consists of all ignorable code points.
+     <li>Input is the string "<code>xn--</code>".
+     <li>Input is the empty string and the <i>VerifyDnsLength</i> parameter is false.
+    </ul>
+   <td>✅
+  <tr>
+   <td><dfn>domain-to-Unicode</dfn>
+   <td>
+    <p>The result of <a abstract-op lt=ToUnicode>Unicode toUnicode</a> returns an error while
```
domenic, line 123:
```suggestion
    <p>The result of <a abstract-op lt=ToUnicode>Unicode ToUnicode</a> returns an error while
```

url.bs
```diff
@@ -88,6 +88,334 @@ valid input. User agents, especially conformance checkers, are encouraged to rep
...
+   <td>
+    <p>The input to the URL parser contains <a>ASCII tab or newlines</a>. The URL parser
+    subsequently strips any matching code points.
+    <p class=example id=example-unexpected-ascii-tab-or-newline>"<code>ht<br>tps://example.org</code>"
+   <td>❌
+  <tr>
+   <td><dfn>invalid-scheme-start</dfn>
+   <td>
+    <p>The first code point of a URL's <a for=url>scheme</a> is not an <a>ASCII alpha</a>.
+    <p class=example id=example-invalid-scheme-start>"3ttps://example.org"
```
domenic, line 289:
```suggestion
    <p class=example id=example-invalid-scheme-start>"<code>3ttps://example.org</code>"
```

url.bs
```diff
@@ -88,6 +88,334 @@ valid input. User agents, especially conformance checkers, are encouraged to rep
...
+     <pre><code class=lang-javascript>
+let url = new URL("💩", "mailto:user@example.org");</code></pre>
+    </div>
+   <td>✅
+  <tr>
+   <td><dfn>relative-URL-missing-beginning-solidus</dfn>
+   <td>
+    <p>The input is a <a>relative-URL String</a> that does not begin with U+002F (/).
+    <pre class=example id=example-relative-url-missing-beginning-solidus><code class="lang-javascript">
+let url = new URL("foo.html", "https://example.org/");</code></pre>
```
domenic, line 326: Why the heck would this be an error? Does this mean `<a href="foo.html">` is an error?

url.bs
```diff
@@ -88,6 +88,334 @@ valid input. User agents, especially conformance checkers, are encouraged to rep
...
+   <th colspan=3 scope=rowgroup><a href=#idna>IDNA</a>
+  <tr>
+   <td><dfn id=validation-error-domain-to-ascii>domain-to-ASCII</dfn>
+   <td>
+    <p>The result of <a abstract-op lt=ToASCII>Unicode toASCII</a> records an error while processing
+    the input domain.
+    <p class=note>[[!UTS46]] conformance does not require the reporting of precise errors, only that
+    an error has occurred. If the [[!UTS46]] implementation reports precise error codes, user agents
+    are encouraged pass those codes along.
+   <td>✅
```
domenic, line 108: Although it may have been my suggestion, I think the green checkmarks / red Xs are confusing, because they feel too much like "success" / "failure".

Maybe "Yes" / "&middot;" similar to https://html.spec.whatwg.org/#linkTypes would work better.

url.bs
```diff
@@ -88,6 +88,334 @@ valid input. User agents, especially conformance checkers, are encouraged to rep
...
+   <td>❌
+  <tr>
+   <td><dfn>unexpected-Windows-drive-letter-host</dfn>
+   <td>
+    <p>The file URL's host is a Windows drive letter.
+    <p class=example id=example-unexpected-windows-drive-letter-host>"<code>file://c:</code>"
+   <td>❌
+ <tbody>
+  <tr>
+   <th colspan=3 scope=rowgroup>URL parsing and <a>opaque-host parser</a>
```
domenic, line 402: I don't really understand why this is split out from the above section.

url.bs
```diff
@@ -88,6 +88,334 @@ valid input. User agents, especially conformance checkers, are encouraged to rep
...
+     <p>Input's <a for=url>scheme</a> is missing, but the <a>base URL</a> has an
+     <a for=url>opaque path</a>.
+     <pre><code class=lang-javascript>
+let url = new URL("💩", "mailto:user@example.org");</code></pre>
+    </div>
+   <td>✅
+  <tr>
+   <td><dfn>relative-URL-missing-beginning-solidus</dfn>
+   <td>
+    <p>The input is a <a>relative-URL String</a> that does not begin with U+002F (/).
```
domenic, line 324:
```suggestion
    <p>The input is a <a>relative-URL string</a> that does not begin with U+002F (/).
```

domenic (commented):
+1 for better consistency in naming the errors, if possible. Some suggestions, none of which I'm 100% confident in:

- All the file: URL specific ones could be prefixed with `file-`.
- All the authority/credentails ones could be prefixed with `authority-`.
- Choose between "unexpected" (e.g. unexpected-C0-control-or-space), "invalid" (e.g. invalid-URL-code-point, invalid-scheme-start), and "forbidden" (e.g. forbidden-host-code-point").

I also wonder about redundancy for some of these. Isn't invalid-URL-code-point a superset of unexpected-C0-control-or-space and unexpected-ASCII-tab-or-newline? Similarly isn't unexpected-at-sign a superset of unexpected-credentials-without-host and missing-solidus-before-authority? Maybe the extra specificity is useful, but I'm not so sure...

url.bs
annevk: Good point. I think this would only be hit by `new URL("https:foo.html", "https://example.com/")`, but that clearly requires a better name.

url.bs
annevk: I wanted to group validation errors per somewhat major section. However these validation errors were reused across Hosts and URLs. I could just put it under Hosts as that's the first appearance, but it might look surprising?

annevk:
I pushed a large number of changes after a fairly lengthy editing session, but it's still not done unfortunately.

url.bs
domenic: Yeah, I think just putting them under URLs would make more sense. Unlike IPv4/IPv6/ToASCII, there's no clear differentiation here IMO.

url.bs
```diff
@@ -88,6 +88,301 @@ valid input. User agents, especially conformance checkers, are encouraged to rep
...
+   <td>Yes
+ <tbody>
+  <tr>
+   <th colspan=3 scope=rowgroup><a href=#url-parsing>URL parsing</a>
+  <tr>
+   <td><dfn>scheme-invalid-code-point</dfn>
+   <td>
+    <p>The URL's <a for=url>scheme</a> contains an invalid code point.
+    <div class=example id=example-scheme-invalid-code-point>
+     TODO: actual setter examples as prior examples were wrong
```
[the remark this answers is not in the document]
TRowbotham: Not emitting validation errors for API inputs seems reasonable.

url.bs
```diff
@@ -88,6 +88,256 @@ valid input. User agents, especially conformance checkers, are encouraged to rep
...
+  <tr>
+   <td><dfn>IPv4-too-many-parts</dfn>
+   <td>
+    <p>An <a for=/>IPv4 address</a> has to consist of 4 parts.
+    <p class=example id=example-ipv4-too-many-parts>"<code>https://1.2.3.4.5/</code>"
+   <td>Yes
+  <tr>
+   <td><dfn>IPv4-non-numeric-part</dfn>
+   <td>
+    <p>An <a for=/>IPv4 address</a> part was not numeric.
```
domenic, line 151:
```suggestion
    <p>An <a for=/>IPv4 address</a> part is not numeric.
```

url.bs
```diff
@@ -88,6 +88,256 @@ valid input. User agents, especially conformance checkers, are encouraged to rep
...
+   <td>
+    <p>The <a for=/>IPv4 address</a> contains numbers expressed using hexadecimal or octal digits.
+    <p class=example id=example-ipv4-non-decimal-part>"<code>https://127.0.0x0.1</code>"
+   <td>·
+  <tr>
+   <td><dfn>IPv4-out-of-range-part</dfn>
+   <td>
+    <p>An <a for=/>IPv4 address</a> part exceeds 255.
+    <p class=example id=example-ipv4-out-of-range-part>"<code>https://255.255.4000.1</code>"
+   <td>Yes (only if applicable to the last part)
```
domenic, line 165:
```suggestion
   <td>Yes<br>(only if applicable to the last part)
```

maybe?

url.bs
```diff
@@ -88,6 +88,256 @@ valid input. User agents, especially conformance checkers, are encouraged to rep
...
+   <td>Yes
+  <tr>
+   <td><dfn>IPv4-in-IPv6-too-many-pieces</dfn>
+   <td>
+    <p>An <a for=/>IPv6 address</a> with <a for=/>IPv4 address</a> syntax: the IPv6 address has more
+    than 6 pieces.
+    <p class=example id=example-ipv4-in-ipv6-too-many-pieces>"<code>https://[1:1:1:1:1:1:1:127.0.0.1]</code>"
+   <td>Yes
+  <tr>
+   <td><dfn>IPv4-in-IPv6-invalid-code-point</dfn>
```
domenic, line 193: Really seems like this should be split into multiple errors, but I guess that could be refactored later.

domenic (approved):
LGTM with nits. Very clean.

url.bs
annevk: I think I ended up grouping some obscure ones into this one as there was no great logical separation. And IPv4-in-IPv6 is already pretty obscure so I'm not sure how detailed we really want to go.

annevk:
@TRowbotham did you want to give this a final look before it's merged?

TRowbotham:
LGTM. Thank you for pushing this over the finish line!

Issue #369, opened 2018-01-19
Need an "unreserved" character set (and better define how to percent-encode arbitrary strings)


mgiuca:
This is a bit of a jumble of related issues that all stem from one root problem: URL Standard (unlike [RFC 3986](https://www.ietf.org/rfc/rfc3986.txt)) does not have a concept of an "unreserved" character set. Apologies that this is a bit of an essay, but since these are all inter-related, I thought I would just group them into one discussion.

## Why an unreserved set?

To give some background, [RFC 3986](https://www.ietf.org/rfc/rfc3986.txt)'s **unreserved** set (ASCII alphanumeric plus `-._~`) is the set of characters that are interchangeable in their percent-encoded and non-encoded forms: "URIs that differ in the replacement of an unreserved character with its corresponding percent-encoded US-ASCII octet are equivalent: they identify the same resource." (The earlier [RFC 2396](https://www.ietf.org/rfc/rfc2396.txt) defined a slightly larger unreserved set: ASCII alphanumeric plus `!'()*-._~`, which will be relevant later.)

In other words, the RFC divides the set of valid URL characters into two subsets: reserved and unreserved. Percent-encoding or percent-decoding a reserved character *may* change the meaning of the URL (e.g., "?abc=def" and "?abc%3Ddef" have different meanings). Percent-encoding or percent-decoding an unreserved character *does not* change the meaning of the URL (e.g., "/abc/" and "/%61%62%63/" should be considered equivalent, with "/abc/" being the normalized representation).

URL Standard does not have an equivalent concept, and this manifests as several problems (each of which could have its own bug, but I think it helps to group these together):

1. **URL equivalence is broken.** RFC 3986 considers "/abc/" and "/%61%62%63/" to be equivalent; URL Standard does not. URL Standard treats "=" / "%3D" the same way as "a" / "%61" --- in both cases, it does not consider these equivalent. It needs to recognise the equivalence of "a" and "%61". Furthermore, it doesn't even recognise equivalence of uppercase and lowercase percent-encoded bytes: "%3D" and "%3d" are not considered equivalent, because the URL Parser does not normalize lowercase percent-encoded bytes to uppercase. (Note: I like that URL Parser serves as a "normalization" pass, with equivalence being trivially defined as compare serialization for equality. Therefore, I would like to address this in the Parser, rather than modifying the equivalence algorithm itself.)
[...]

PR #503, opened 2020-05-09
Add string UTF-8 percent encode

### LINKED ISSUE #296: "UTF-8 percent encode c using the path percent-e..."

annevk:
And use it internally. This is also an initial step for #369.

If reviewers have thoughts on how to resolve #296 that'd be appreciated. Renaming things is in scope for this PR modulo the caveats mentioned there. Resolving to continue using the existing names is also fine with me.

***
<a href="https://whatpr.org/url/503.html" title="Last updated on May 12, 2020, 7:23 AM UTC (d0fc83f)">Preview</a> | <a href="https://whatpr.org/url/503/ea3b75d...d0fc83f.html" title="Last updated on May 12, 2020, 7:23 AM UTC (d0fc83f)">Diff</a>

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -124,8 +124,8 @@ return failure.
 return a <a for=/>string</a> consisting of U+0025 (%), followed by two <a>ASCII upper hex digits</a>
 representing <var>byte</var>.
 
-<p>To <dfn export for="byte sequence" id=percent-decode>percent-decode</dfn> a <a>byte sequence</a>
-<var>input</var>, run these steps:
+<p>To <dfn export for="byte sequence" id=percent-decode>percent-decode</dfn> a
+<a for=/>byte sequence</a> <var>input</var>, run these steps:
 
 <p class=warning>Using anything but <a>UTF-8 decode without BOM</a> when <var>input</var> contains
 bytes that are not <a>ASCII bytes</a> might be insecure and is not recommended.
@@ -162,7 +162,8 @@ bytes that are not <a>ASCII bytes</a> might be insecure and is not recommended.
  <li><p>Return <var>output</var>.
 </ol>
 
-<p>To <dfn export for=string>percent-decode</dfn> a string <var>input</var>, run these steps:
+<p>To <dfn export for=string>percent-decode</dfn> a <a for=/>string</a> <var>input</var>, run these
+steps:
 
 <ol>
  <li><p>Let <var>bytes</var> be the <a>UTF-8 encoding</a> of <var>input</var>.
--- a/url.bs
+++ b/url.bs
@@ -183,7 +183,11 @@ U+0020 SPACE, U+0022 ("), U+003C (&lt;), U+003E (&gt;), and U+0060 (`).
 <a>path percent-encode set</a> and U+002F (/), U+003A (:), U+003B (;), U+003D (=), U+0040 (@),
 U+005B ([) to U+005E (^), inclusive, and U+007C (|).
 
-<p>To <dfn>UTF-8 percent encode</dfn> a <a for=/>code point</a> <var>codePoint</var>, using a
+<p class=note>The <a><code>application/x-www-form-urlencoded</code></a> format's
+<a lt="urlencoded byte serializer">byte serializer</a> and the <a>URL parser</a>'s
+<a>query state</a> use <a>percent encode</a> directly without any of these sets.
+
+<p>To <dfn>UTF-8 percent encode</dfn> a <a for=/>code point</a> <var>codePoint</var> using a
 <var>percentEncodeSet</var>, run these steps:
 
 <ol>
@@ -197,9 +201,17 @@ U+005B ([) to U+005E (^), inclusive, and U+007C (|).
  concatenated, in the same order.
 </ol>
 
-<p class="note no-backref">The <a><code>application/x-www-form-urlencoded</code></a> format's
-<a lt="urlencoded byte serializer">byte serializer</a> and the <a>URL parser</a>'s
-<a>query state</a> use <a>percent encode</a> directly.
+<p>To <dfn>string UTF-8 percent encode</dfn> a <a for=/>string</a> <var>input</var> using a
+<var>percentEncodeSet</var>, run these steps:
+
+<ol>
+ <li><p>Let <var>output</var> be the empty string.</p></li>
+
+ <li><p>For each <var>codePoint</var> of <var>input</var>, <a>UTF-8 percent encode</a>
+ <var>codePoint</var> using <var>percentEncodeSet</var> and append the result to <var>output</var>.
+
+ <li><p>Return <var>output</var>.
+</ol>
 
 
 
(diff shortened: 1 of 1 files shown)


annevk:
@rmisev so what do you think? Keep the existing names and close #296?

rmisev:
I am fine with existing names. Although "conditionally-UTF-8-percent encode" isn't bad too.

masinter:
I've been thinking about this a lot and I think part of the problem is the use of the word "string" to sometimes mean a sequence of bytes, a sequence of code points, a sequence of abstract characters.
The charset-named translations like utf-8 or us-ascii are mappings between sequence of abstract characters and sequence of bytes.

If you name the levels of encoding, some of the ambiguity drops away.

annevk:
If we use string (as defined by https://infra.spec.whatwg.org/#string) to mean bytes anywhere that's a bug. With regards to UTF-8 percent encode there's two confusing things I think:

* Not all input is necessarily percent encoded. This could be addressed by adding "conditional" if this is confusing enough.
* Percent encoding doesn't always mean bytes > strings. It just means that the input will be transformed to use the %.. encoding, but the inputs and outputs in terms of bytes/strings differ and are mostly made to suit the callers. I'm not sure there's clean way out of this other than avoiding the word "encoding" altogether.

url.bs
```diff
@@ -183,7 +183,11 @@ U+0020 SPACE, U+0022 ("), U+003C (&lt;), U+003E (&gt;), and U+0060 (`).
 <a>path percent-encode set</a> and U+002F (/), U+003A (:), U+003B (;), U+003D (=), U+0040 (@),
 U+005B ([) to U+005E (^), inclusive, and U+007C (|).
 
-<p>To <dfn>UTF-8 percent encode</dfn> a <a for=/>code point</a> <var>codePoint</var>, using a
+<p class=note>The <a><code>application/x-www-form-urlencoded</code></a> format's
+<a lt="urlencoded byte serializer">byte serializer</a> and the <a>URL parser</a>'s
+<a>query state</a> use <a>percent encode</a> directly without any of these sets.
+
+<p>To <dfn>UTF-8 percent encode</dfn> a <a for=/>code point</a> <var>codePoint</var> using a
```
saschanaz, line 190:
```suggestion
<p>To <dfn export>UTF-8 percent encode</dfn> a <a for=/>code point</a> <var>codePoint</var> using a
```

url.bs
```diff
@@ -197,9 +201,17 @@ U+005B ([) to U+005E (^), inclusive, and U+007C (|).
  concatenated, in the same order.
 </ol>
 
-<p class="note no-backref">The <a><code>application/x-www-form-urlencoded</code></a> format's
-<a lt="urlencoded byte serializer">byte serializer</a> and the <a>URL parser</a>'s
-<a>query state</a> use <a>percent encode</a> directly.
+<p>To <dfn>string UTF-8 percent encode</dfn> a <a for=/>string</a> <var>input</var> using a
```
saschanaz, line 204:
```suggestion
<p>To <dfn export>string UTF-8 percent encode</dfn> a <a for=/>string</a> <var>input</var> using a
```

url.bs
annevk: Which spec will use this after https://github.com/whatwg/html/pull/5524?

saschanaz: Ah, I mean, other spec (without s) 😅

AFAIK HTML doesn't depend on Shepherd but still `export` should indicate a term is potentially being used elsewhere, if anyone ever intend to remove it later.

saschanaz: Ah no, not this specific one. My confusion.

url.bs
domenic, line 204: IMO this should also be named "UTF-8 percent encode", distinguishing between the two with `for="code point"` vs. `for=string`. I.e., you should overload the algorithm on its input type.

This is especially intuitive here because the reader then blurs the difference between code points and single-character strings, which is a reasonable mental model that matches many programming languages.

domenic:
I think it would help to have an example table with clear inputs and outputs. Something like:

Here is a summary, by way of example, of these various operations:

|Operation|Example Input|Example Output|
|---------|-------------|--------------|
|percent encode|0x23|"`%23`"|
|percent decode|\``blah`\`|\``blah`\`|
|string percent decode|"`blah`"|\``blah`\`|
|UTF-8 percent encode with userinfo percent-encode set|U+LARGENUMBER|"`%12%34`"|
|UTF-8 percent encode with userinfo percent-encode set|"`blah`"|"`%blah%blah`"|

I think having these examples of inputs and outputs, which use the appropriate typography, will convey more than simply stating the return types.

url.bs
domenic: You could even rename "string percent decode" to just "percent decode" with for=string. And then you could add a note afterward saying something like

> In general, percent encoding results in a string, whereas percent decoding results in a byte sequence.

annevk:
That sounds pretty good. Note to self: Fetch and HTML have dependencies on "string percent decode".

url.bs
domenic: Maybe even

> In general, percent encoding results in a string with "more percent signs" than the input byte/code point/string, and percent decoding results in a byte sequence with "less percent signs" than the input string/byte sequence.

sideshowbarker:
(Warning: This long comment may seem like getting into “how many angels can fit on head of pin” territory but is offered FWIW from my (long-ago) training and experience as a technical writer.) 

As far as use of hyphens/dashes in the operation names, some options to consider:

1. (which I recommend) Always hyphenate *percent-encode* and *percent-decode*, and:
   * Don’t put a hyphen before *percent-encode* or *percent-decode* in any operation name that just has a single qualifier; so “*UTF-8 percent-encode*” and “*string percent-encode*”.
   * Do put a hyphen before *percent-encode* or *percent-decode* in any operation name that has more than one qualifier; so ”*conditionally UTF-8-percent-encode*”.
2. Hyphenate everything: “*UTF-8-percent-encode*” and “*conditionally-UTF-8-percent-encode*” and “*UTF-8-percent-encode-with-userinfo-percent-encode-set*”.
3. Hyphenate nothing (except *UTF-8* of course, and except when “*percent-encode*” or “*percent-decode*” are adjectives): “*UTF-8 percent encode*” and “*conditionally UTF-8 percent encode*” and “*UTF-8 percent encode with userinfo percent-encode set*.

Option 2 above is most-clearly not the best choice in this case, because if you apply it, you end up with that “*UTF-8-percent-encode-with-userinfo-percent-encode-set*” monstrosity.

Option 3 is also sub-optimal in this case, since in the same operation name, you end up with both “*percent encode*” (unhyphenated) and “*percent-encode*” (hyphenated). That risks making readers stumble — because at first reading, it might seem odd/inconsistent (even though it really isn’t).

Option 1 in this case therefore kind of hits the sweet spot (given all of the above).

---

**Further rationale**

While there aren’t set-in-stone rules for when to add hyphens — it’s always partly a judgement call — the main guiding principle (especially in technical docs) is to add hyphens to avoid ambiguity; that is, to avoid having readers mis-read or mis-parse or stumble over a particular phrase.

That’s especially important in cases where what you’re hyphenating is a compound adjective — because those are the cases where there’s the greatest risk of ambiguity and misreading. It’s less important in cases where what you’re hyphenating is a noun or verb — which is essentially the case here, because these are operation names.

But another principle is consistency: If you know you’re otherwise gonna end up with cases like “*UTF-8 percent encode with userinfo percent-encode set*” — one operation name that has the same combination of words both unhyphenated (“*percent encode*”) and hyphenated (“*percent-encode*”) — then avoid that by just always hyphenating it.

And the final principle is: When you’ve ended up with some term that is itself referencing/qualifying some other multi-word term, use hyphens to make that more clear; so that’s what leads you to *“conditionally UTF-8-percent-encode”* rather than just *“conditionally UTF-8 percent-encode”*.
</details>

PR #513, opened 2020-05-12
Add the component percent-encode set

### LINKED ISSUE #369: Need an "unreserved" character set (and better define how to percent-encode arbitrary strings)

annevk:
Complements https://github.com/whatwg/html/pull/5524. 

Closes #369.

***
<a href="https://whatpr.org/url/513.html" title="Last updated on Jun 26, 2020, 5:55 AM UTC (7c6cae3)">Preview</a> | <a href="https://whatpr.org/url/513/e5334dd...7c6cae3.html" title="Last updated on Jun 26, 2020, 5:55 AM UTC (7c6cae3)">Diff</a>

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -7,9 +7,8 @@ Abstract: The URL Standard defines URLs, domains, IP addresses, the <code>applic
 Translation: ja https://triple-underscore.github.io/URL-ja.html
 </pre>
 
-<!-- TODO: remove after Bikeshed database updates itself to no longer have url version -->
-<pre class=link-defaults>
-spec:html; type:dfn; text:same site
+<pre class=anchors>
+url: https://tc39.es/ecma262/#sec-encodeuricomponent-uricomponent; text: "encodeURIComponent() [sic]"; type: method; spec: ECMA-262
 </pre>
 
 
@@ -198,9 +197,19 @@ U+0027 (').
 <a>path percent-encode set</a> and U+002F (/), U+003A (:), U+003B (;), U+003D (=), U+0040 (@),
 U+005B ([) to U+005E (^), inclusive, and U+007C (|).
 
+<p>The <dfn export>component percent-encode set</dfn> is the <a>userinfo percent-encode set</a> and
+U+0024 ($) to U+0026 (&amp;), inclusive, U+002B (+), and U+002C (,).
+
+<p class=note>This is used by <cite>HTML</cite> for
+{{NavigatorContentUtils/registerProtocolHandler()}}, and could also be used by other standards to
+percent-encode data that can then be embedded in a <a for=/>URL</a>'s <a for=url>path</a>,
+<a for=url>query</a>, or <a for=url>fragment</a>. Using it with
+<a for=string>UTF-8 percent-encode</a> gives identical results to JavaScript's
+<a method><code>encodeURIComponent()</code> [sic]</a>. [[HTML]] [[ECMA-262]]
+
 <p>The <dfn><code>application/x-www-form-urlencoded</code> percent-encode set</dfn> is the
-<a>userinfo percent-encode set</a> and U+0021 (!), U+0024 ($) to U+0029 RIGHT PARENTHESIS,
-inclusive, U+002B (+), U+002C (,), and U+007E (~).
+<a>component percent-encode set</a> and U+0021 (!), U+0027 (') to U+0029 RIGHT PARENTHESIS,
+inclusive, and U+007E (~).
 
 <p class=note>The <a><code>application/x-www-form-urlencoded</code> percent-encode set</a> contains
 all code points, except the <a>ASCII alphanumeric</a>, U+002A (*), U+002D (-), U+002E (.), and
@@ -2571,8 +2580,8 @@ then runs these steps, returning an <a>ASCII string</a>:
 <h3 id=origin>Origin</h3>
 <!-- Still need to watch the final bits -->
 
-<p class=note>See <a for=/>origin</a>'s definition in HTML for the necessary
-background information. [[!HTML]]
+<p class=note>See <a for=/>origin</a>'s definition in <cite>HTML</cite> for the necessary background
+information. [[HTML]]
 
 <p>A <a for=/>URL</a>'s <dfn export for=url id=concept-url-origin>origin</dfn> is the
 <a for=/>origin</a> returned by running these steps, switching on
@@ -2805,8 +2814,7 @@ takes a list of name-value tuples <var>tuples</var>, optionally with an <a for=/
  <li>Return <var>output</var>.
 </ol>
 
-<p class="note no-backref">The <cite>HTML standard</cite> invokes this algorithm with values that
-are files. [[HTML]]
+<p class=note><cite>HTML</cite> invokes this algorithm with values that are files. [[HTML]]
 
 
 <h3 id=urlencoded-hooks>Hooks</h3>
@@ -3361,8 +3369,8 @@ such a feature "url" (i.e., lowercase and with an "l" at the end). Names such as
 "IRI" should not be used. However, if the name is a compound, "URL" (i.e., uppercase) is preferred,
 e.g., "newURL" and "oldURL".
 
-<p class=note>The {{EventSource}} and {{HashChangeEvent}} interfaces in HTML are examples of proper
-naming. [[!HTML]]
+<p class=note>The {{EventSource}} and {{HashChangeEvent}} interfaces in <cite>HTML</cite> are
+examples of proper naming. [[!HTML]]
 
 
 


url.bs
```diff
@@ -189,6 +188,16 @@ U+0020 SPACE, U+0022 ("), U+003C (&lt;), U+003E (&gt;), and U+0060 (`).
 <a>path percent-encode set</a> and U+002F (/), U+003A (:), U+003B (;), U+003D (=), U+0040 (@),
 U+005B ([) to U+005E (^), inclusive, and U+007C (|).
 
+<p>The <dfn export>component percent-encode set</dfn> is the <a>userinfo percent-encode set</a> and
+U+0024 ($) to U+0026 (&amp;), inclusive, U+002B (+), and U+002C (,).
+
+<p class=note>This is used by the <cite>HTML standard</cite> for
```
domenic, line 194:
```suggestion
<p class=note>This is used by the <cite>HTML Standard</cite> for
```

domenic: Or maybe "...used by <cite>HTML</cite> for..."

url.bs
```diff
@@ -189,6 +188,16 @@ U+0020 SPACE, U+0022 ("), U+003C (&lt;), U+003E (&gt;), and U+0060 (`).
...
+{{NavigatorContentUtils/registerProtocolHandler()}} and could also be used by other standards to
```
domenic, line 195:
```suggestion
{{NavigatorContentUtils/registerProtocolHandler()}}, and could also be used by other standards to
```

annevk:
I'll wait with this until we're all certain about https://github.com/whatwg/html/pull/5524.

annevk:
Now that bc7ac863aead17d6447133eb0de89a88bd6801f1 merged I should probably add a note here that the special-query percent-encode set's U+0027 (') is not covered by the component percent-encode set. Apart from that the component percent-encode set is a superset of all others I believe.

annevk:
Per https://whatpr.org/url/518.html#application-x-www-form-urlencoded-percent-encode-set this component percent-encode set would end up being a subset of the `application/x-www-form-urlencoded` percent-encode set. That seems fine.

Issue #411, opened 2018-08-08
Clarify query state percent-encoding (parsing)


mkgl:
Hi,

I've been scratching my head trying to figure out where do square brackets (e.g. U+005B) get exempted from [percent-encoding](https://url.spec.whatwg.org/#percent-encode), in the [query state](https://url.spec.whatwg.org/#query-state).

To compare with the [path state](https://url.spec.whatwg.org/#path-state):

> 3. [UTF-8 percent encode](https://url.spec.whatwg.org/#utf-8-percent-encode) [c](https://url.spec.whatwg.org/#c) using the [path percent-encode set](https://url.spec.whatwg.org/#path-percent-encode-set), and append the result to _buffer_.

I understand that—since U+005B ([) is not present in that set—it is returned _as is_ (as per [UTF-8 percent encode](https://url.spec.whatwg.org/#utf-8-percent-encode)). 

In the [query state](https://url.spec.whatwg.org/#query-state), encoding is expressed on a _byte_ basis:

> 1. If one of the following is true
>     * _byte_ is less than 0x21 (!)
>     * _byte_ is greater than 0x7E (~)
>     * _byte_ is 0x22 ("), 0x23 (#), 0x3C (<), or 0x3E (>)
>     * _byte_ is 0x27 (') and url is special
>
>     then append _byte_, [percent encoded](https://url.spec.whatwg.org/#percent-encode), to _url_’s [query](https://url.spec.whatwg.org/#concept-url-query).
> 
> 2. Otherwise, append a code point whose value is _byte_ to _url_’s [query](https://url.spec.whatwg.org/#concept-url-query).

Wouldn't it be equivalent to express a query percent-encode set as follows? That would make it much less convoluted imho.

> 💡 The **query percent-encode set** is the [C0 control percent-encode set](https://url.spec.whatwg.org/#c0-control-percent-encode-set) and U+0020 SPACE, U+0022 ("), U+0023 (#), U+003C (<), and U+003E (>).

Only the [special](https://url.spec.whatwg.org/#is-special) case about U+0027 (') would have to remain, not sure what to do about that.

Any thoughts?

Secondary question: in various states there is early validation:

> If [c](https://url.spec.whatwg.org/#c) is not a [URL code point](https://url.spec.whatwg.org/#url-code-points) and not U+0025 (%), [validation error](https://url.spec.whatwg.org/#validation-error).

[...]

PR #517, opened 2020-05-13
Revert "Editorial: use an encode set in the query state"

annevk:
This reverts commit bc7ac863aead17d6447133eb0de89a88bd6801f1.

When I started looking at what it would take to do this for application/x-www-form-urlencoded I realize that this is not editorial at all and in fact changes behavior.

If the input is U+0080 that would become 0xC2 0x80 when UTF-8 encoded. We then turn that into U+00C2 U+0080. We then UTF-8 percent-encode that, creating %C3%82%C2%80 rather than %C2%80.

I think the only way to go here is to have a byte version of UTF-8 percent-encoding. I guess that could still use the code point sets but it would be a bit messy/weird?

***
<a href="https://whatpr.org/url/517.html" title="Last updated on May 13, 2020, 4:00 PM UTC (e236f60)">Preview</a> | <a href="https://whatpr.org/url/517/bc7ac86...e236f60.html" title="Last updated on May 13, 2020, 4:00 PM UTC (e236f60)">Diff</a>

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -179,14 +179,8 @@ the input, and percent-decoding results in a byte sequence with less 0x25 (%) by
 <p>The <dfn oldids=simple-encode-set>C0 control percent-encode set</dfn> are the <a>C0 controls</a>
 and all <a>code points</a> greater than U+007E (~).
 
-<p>The <dfn>query percent-encode set</dfn> is the <a>C0 control percent-encode set</a> and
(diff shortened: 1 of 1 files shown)


annevk:
(Also need to reopen #411 if I'm correct.)

annevk:
Also, I should document this problem more clearly, at least in the source, as I vaguely recall now trying this at least once before and running into this.

PR #518, opened 2020-05-14
Editorial: make everything use percent-encode sets

### LINKED ISSUE #411: Clarify query state percent-encoding (parsing)

annevk:
This switches the URL parser's query state and the application/x-www-form-urlencoded's serializer to also use percent-encode sets.

Closes #411.

I'll see about writing a whatwg-url patch.

***
<a href="https://whatpr.org/url/518.html" title="Last updated on May 20, 2020, 5:18 AM UTC (91bead4)">Preview</a> | <a href="https://whatpr.org/url/518/19494f1...91bead4.html" title="Last updated on May 20, 2020, 5:18 AM UTC (91bead4)">Diff</a>

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -245,6 +245,7 @@ U+005F (_).
 
    <li><p>Otherwise, <a for=byte>percent-encode</a> <var>byte</var> and append the result to
    <var>output</var>.
+  </ol>
 
  <li><p>Return <var>output</var>.
 </ol>
--- a/url.bs
+++ b/url.bs
@@ -235,13 +235,13 @@ U+005F (_).
   <p>For each <var>byte</var> of <var>bytes</var>:
 
   <ol>
-   <li><p>Let <var>codePoint</var> be a <a for=/>code point</a> whose <a for="code point">value</a>
+   <li><p>Let <var>isomorph</var> be a <a for=/>code point</a> whose <a for="code point">value</a>
    is <var>byte</var>'s <a for=byte>value</a>.
 
    <li><p>Assert: <var>percentEncodeSet</var> includes all non-<a>ASCII code points</a>.
 
-   <li><p>If <var>codePoint</var> is not in <var>percentEncodeSet</var>, then append
-   <var>codePoint</var> to <var>output</var>.
+   <li><p>If <var>isomorph</var> is not in <var>percentEncodeSet</var>, then append
+   <var>isomorph</var> to <var>output</var>.
 
    <li><p>Otherwise, <a for=byte>percent-encode</a> <var>byte</var> and append the result to
    <var>output</var>.
--- a/url.bs
+++ b/url.bs
@@ -268,8 +268,8 @@ of running <a for="code point">percent-encode after encoding</a> with <a for=/>U
 
 <p>To <dfn export for=string>UTF-8 percent-encode</dfn> a <a for=/>string</a> <var>input</var> using
 a <var>percentEncodeSet</var>, return the result of running
-<a for=string>percent-encode after encoding</a> with <a for=/>UTF-8</a>, <var>input</var>, and
-<var>percentEncodeSet</var>.
+<a for=string>percent-encode after encoding</a> with <a for=/>UTF-8</a>, <var>input</var>,
+<var>percentEncodeSet</var>, and false.
 
 <hr>
 
--- a/url.bs
+++ b/url.bs
@@ -202,17 +202,15 @@ U+005B ([) to U+005E (^), inclusive, and U+007C (|).
 <a>userinfo percent-encode set</a> and U+0021 (!), U+0024 ($) to U+0029 RIGHT PARENTHESIS,
 inclusive, U+002B (+), U+002C (,), and U+007E (~).
 
-<p class=note>The <a>application/x-www-form-urlencoded percent-encode set</a> contains all code
-points, except the <a>ASCII alphanumeric</a>, U+002A (*), U+002D (-), U+002E (.), and U+005F (_).
+<p class=note>The <a><code>application/x-www-form-urlencoded</code> percent-encode set</a> contains
+all code points, except the <a>ASCII alphanumeric</a>, U+002A (*), U+002D (-), U+002E (.), and
+U+005F (_).
 
 <p>To <dfn for="code point">percent-encode after encoding</dfn>, given an <a for=/>encoding</a>
 <var>encoding</var>, <a for=/>code point</a> <var>codePoint</var>, and a
 <var>percentEncodeSet</var>, run these steps:
 
 <ol>
- <li><p>If <var>codePoint</var> is not in <var>percentEncodeSet</var>, then return
- <var>codePoint</var>.
-
  <li><p>Let <var>bytes</var> be the result of <a lt=encode>encoding</a> <var>codePoint</var> using
  <var>encoding</var>.
 
@@ -233,8 +231,20 @@ points, except the <a>ASCII alphanumeric</a>, U+002A (*), U+002D (-), U+002E (.)
 
  <li><p>Let <var>output</var> be the empty string.</p></li>
 
- <li><p>For each <var>byte</var> of <var>bytes</var>, <a for=byte>percent-encode</a>
- <var>byte</var> and append the result to <var>output</var>.
+ <li>
+  <p>For each <var>byte</var> of <var>bytes</var>:
+
+  <ol>
+   <li><p>Let <var>codePoint</var> be a <a for=/>code point</a> whose <a for="code point">value</a>
+   is <var>byte</var>'s <a for=byte>value</a>.
+
+   <li><p>Assert: <var>percentEncodeSet</var> includes all non-<a>ASCII code points</a>.
+
+   <li><p>If <var>codePoint</var> is not in <var>percentEncodeSet</var>, then append
+   <var>codePoint</var> to <var>output</var>.
+
+   <li><p>Otherwise, <a for=byte>percent-encode</a> <var>byte</var> and append the result to
+   <var>output</var>.
 
  <li><p>Return <var>output</var>.
 </ol>
@@ -3172,7 +3182,7 @@ console.log(url.searchParams.get('a')); // "~"
 console.log(url.searchParams.get('b')); // "~"</code></pre>
 
  <p>{{URLSearchParams}} objects will percent-encode anything in the
- <a><code>application/x-www-form-urlencoded</code> percent-encode set</a>. And will encode
+ <a><code>application/x-www-form-urlencoded</code> percent-encode set</a>, and will encode
  U+0020 SPACE as U+002B (+).
 
  <p>Ignoring encodings (use <a>UTF-8</a>), {{URL/search}} will percent-encode anything in the
--- a/url.bs
+++ b/url.bs
@@ -180,27 +180,56 @@ the input, and percent-decoding results in a byte sequence with less 0x25 (%) by
 and all <a>code points</a> greater than U+007E (~).
 
 <p>The <dfn>fragment percent-encode set</dfn> is the <a>C0 control percent-encode set</a> and
-U+0020 SPACE, U+0022 ("), U+003C (&lt;), U+003E (&gt;), and U+0060 (`).
+U+0020 SPACE, U+0022 ("), U+003C (&lt;), U+003E (>), and U+0060 (`).
+
+<p>The <dfn>query percent-encode set</dfn> is the <a>C0 control percent-encode set</a> and
+U+0020 SPACE, U+0022 ("), U+0023 (#), U+003C (&lt;), and U+003E (>).
+
+<p class=note>The <a>query percent-encode set</a> cannot be defined in terms of the
+<a>fragment percent-encode set</a> due to the omission of U+0060 (`).
+
+<p>The <dfn>special-query percent-encode set</dfn> is the <a>query percent-encode set</a> and
+U+0027 (').
 
 <p>The <dfn oldids=default-encode-set>path percent-encode set</dfn> is the
-<a>fragment percent-encode set</a> and U+0023 (#), U+003F (?), U+007B ({), and U+007D (}).
+<a>query percent-encode set</a> and U+003F (?), U+0060 (`), U+007B ({), and U+007D (}).
 
 <p>The <dfn oldids=userinfo-encode-set>userinfo percent-encode set</dfn> is the
 <a>path percent-encode set</a> and U+002F (/), U+003A (:), U+003B (;), U+003D (=), U+0040 (@),
 U+005B ([) to U+005E (^), inclusive, and U+007C (|).
 
-<p class=note>The <a><code>application/x-www-form-urlencoded</code></a> format's
-<a lt="urlencoded byte serializer">byte serializer</a> and the <a>URL parser</a>'s
-<a>query state</a> use <a for=byte>percent-encode</a> directly without any of these sets.
+<p>The <dfn><code>application/x-www-form-urlencoded</code> percent-encode set</dfn> is the
+<a>userinfo percent-encode set</a> and U+0021 (!), U+0024 ($) to U+0029 RIGHT PARENTHESIS,
+inclusive, U+002B (+), U+002C (,), and U+007E (~).
 
-<p>To <dfn for="code point" id=utf-8-percent-encode>UTF-8 percent-encode</dfn> a
-<a for=/>code point</a> <var>codePoint</var> using a <var>percentEncodeSet</var>, run these steps:
+<p class=note>The <a>application/x-www-form-urlencoded percent-encode set</a> contains all code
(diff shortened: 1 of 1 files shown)


url.bs
```diff
@@ -210,19 +239,38 @@ U+005B ([) to U+005E (^), inclusive, and U+007C (|).
...
 
+<p>To <dfn for="code point" id=utf-8-percent-encode>UTF-8 percent-encode</dfn> a
+<a for=/>code point</a> <var>codePoint</var> using a <var>percentEncodeSet</var>, return the result
+of running <a for="code point">percent-encode after encoding</a> with <a for=/>UTF-8</a>,
+<var>codePoint</var>, and <var>percentEncodeSet</var>.
+
+<p>To <dfn export for=string>UTF-8 percent-encode</dfn> a <a for=/>string</a> <var>input</var> using
+a <var>percentEncodeSet</var>, return the result of running
+<a for=string>percent-encode after encoding</a> with <a for=/>UTF-8</a>, <var>input</var>, and
+<var>percentEncodeSet</var>.
```
rmisev, line 272:
```suggestion
<a for=string>percent-encode after encoding</a> with <a for=/>UTF-8</a>, <var>input</var>,
<var>percentEncodeSet</var>, and false.
```

It seems you omitted `spaceAsPlus` value.

url.bs
```diff
@@ -3174,18 +3166,13 @@ console.log(url.search);                // "?a=~&b=%7E"
...
- <!-- From https://url.spec.whatwg.org/#concept-urlencoded-byte-serializer, inverted.
-      Do not change this without double checking URLENCODED-UNITS -->
-
- <p>Ignoring encodings (use <a>UTF-8</a>), {{URL/search}} will percent-encode U+0000 NULL to
- U+0020 SPACE, inclusive, U+0022 ("), U+0023 (#), U+0027 (') varying on <a>is special</a>,
- U+003C (&lt;), U+003E (>), and anything greater than U+007E (~).
- <!-- From https://url.spec.whatwg.org/#query-state.
-      Do not change this without double checking QUERY-UNITS -->
+ <p>{{URLSearchParams}} objects will percent-encode anything in the
+ <a><code>application/x-www-form-urlencoded</code> percent-encode set</a>. And will encode
```
domenic, line 3170:
```suggestion
 <a><code>application/x-www-form-urlencoded</code> percent-encode set</a>, and will encode
```

url.bs
```diff
@@ -180,27 +180,56 @@ the input, and percent-decoding results in a byte sequence with less 0x25 (%) by
...
-<p class=note>The <a><code>application/x-www-form-urlencoded</code></a> format's
-<a lt="urlencoded byte serializer">byte serializer</a> and the <a>URL parser</a>'s
-<a>query state</a> use <a for=byte>percent-encode</a> directly without any of these sets.
+<p>The <dfn><code>application/x-www-form-urlencoded</code> percent-encode set</dfn> is the
+<a>userinfo percent-encode set</a> and U+0021 (!), U+0024 ($) to U+0029 RIGHT PARENTHESIS,
+inclusive, U+002B (+), U+002C (,), and U+007E (~).
 
-<p>To <dfn for="code point" id=utf-8-percent-encode>UTF-8 percent-encode</dfn> a
-<a for=/>code point</a> <var>codePoint</var> using a <var>percentEncodeSet</var>, run these steps:
+<p class=note>The <a>application/x-www-form-urlencoded percent-encode set</a> contains all code
```
domenic, line 205:
```suggestion
<p class=note>The <a><code>application/x-www-form-urlencoded</code> percent-encode set</a> contains all code
```

rmisev:
It seems query state with [`ISO-2022-JP` encoding](https://encoding.spec.whatwg.org/#iso-2022-jp-encoder) will not work well after this update. For example `ISO-2022-JP` encodes U+00A5 to bytes: 0x1B 0x28 0x4A 0x5C 0x1B 0x28 0x42.

Before this update, result of encoding U+00A5 in query is: `%1B(J\%1B(B`
After this update - result is: `%1B%28%4A%5C%1B%28%42`

mgiuca:
> It seems query state with [`ISO-2022-JP` encoding](https://encoding.spec.whatwg.org/#iso-2022-jp-encoder) will not work well after this update. For example `ISO-2022-JP` encodes U+00A5 to bytes: 0x1B 0x28 0x4A 0x5C 0x1B 0x28 0x42.
> 
> Before this update, result of encoding U+00A5 in query is: `%1B(J\%1B(B`
> After this update - result is: `%1B%28%4A%5C%1B%28%42`

True, I agree. That's because the old query state logic performed encoding, then checked whether the byte was in the query percent encode set in deciding whether to percent-encode it. The new "percent-encode after encoding" logic checks whether the code point is in the percent encode set, then encodes it, then percent-encodes those bytes unconditionally.

Fixing this is a bit messy. I think we need to shuffle around the "percent-encode after encoding" algorithm:

* Delete 1. "If _codePoint_ is not in _percentEncodeSet_, then return _codePoint_."
* Replace 5. "For each _byte_ of _bytes_, percent-encode _byte_ and append the result to _output_." with "For each _byte_ of _bytes_, if _byte_ is not an [ASCII byte](https://infra.spec.whatwg.org/#ascii-byte), or if the code point whose value is _byte_ is not in _percentEncodeSet_, percent-encode _byte_ and append the result to _output_. Otherwise, append the code point whose value is _byte_ to _output_.

This involved a fair amount of byte/codepoint gymnastics, which is unavoidable if we want to preserve the existing behaviour (which literally runs an encoder, then represents a selection of bytes using their ASCII equivalents). I'm using the "code point whose value is _byte_" language, borrowed from the old query state algorithm.

This change should only affect non-UTF-8 encodings. When using UTF-8, all ASCII code points encode to the corresponding ASCII byte (so either way, characters in the encode set will be encoded or not according to the percent-encode set), and all non-ASCII code points encode to a sequence of non-ASCII bytes (so either way, all of those bytes will be percent-encoded).

You might note that in fact this is the only usage of percent-encode sets, which are sets of code points. It would make the above algorithm easier if we converted all the percent-encode sets into sets of bytes (e.g., "U+0023 (#), U+003F (?)" becomes "0x23 (#), 0x3F (?)"), and then "the code point whose value is _byte_ is not in _percentEncodeSet_" becomes "_byte_ is not in _percentEncodeSet_". But I would rather not do this, since fundamentally the percent encode sets should represent characters, not bytes.

url.bs
```diff
@@ -180,27 +180,56 @@ the input, and percent-decoding results in a byte sequence with less 0x25 (%) by
...
+points, except the <a>ASCII alphanumeric</a>, U+002A (*), U+002D (-), U+002E (.), and U+005F (_).
+
+<p>To <dfn for="code point">percent-encode after encoding</dfn>, given an <a for=/>encoding</a>
+<var>encoding</var>, <a for=/>code point</a> <var>codePoint</var>, and a
+<var>percentEncodeSet</var>, run these steps:
 
 <ol>
  <li><p>If <var>codePoint</var> is not in <var>percentEncodeSet</var>, then return
```
mgiuca, line 213: Delete this.

url.bs
```diff
@@ -180,27 +180,56 @@ the input, and percent-decoding results in a byte sequence with less 0x25 (%) by
...
+   <li><p>Replace the first two code points of <var>output</var> with "<code>%26%23</code>".
+
+   <li><p>Replace the last code point of <var>output</var> with "<code>%3B</code>".
+
+   <li><p>Return <var>output</var>.
+  </ol>
+
+  <p class="note no-backref">This can happen when <var>encoding</var> is not <a>UTF-8</a>.
 
  <li><p>Let <var>output</var> be the empty string.</p></li>
```
mgiuca, line 235: Replace this with:

```html
<li><p>For each <var>byte</var> of <var>bytes, if <var>byte</var> is not an <a>ASCII byte</a>,
or if the code point whose value is <var>byte</var> is not in <var>percentEncodeSet</var>,
<a for=byte>percent-encode</a> <var>byte</var> and append the result to <var>output</var>.
Otherwise, append the code point whose value is <var>byte</var> to <var>output</var>.
```

(Possibly splitting into some nested list items.)

annevk: The additional ASCII byte check doesn't seem to be needed?

mgiuca: True, anything that's "not an ASCII byte" is always going to be in every percent-encode set (because the C0 control set includes "all code points greater than U+007E (~).").

It's a little dubious to rely on this, however, especially since if we removed that check, we'd be doing these weird comparisons between non-ASCII bytes and non-ASCII code points. For example, if UTF-8 encoder gives us a byte 0xD2, we would be consulting the percent-encode set for the character U+00D2, which will "work", but it has nothing to do with that particular code point. So I put the ASCII byte check, so that then we straight-up guarantee that any non-ASCII byte will be encoded without being converted to an unrelated code point.

It might also be worth adding a note next to the percent-encode sets to say that the encoding algorithm assumes (either way we decide to do this) that all non-ASCII bytes are in every percent-encode set.

annevk: Yeah, created https://github.com/whatwg/infra/pull/305 to help with this.

annevk: But also, I think you got the steps the wrong way around, if it is in the set, it should be encoded, if it's not in the set, it should not be.

url.bs
```diff
@@ -180,49 +180,108 @@ the input, and percent-decoding results in a byte sequence with less 0x25 (%) by
...
 
- <li><p>For each <var>byte</var> of <var>bytes</var>, <a for=byte>percent-encode</a>
- <var>byte</var> and append the result to <var>output</var>.
+ <li>
+  <p>For each <var>byte</var> of <var>bytes</var>:
+
+  <ol>
+   <li><p>Let <var>codePoint</var> be a <a for=/>code point</a> whose <a for="code point">value</a>
```
domenic, line 238: This redeclares the codePoint variable, which is kind of bad.

domenic:
I've implemented this in whatwg-url in https://github.com/jsdom/whatwg-url/pull/158, so am happy to give my usual sign-off. However, that doesn't test non-UTF-8 encodings, so manual review would be needed for those code branches.

annevk:
@domenic https://fetch.spec.whatwg.org/#concept-body-package-data calls it. I think the main reason I did it this way is that it seems natural to me that for a MIME-type-based format the parser starts from bytes. Though to be fair I guess that should also mean that the serializer goes to bytes which isn't the case (though happens to match the HTML parser).

annevk:
@rmisev @mgiuca any final thoughts?

rmisev:
Isn't it better to define percent encode sets in terms of bytes (not code points)? This will simplify percent encoding algorithm a bit.

annevk:
The percent-encode algorithms take code points or strings. It seems a bit weird to contrast that with sets of bytes. But it's also not clear what would be simplified. We'd still need to turn the byte into a code point. I guess we could maybe drop the assert?

rmisev:
Yes, drop assert. But maybe it is not worth changing.

masinter:
percent encoding takes a sequence of bytes and produces a sequence of (ASCII) characters
A charset like UTF-8 or ISO8859-1 or shift-jis maps a sequence of characters to a sequence of bytes.
no codepoints for shift-jis, for example.

masinter:
Keep in mind that percent encoding is used outside of URL processing as a kind of content-encoding.

url.bs
```diff
@@ -180,49 +180,108 @@ the input, and percent-decoding results in a byte sequence with less 0x25 (%) by
...
- <li><p>For each <var>byte</var> of <var>bytes</var>, <a for=byte>percent-encode</a>
- <var>byte</var> and append the result to <var>output</var>.
+ <li>
+  <p>For each <var>byte</var> of <var>bytes</var>:
+
+  <ol>
+   <li><p>Let <var>isomorph</var> be a <a for=/>code point</a> whose <a for="code point">value</a>
+   is <var>byte</var>'s <a for=byte>value</a>.
+
+   <li><p>Assert: <var>percentEncodeSet</var> includes all non-<a>ASCII code points</a>.
```
mgiuca, line 241: Nit: Link to [Assert](https://infra.spec.whatwg.org/#assert)

url.bs
```diff
@@ -246,9 +305,28 @@ a <var>percentEncodeSet</var>, run these steps:
...
+  <tr>
+   <td>U+203D (‽)
+   <td>"<code>%26%238253%3B</code>"
+  <tr>
+   <td><a for=string>Percent-encode after encoding</a> with <a>Shift_JIS</a>, <var>input</var>, the
+   <a>userinfo percent-encode set</a>, and true
+   <td>"<code>1+1 ≡ 2%20‽</code>"
+   <td>"<code>1+1+%81%DF+2%20%26%238253%3B</code>"
+  <tr>
+   <td rowspan=2><a for="code point">UTF-8 percent-encode</a> <var>input</var> using the
```
mgiuca, line 324: Nit: Perhaps move the UTF-8 examples up above Shift-JIS since they are the far more common example.

url.bs
```diff
@@ -246,9 +305,28 @@ a <var>percentEncodeSet</var>, run these steps:
    <td>"<code>‽%25%2E</code>"
    <td>0xE2 0x80 0xBD 0x25 0x2E
   <tr>
-   <td><a for="code point">UTF-8 percent-encode</a> <var>input</var> using the
+   <td rowspan=3><a for="code point">Percent-encode after encoding</a> with <a>Shift_JIS</a>,
```
mgiuca, line 308: I'd like to see @rmisev 's example added here, since it's a key case (i.e., all of these examples present would work in your old algorithm and your new one; your fixed algorithm differs only by that special case where some of the bytes that encode the code point are < 128, which seems to happen in ISO-2022-JP but not UTF-8 or Shift-JIS).

So, the example is:
- Operation: Percent-encode after encoding with ISO-2022-JP, _input_, and the userinfo percent-encode set
- Input: U+00A5 (¥)
- Output: "`%1B(J\%1B(B`"

mgiuca:
@masinter :
> percent encoding takes a sequence of bytes and produces a sequence of (ASCII) characters

This spec defines both bytes-to-ASCII and codepoints-to-ASCII percent encoding functions. Both are needed by the URL parser (the latter calls the former).

> Keep in mind that percent encoding is used outside of URL processing as a kind of content-encoding.

I don't believe these terms are exported. The way it is defined here is specifically for use inside the URL parser / encoder. (For example, JavaScript's `encodeURIComponent` function defines its own encoding that _we believe_ is compatible with this one, but that doesn't call into here.)

Either way, I believe the set of functions defined here are adequate to serve any external use. If you want to encode a byte sequence (rather than a CP sequence), use the "percent-decode a byte sequence" algorithm above.

PR #527, opened 2020-06-13
Meta: update repository files

foolip:
See https://github.com/whatwg/spec-factory for details.


foolip:
Success: https://url.spec.whatwg.org/

Issue #302, opened 2017-04-28
Why is the host dropped if the path contains a Windows drive letter?


felixfbecker:
The `file:` URL scheme does not define that the URL cannot have a host:

> A file URI takes the form of
>     file://host/path
> where host is the fully qualified domain name of the system on which the path is accessible, and path is a hierarchical directory path of the form directory/directory/.../name. If host is omitted, it is taken to be "localhost", the machine from which the URL is being interpreted.

https://en.wikipedia.org/wiki/File_URI_scheme

It looks like the spec says in https://url.spec.whatwg.org/#path-state 1.4.1.1 that a host is to be dropped if the path contains a Windows drive letter:

> - If url’s scheme is "file", url’s path is empty, and buffer is a Windows drive letter, then:
>   - If url’s host is neither the empty string nor null, validation error, set url’s host to the empty string.

This means it's impossible to have a Windows file URL have a different host than the empty string:

```js
const uri = new URL('file://host/foo')
uri.host // 'host'
uri.pathname = '/foo'
uri.host // ''

new URL('/c:/baz/qux', 'file://host/foo/bar').href
"file:///c:/baz/qux"
```

This can be seen in [jsdom/whatwg-url](https://github.com/jsdom/whatwg-url). Chrome actually behaves differently and keeps the host. NodeJS 7's bundled URL implementation also keeps the host.

but it _is_ possible with a Unix-style file URL. Is there any reason for this?
It should be possible to refer to a file on a remote Windows machine just like on a remote Unix machine.

annevk:
What does Edge do?

And just to be clear, you mean that on Windows `file://somehost/C:/test/` can be a legit location?

felixfbecker:
Edge keeps the host, but throws an Invalid URL Error upon serializing with `href` or `toString()`:

![image](https://cloud.githubusercontent.com/assets/10532611/25525434/37a1b502-2c0f-11e7-9a14-7dc5a0771ed0.png)

So it doesn't really follow the spec either, but if they changed it to support this it wouldn't be a breaking change.

There is no indication why it would _not_ be a valid location, and why it would be valid for remote Unix machines but not for remote Windows machines.

[...]

PR #544, opened 2020-09-25
Changes to file URL path normalization

### LINKED ISSUE #405: Review slashes after file:/// and file path normalization
The current standard asks to remove leading empty directories from file paths. 
This seems to have been motivated by the desire to ignore slashes after `file:///` whilst avoiding reparsing issues. However,

- It introduces new behavior that has no precedence in browsers. 
- It is inconsistent with path normalization of other special URLs, and less expressive.  

Examples, tested in Chrome canary, Safari preview and Firefox nightly:

|   | Input                                        | Currently specified          | IE8 upto Edge18               | Firefox and Safari        | Chrome                        |
| - | -------------------------------------------- | ---------------------------- | ----------------------------- | ------------------------- | ----------------------------- |
| 1 | `/..//localhost//pig` against `file://lion/` | `file://lion/localhost//pig` | `file://lion//localhost//pig` | `file:////localhost//pig` | `file://lion//localhost//pig` |
| 2 | `file://localhost//a//../..//`               | `file:///`                   | `file:///.//`                 | `file://///`              | `file://localhost///`         |
| 3 | `file://localhost//a//../..//foo`            | `file:///foo`                | `file:///.//foo`              | `file://///foo`           | `file://localhost///foo`      |
| 4 | `file://localhost////foo`                    | `file:///foo`                | `file://foo/`                 | `file://////foo`          | `file://localhost////foo`     |
| 5 | `file:////foo`                               | `file:///foo`                | `file://foo/`                 | `file:////foo`            | `file:///foo`                 |

Notes:
- The spec agrees only with Chrome, only in the fourth case. 
- Browsers agree on the pathName except for the fourth case. 
(Dropping the host in the first seems off). 
- Firefox and Safari do not ignore slashes and thereby avoid reparsing issues. 
- Chrome only ignores slashes immediately following `file:///` and keeps `localhost`, avoiding a reparsing issue.

alwinb:
Changes to file URL path normalization as proposed in #405.  

Summary:

Applicable only to file URLs:
- Leading empty path segements are no longer removed. 
- If absent, the host is copied from the base URL,
- Otherwise the host is preserved with one exception:
- If the host is `localhost` then it is set to the empty string. 

Intends to close #405 and #302. 

- Tests: https://github.com/web-platform-tests/wpt/pull/25716. 
- Reference implementation: https://github.com/jsdom/whatwg-url/pull/170.

***
<a href="https://whatpr.org/url/544.html" title="Last updated on Sep 29, 2020, 9:54 AM UTC (3469a42)">Preview</a> | <a href="https://whatpr.org/url/544/83adf0c...3469a42.html" title="Last updated on Sep 29, 2020, 9:54 AM UTC (3469a42)">Diff</a>

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -2209,6 +2209,8 @@ string <var>input</var>, optionally with a <a>base URL</a> <var>base</var>, opti
         "<code>file</code>", then:
 
         <ol>
+         <li><p>Set <var>url</var>'s <a for=url>host</a> to <var>base</var>'s <a for=url>host</a>.
+
          <li>
           <p>If the substring from <var>pointer</var> in <var>input</var> does not
           <a>start with a Windows drive letter</a> and <var>base</var>'s <a for=url>path</a>[0] is a
@@ -2216,9 +2218,7 @@ string <var>input</var>, optionally with a <a>base URL</a> <var>base</var>, opti
           <a for=url>path</a>[0] to <var>url</var>'s <a for=url>path</a>.
 
           <p class=note>This is a (platform-independent) Windows drive letter quirk.
-          <var>url</var>'s <a for=url>host</a> is null under these conditions.
 
-         <li><p>Set <var>url</var>'s <a for=url>host</a> to <var>base</var>'s <a for=url>host</a>.
         </ol>
 
        <li><p>Set <var>state</var> to <a>path state</a>, and decrease <var>pointer</var> by 1.
--- a/url.bs
+++ b/url.bs
@@ -2353,7 +2353,9 @@ string <var>input</var>, optionally with a <a>base URL</a> <var>base</var>, opti
          <li>
           <p>If <var>url</var>'s <a for=url>scheme</a> is "<code>file</code>", <var>url</var>'s
           <a for=url>path</a> <a for=list>is empty</a>, and <var>buffer</var> is a
-          <a>Windows drive letter</a>, then replace the second code point in <var>buffer</var> with U+003A (:).
+          <a>Windows drive letter</a>, then replace the second code point in <var>buffer</var> with
+          U+003A (:).
+
           <p class=note>This is a (platform-independent) Windows drive letter quirk.
 
          <li><p><a for=list>Append</a> <var>buffer</var> to <var>url</var>'s <a for=url>path</a>.
--- a/url.bs
+++ b/url.bs
@@ -2175,8 +2175,7 @@ string <var>input</var>, optionally with a <a>base URL</a> <var>base</var>, opti
           <ol>
            <li><p><a>Validation error</a>.
 
-           <li><p>Set <var>url</var>'s <a for=url>host</a> to null and <var>url</var>'s
-           <a for=url>path</a> to an empty list.
+           <li><p>Set <var>url</var>'s <a for=url>path</a> to an empty list.
           </ol>
 
           <p class=note>This is a (platform-independent) Windows drive letter quirk.
@@ -2206,22 +2205,20 @@ string <var>input</var>, optionally with a <a>base URL</a> <var>base</var>, opti
 
       <ol>
        <li>
-        <p>If <var>base</var> is non-null, <var>base</var>'s <a for=url>scheme</a> is
-        "<code>file</code>", and the substring from <var>pointer</var> in <var>input</var> does not
-        <a>start with a Windows drive letter</a>, then:
+        <p>If <var>base</var> is non-null and <var>base</var>'s <a for=url>scheme</a> is
+        "<code>file</code>", then:
 
         <ol>
          <li>
-          <p>If <var>base</var>'s <a for=url>path</a>[0] is a
+          <p>If the substring from <var>pointer</var> in <var>input</var> does not
+          <a>start with a Windows drive letter</a> and <var>base</var>'s <a for=url>path</a>[0] is a
           <a>normalized Windows drive letter</a>, then <a for=list>append</a> <var>base</var>'s
           <a for=url>path</a>[0] to <var>url</var>'s <a for=url>path</a>.
 
-          <p class=note>This is a (platform-independent) Windows drive letter quirk. Both
-          <var>url</var>'s and <var>base</var>'s <a for=url>host</a> are null under these conditions
-          and therefore not copied.
+          <p class=note>This is a (platform-independent) Windows drive letter quirk.
+          <var>url</var>'s <a for=url>host</a> is null under these conditions.
 
-         <li><p>Otherwise, set <var>url</var>'s <a for=url>host</a> to <var>base</var>'s
-         <a for=url>host</a>.
+         <li><p>Set <var>url</var>'s <a for=url>host</a> to <var>base</var>'s <a for=url>host</a>.
         </ol>
 
        <li><p>Set <var>state</var> to <a>path state</a>, and decrease <var>pointer</var> by 1.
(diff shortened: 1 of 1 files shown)


alwinb:
The build fails because bikeshed is down at the moment.

domenic (approved):

annevk:
Thanks a lot @alwinb for your work on this!

One thing I failed to check in on is bugs. I'll file one for Node.js and rust-url. I don't think Chrome/Firefox is needed as they have bigger issues. Should there be one for Safari?

alwinb:
Yes, there are corner cases where Safari is different. My understanding is this happens only when a host-less url is parsed against a base url. Quote:

> As an aside, Safari fails the mentioned tests because it (sometimes? or always?) does not copy the host from the base URL when parsing host-less URLs. I suspect that this is not intentional; 

From https://github.com/whatwg/url/issues/405#issuecomment-698208289. Note that this is not drive letter specific behaviour. 

And, thanks, it's the first time I did something like this so it's nice to see it work out!
Should I file a bug for Safari? I've not done that before either.

annevk:
Thanks, I filed https://bugs.webkit.org/show_bug.cgi?id=217170. (You're free to do this yourself next time if you want, https://github.com/whatwg/meta/blob/master/MAINTAINERS.md has some instructions.)

guybedford:
There is a concern here for Node.js modules that `import '//path'` and `import '///path'` will effectively now reference separate module URL specifiers despite referring to the same file path which might lead to unexpected identity issues. In turn this might affect scope lookups for import maps as well.

There are also xisting examples of `new URL('file:///' + path.resolve(path))` as well to get cross-platform path resolution which will now similarly return a non unique URL.

Are these properties hard constraints on the change here? The alternative in the Node.js modules system might be to throw for these `file:///////path` return cases in the resolver and hope that isn't too much of a breaking change. We are discussing it at the meeting on Wednesday so I can report back further here, but any thoughts on the above would help to hear before then too.

domenic:
In general the mapping from URLs to files on the filesystem will not be 1:1. E.g. on Windows `/path` and `/PATH` are the same file. This just adds one more case of this type.

This generally comes down to the question as to how module maps are to be keyed: by "request URL" (~ module specifier) or "response URL" (~file inode).

achristensen07:
It is true that how a filesystem handles a file URL is outside of the scope of the URL specification.  We support both case-sensitive and non-case-sensitive filesystems, which causes strange problems like domenic said.
While I expressed support in changing the spec to match Safari and Firefox, I don't have a strong preference here.

alwinb:
For NodeJS, another thing to maybe take into account is code portability. I think that e.g. `import '//path'` was non-portable in the old situation, where in a http context it would resolve to `http://path/`.

guybedford:
Various JS platforms and tools rely on file URLs now as their module registry identifier to align with browsers and support HTTPs loading alongside files. Node.js and Deno are two. More build tools may follow suit here as well.

Yes this is very much about edge cases. A better example might be something like:

```js
export async function loadFileSystemPath (path) {
  return await import('file:///' + path);
}
```

or any other registry full URL construction.

Case sensitivity is a problem Node.js had before the shift to URLs so is probably best treated out of scope certainly.

Might it be possible to more easily support these base separator renormalization cases here?

It not, Node.js will likely implement some extra renormalization / validations before landing these changes. But it's a class of bugs for server JS platforms going forward so if there's ways to make it easier that could be useful. Otherwise these separator normalization issues could cause these edge case bugs in environments that are not aware of the pitfalls.

guybedford:
Also note these cases may not even be that rare. Here is the list of tests that needed to be changed in Node.js core for this PR to land - https://github.com/nodejs/node/issues/35429#issuecomment-703274005.

domenic:
I think handling this in a normalization step is the right thing to do. The URL parser can produce many URLs that map to the same file; it's the function of systems built on top of the URL parser (such as HTTP servers, module registries, or standalone helpers like `function urlToFilePath`) to handle that.

alwinb:
I am not sure if the example is to be taken literally, but just in case, `new URL('file://' + path.resolve(filepath))` would solve some issues (assuming resolve returns absolute paths and collapses slashes). Better yet, you can use the pathname setter to be safe about `#` and `?` characters in directory/file names, but you do still have to manually escape percent signs, and backslashes in names within posix style paths. Ideally there would be a path.toURL library method for such conversions.

bmeck:
@alwinb [url.pathToFileURL](https://nodejs.org/dist/latest-v14.x/docs/api/url.html#url_url_pathtofileurl_path) exists already. I'm not entirely sure the recommendation here as the test for policies at least is failing with the existing implementation of that. We could update that helper but is the expectation that all comparisons of these leading `/` prefixes be coalesced when comparing? If so, does that mean URL comparison should not be done using `===` of the string representation / components?

guybedford:
@alwinb the specific issue is that the path being concatenated will be either `C:\\some\\path` or `/some/path`. Certainly `pathToFileURL` is recommended for encoding handling etc, but people still do the above.

domenic:
URL comparison can be done using ===. `file://foo` and `file://///foo` are different URLs.

File comparison cannot be done using === comparison on URLs. `file://foo` and `file:////foo` point to the same file, despite being different URLs.

Whether you want to use URL comparison, or file comparison, for your module specifiers, is up to you.

bmeck:
@domenic so, to clarify, if `file://foo` and `file:///foo` are expected to point to the same file does that mean that scoping mechanisms *should* treat them as 2 different URL spaces for things like import maps?

domenic:
Import maps (and the web in general) do not have a concept of files, only URLs. I don't know how Node, or "scoping mechanisms", want to deal with files.

guybedford:
@domenic in both Node.js and Deno, `import.meta.url` is a `file:///` URL from the module registry. In addition resolution is URL resolution when importing.

bmeck:
@domenic https://fetch.spec.whatwg.org/ is vague on the behavior of "file" URLs and I'm more just trying to ensure we don't have some problem with forwards compatibility compared to current behavior for what URLs are assigned when you use files (such as <input type="file">). It seems like prohibiting fetch operations from succeeding for more than 2 `/` would be forwards compatible for now and possible to relax in the future? I guess a different question is, would the web ever have a place where it expects to generate a URL with more than 2 leading `/` for a file. More asking to avoid divergence as much as possible rather than having opinions.

domenic:
Right, the translation from URLs to files on the filesystem is not specified (certainly not by Fetch), and not really part of the web. It's browser- and OS-specific. I'm not sure how to give compatibility guidance about that subject.

> I guess a different question is, would the web ever have a place where it expects to generate a URL with more than 2 leading / for a file.

Yes, for example if someone did `<a href="file://////foo">`, or `new URL("file://///foo")`, those would all generate such URLs.

masinter:
don't forget there are normalization issues with Unicode non-ascii file names.
Case insensitive file name matching is only a small part of the problem.

ExE-Boss:
> In general the mapping from URLs to files on the filesystem will not be 1:1. E.g. on Windows `/path` and `/PATH` are the same file. This just adds one more case of this type.

Not if you have the case sensitive flag enabled on the directory in which the file is located:
- <https://devblogs.microsoft.com/commandline/per-directory-case-sensitivity-and-wsl/>
- <https://devblogs.microsoft.com/commandline/improved-per-directory-case-sensitivity-support-in-wsl/>

Issue #549, opened 2020-10-05
File URL reparse issue


rmisev:
URLs to test:
`file:.//p`
`file:/.//p`

I pass: `file:///.//p`
II pass: `file:////p`

It seems there is a bug in the URL serializer. When parsing first time, host is set to null, so URL serializer serializes path as it does with not special URL -  in the 5.1. step it adds `/.` to the output. When parsing second time, host is empty, 5.1. conditions are not meet, so `/.` is not added, and now result is as expected.

Issue #548, opened 2020-10-03
Allow ports in file URLs


achristensen07:
Currently, with the test URL file://example:1/ we have three different behaviors:
1. Firefox will remove the host and return file:///
2. Chrome does not allow the port and throws an exception.  This is currently the behavior described in the specification.
3. Safari allows the port.

While attempting to disallow ports from file URLs in https://bugs.webkit.org/show_bug.cgi?id=217252 I found a test written 13 years ago in https://trac.webkit.org/changeset/24484/webkit that indicates there is software that relies on the port not being forbidden in file URLs, and that there has been such software for a very long time.

I'm wondering if there is a good reason to not allow a port in file URLs?  It is true that most file URLs won't have it, and I'm not sure what anything would do with the port, but I don't see a good reason to forbid it and fail parsing.

Would we be willing to add a transition from file host state to port state if c is ':'?

annevk:
https://github.com/whatwg/url/issues/97#issuecomment-270461963 has some discussion on this. There's a concern around Windows drive letters there and your suggested patch would end up doing the wrong thing for those, so a solution here would likely be more involved. If we allowed this it would require some big changes to the API side as well (potentially simplifications though).

cc @alwinb @valenting @sleevi @jasnell @domenic

alwinb:
I have not looked into the port issue, but FWIW I'm pretty sure you can do away with the file host state in the standard fairly easily. Just parse them as special URLs and enforce that the authority does not have credentials later. Then you can handle whatever is decided about the port there too. You need to go back to using nulls for empty user/pass though, which I think I saw somewhere has changed before. 

Right now the host parser (which is called from the file host state) raises an error on `:` and `@`, so it just changes where the error is raised.

achristensen07:
I'm also curious why we prevent a username and password in a file URL when https://tools.ietf.org/html/rfc8089#appendix-E.1 clearly includes them.  I can easily imagine a use case when the host requires credentials, but that would add another place where a colon could be in a file URL, although not in the path.  Allowing a port would also allow a colon, but not in the path.  The comment in issue #97 says "that would require a lot more monkey-ing about with the UNC & drive-letter sniffing logic."  What would it take to do that "monkey-ing about" and make it only look in the path for drive letters?
FWIW, as of this morning this is the last url-constructor.html web platform test that is failing in WebKit.

alwinb:
[...]

PR #550, opened 2020-10-05
Fix file URL reparse issue

rmisev:
Fixes issue by ensuring file URL's host is not null.

Changes:
- Set url's host to the empty string at the start of file state.
- Updates URL parser and serializer accordingly.

Fixes: https://github.com/whatwg/url/issues/549

   * …
   * …
   * https://github.com/web-platform-tests/wpt/pull/25989
   * Chrome: no bugs
   * Firefox: no bugs
   * Safari: no bugs

***
<a href="https://whatpr.org/url/550.html" title="Last updated on Oct 7, 2020, 9:56 AM UTC (bcad0c4)">Preview</a> | <a href="https://whatpr.org/url/550/47efa00...bcad0c4.html" title="Last updated on Oct 7, 2020, 9:56 AM UTC (bcad0c4)">Diff</a>

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -2536,9 +2536,10 @@ then runs these steps, returning an <a>ASCII string</a>:
   <p>Otherwise:
 
   <ol>
-   <li><p>If <var>url</var>'s <a for=url>host</a> is null, <var>url</var>'s <a for=url>path</a>'s
-   <a for=list>size</a> is greater than 1, and <var>url</var>'s <a for=url>path</a>[0] is the empty
-   string, then append U+002F (/) followed by U+002E (.) to <var>output</var>.
+   <li><p>If <var>url</var>'s <a for=url>host</a> is null, <var>url</var> <a>is not special</a>,
+   <var>url</var>'s <a for=url>path</a>'s <a for=list>size</a> is greater than 1, and
+   <var>url</var>'s <a for=url>path</a>[0] is the empty string, then append U+002F (/) followed by
+   U+002E (.) to <var>output</var>.
 
    <li><p><a for=list>For each</a> <var>segment</var> of <var>url</var>'s <a for=url>path</a>:
    append U+002F (/) followed by <var>segment</var> to <var>output</var>.
--- a/url.bs
+++ b/url.bs
@@ -2536,7 +2536,7 @@ then runs these steps, returning an <a>ASCII string</a>:
   <p>Otherwise:
 
   <ol>
-   <li><p>If <var>url</var>'s <a for=url>host</a> is null, <var>url</var> <a>is not special</a>,
+   <li><p>If <var>url</var> <a>is not special</a>, <var>url</var>'s <a for=url>host</a> is null,
    <var>url</var>'s <a for=url>path</a>'s <a for=list>size</a> is greater than 1, and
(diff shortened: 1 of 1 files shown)


annevk:
This only affects file URLs, right? Not all special URLs. I wonder if we should make that more explicit. There's a couple other cases where the differences between non-file special URLs and file URLs have caused issues, though turning specialness into a tri-state at this point does not seem super attractive either.

rmisev:
> This only affects file URLs, right? Not all special URLs.

Yes. Non-file special URLs always have not empty host, so they aren't affected. The condition `url’s sheme is not "file:"` can be used here as well, but 5.1. step is intended for non-special URLs, so I added `url is not special`. Which one is clearer?

annevk (commented):
That's a good point. Let's do the not special check first though so we start from the most significant component.

annevk (approved):

annevk:
Perhaps @alwinb also wants to review this? And @domenic would ideally verify the tests using whatwg-url.

Writing this I wonder if another fix here could be that file URLs never have a null host. Did you look at that? (That would also allow for removal of step 3 in the serializer and at least one check elsewhere would be simplified so perhaps that's a bigger change best tackled separately.)

domenic:
Tested against whatwg-url with https://github.com/web-platform-tests/wpt/pull/25989 and this works. Happy to continue testing other simplifications that folks suggest.

I wonder also if it'd be useful to add a new test suite that uses all of urltestdata.json to test idempotency.

alwinb:
Hmm. Resolved file URLs should have a non-null host I think. For a moment I thought that maybe I failed to enforce that in the changes with #405, but I'm not sure now, I guess it was possible before. Anyway, if you can end up with resolved file URLs with a non-null host, I think it's better to change that than to change the serialiser. 

The ambiguity can only arise with a null host, and the host being non-null is (or should be) a property of resolved special URLs, so I think the serialiser was correct.

annevk:
Well, the serializer cannot be entirely correct due to step 3 also existing, right? And there's step 2.1.4 of the scheme state which should not have to check for null if that is true. (I think that's all.)

rmisev:
Ok, updated URL parser to ensure file url's host is not null.

annevk (approved):
Thank you @rmisev. I like this a lot. I'll wait for @domenic to verify and would appreciate a quick skim from @alwinb as well.

alwinb:
@annevk that's true. Step 3 does a bit of resolving inside the serializer, and the check for a null host in 2.1.4 is superfluous if you maintain the invariant that (resolved) file URLs have a non-null host. I don't see any other places either, but I'm not _that_ familiar with this parser so I may have missed something. It seems removing step 3. in the serializer will also solve the idempotency issue, just then you allow null hosts in file URLs. 

You could choose to merge this request and make larger changes later, for example together with changes such as for #548. I would try to move any resolving logic out of the serializer, but also prepare for moving it out of the parser. That's another, much larger topic though.

alwinb:
Oh I missed the latest changes! I'l have a look in a moment.

alwinb:
Yeah looks good.

domenic (approved):
These changes also pass the web platform tests PR and existing suite.

PR #554, opened 2020-10-13
Make it easier to talk about parts of the URL from outside the URL spec

fantasai:
This allows links like [=URL fragment=].

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -1216,18 +1216,18 @@ unified model would be, please file an issue.
 <p>A <dfn export id=concept-url lt="URL|URL record">URL</dfn> is a universal identifier. To
 disambiguate from a <a>valid URL string</a> it can also be referred to as a <a for=/>URL record</a>.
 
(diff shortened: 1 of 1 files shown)


annevk:
Could you share some context? I'm not entirely sure how I feel about this as this is a specific member of a concept and not a standalone concept at the moment.

fantasai:
Just seems that in sentences like

> When a [=url/fragment|URL fragment=] targets an element,  the '':target'' pseudo-element can be used to select it,

it's a bit redundant to have to specify the namespacing structure explicitly when the prose itself has the namespacing built in. I can't think of a reason why the simple term "URL fragment" can't be taken as unambiguously linking to this definition.

annevk:
I think I'm okay with the status quo of making things slightly more awkward for informal sentences such as that. I'd rather not give the impression that works for a formal definition.

annevk:
(There's also the case to consider that we might want to define fragment as a standalone thing in the future, if other objects than a URL can hold one.)

Issue #557, opened 2020-10-19
Percent-encoding of query wrong for ISO-2022-JP


annevk:
If you look at the number of state transitions (hint: count `%1B` instances) in the example below you can see that invoking the encoder code-point-by-code-point does not match implementations.
```js
 function encode(input, output, desc) {
   test(function() {
     var a = document.createElement("a") // <a> uses document encoding for URL's query
     a.href = "https://example.com/?" + input
     assert_equals(a.search.substr(1), output) // remove leading "?"
   }, "iso-2022-jp encoder: " + desc)
 }

 encode("\uFF90\u4F69", "%1B$B%_PP%1B(B");
 encode("\uFF90a\u4F69", "%1B$B%_%1B(Ba%1B$BPP%1B(B");
```
There might be a related thing here, discovered in https://github.com/web-platform-tests/wpt/pull/26158 by @JKingweb, that implementations skip invoking encoders if the input is entirely in the ASCII range. (Again, something that uniquely applies to ISO-2022-JP. Form submission should be checked here as well before deciding on something.)

achristensen07:
We effectively put the code points into a buffer when parsing them then call the encoder once on the whole buffer when transitioning from query state to fragment state or hitting the "EOF code point".  This makes more sense to me than the current spec when using ISO-2022-JP.

achristensen07:
Alternatively, encoding as a stream should fix this, right?

annevk:
I think we should probably do both. I think we want to use a buffer so the URL parser doesn't itself have to keep an encoder alive, but then we do need to process things code-point-for-code-point to do error handling correctly. (I could see an argument for letting the URL parser hold onto an encoder though and passing that into percent-encoding and friends rather than an encoding.)

I filed https://github.com/whatwg/encoding/issues/235 for Encoding standard cleanup as whatever we do here we can no longer use the encode hook it provides.

PR #558, opened 2020-10-20
Refactor query state to operate on a buffer

annevk:
   * Firefox
   * Safari
   * https://github.com/web-platform-tests/wpt/pull/26158
   * Chrome: https://bugs.chromium.org/p/chromium/issues/detail?id=1140965
   * Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1671908
   * Safari: https://bugs.webkit.org/show_bug.cgi?id=218046

***
<a href="https://whatpr.org/url/558.html" title="Last updated on Nov 2, 2020, 9:43 AM UTC (38f6ae8)">Preview</a> | <a href="https://whatpr.org/url/558/a19495e...38f6ae8.html" title="Last updated on Nov 2, 2020, 9:43 AM UTC (38f6ae8)">Diff</a>

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -2432,9 +2432,33 @@ string <var>input</var>, optionally with a <a>base URL</a> <var>base</var>, opti
       <p>then set <var>encoding</var> to <a>UTF-8</a>.
       <!-- https://simon.html5.org/test/url/url-encoding.html -->
 
-     <li><p>If <var>state override</var> is not given and <a>c</a> is U+0023 (#), then set
-     <var>url</var>'s <a for=url>fragment</a> to the empty string and state to
-     <a>fragment state</a>.
+     <li>
+      <p>If one of the following is true:
+
+      <ul class=brief>
+       <li><p><var>state override</var> is not given and <a>c</a> is U+0023 (#)
+       <li><p><a>c</a> is the <a>EOF code point</a>
+      </ul>
+
+      <p>then:
+
+      <ol>
+       <li><p>Let <var>queryPercentEncodeSet</var> be the <a>special-query percent-encode set</a> if
+       <var>url</var> <a>is special</a>; otherwise the <a>query percent-encode set</a>.
+
+       <li>
+        <p><a for=string>Percent-encode after encoding</a>, with <var>encoding</var>,
+        <var>buffer</var>, and <var>queryPercentEncodeSet</var>, and append the result to
+        <var>url</var>'s <a for=url>query</a>.
+
+        <p class=note>This operation cannot be invoked code-point-for-code-point due to the stateful
+        <a>ISO-2022-JP encoder</a>.
+
+       <li><p>Set <var>buffer</var> to the empty string.
+
+       <li><p>If <a>c</a> is U+0023 (#), then set <var>url</var>'s <a for=url>fragment</a> to
+       the empty string and state to <a>fragment state</a>.
+      </ol>
 
      <li>
       <p>Otherwise, if <a>c</a> is not the <a>EOF code point</a>:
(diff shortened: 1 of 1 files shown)


domenic (approved):
This seems to pass all the tests when implemented in jsdom/whatwg-url. It's more complicated, but I assume it's for a good reason related to #557...

annevk:
Thanks, it's incomplete still as the percent-encoding algorithm will also need changes. I do think using a buffer coupled with https://github.com/whatwg/encoding/issues/235#issuecomment-712820914 gives us the "nicest" way of making this work, but happy to receive feedback otherwise. (The one weirdness is that we don't have a per-code-point percent-encode with arbitrary encoding anymore, but no caller really needs that. The UTF-8 caller can just pass its single code point as a string.)

annevk:
The corresponding Encoding PR for the second commit in this PR hasn't landed yet and as such it'll fail Build, but it's still worth reviewing at this point I think, especially for non-editorial concerns.

annevk:
I plan to rebase this Monday and make it pass Build now that https://github.com/whatwg/encoding/pull/238 has landed.

url.bs
```diff
@@ -217,81 +217,70 @@ inclusive, and U+007E (~).
...
 
  <li><p>Return <var>output</var>.
 </ol>
 
 <p>To <dfn for="code point" id=utf-8-percent-encode>UTF-8 percent-encode</dfn> a
 <a for=/>code point</a> <var>codePoint</var> using a <var>percentEncodeSet</var>, return the result
-of running <a for="code point">percent-encode after encoding</a> with <a for=/>UTF-8</a>,
-<var>codePoint</var>, and <var>percentEncodeSet</var>.
+of running <a for=string>percent-encode after encoding</a> with <a for=/>UTF-8</a>,
+<var>codePoint</var> as <a for=/>string</a>, and <var>percentEncodeSet</var>.
```
domenic, line 278:
```suggestion
<var>codePoint</var> as a <a for=/>string</a>, and <var>percentEncodeSet</var>.
```

url.bs
```diff
@@ -217,81 +217,70 @@ inclusive, and U+007E (~).
...
 
 <p>To <dfn export for=string>UTF-8 percent-encode</dfn> a <a for=/>string</a> <var>input</var> using
 a <var>percentEncodeSet</var>, return the result of running
 <a for=string>percent-encode after encoding</a> with <a for=/>UTF-8</a>, <var>input</var>,
-<var>percentEncodeSet</var>, and false.
+<var>percentEncodeSet</var>.
```
domenic, line 283:
```suggestion
and <var>percentEncodeSet</var>.
```

domenic (approved):
It looks like I can't really test this with jsdom/whatwg-url since we're still UTF-8 only. So my review of the second commit is just editorial.

Maybe I should sink some time into fixing that, but doing it properly would probably require a from-scratch implementation of the Encoding standard which exposes all of the spec-level primitives. Probably not something I can tackle any time soon...

domenic:
Can we get web platform tests for this that are URL-specific? Probably not part of the JSON files though.

url.bs
```diff
@@ -217,81 +217,70 @@ inclusive, and U+007E (~).
...
- <li><p>Let <var>bytes</var> be the result of <a lt=encode>encoding</a> <var>codePoint</var> using
- <var>encoding</var>.
+ <li><p>Let <var>encoder</var> be the result of <a>getting an encoder</a> from <var>encoding</var>.
 
- <li>
-  <p>If <var>bytes</var> starts with 0x26 (&amp;) 0x23 (#) and ends with 0x3B (;), then:
-
-  <ol>
-   <li><p>Let <var>output</var> be <var>bytes</var>, <a>isomorphic decoded</a>.
+ <li><p>Let <var>inputQueue</var> be <var>input</var> onverted to an <a for=/>I/O queue</a>.
```
andreubotella, line 227:
```suggestion
 <li><p>Let <var>inputQueue</var> be <var>input</var> converted to an <a for=/>I/O queue</a>.
```

annevk:
Thanks for the feedback! I put up some fairly basic tests over at https://github.com/web-platform-tests/wpt/pull/26317 but they should be easy to expand upon if we all think that format is workable.

PR #582, opened 2021-02-21
Prevent the pathname setter from erasing the path of path-only URLs, …

### LINKED ISSUE #581: Path setter allows cannot-be-a-base URL to be created without the flag
Testing this out using the JSDom implementation:

```
> var url = new whatwgURL.URL("foo:/hello/world?someQuery")

> url.href
"foo:/hello/world?someQuery"

> url.pathname = ""
> url.href
"foo:?someQuery"

> url.pathname = "test"
> url.href
"foo:/test?someQuery"

> var url = new whatwgURL.URL("foo:?someQuery")
> url.pathname = "test"
> url.href
"foo:?someQuery"
```

When setting `pathname` to the empty string, we create a cannot-be-a-base URL (`"foo:?someQuery"`), but it is not flagged as such because the parser won't set this flag after "path start" state, which is where the setter begins. Therefore we can continue to set the `pathname` to something else. But if we parse that same string again, the cannot-be-a-base flag is correctly set and the subsequent `pathname` setter has no effect.

karwa:
…as that would make them cannot-be-a-base URLs.

Fixes #581 

JSDom patch: https://github.com/jsdom/whatwg-url/pull/178

   * Me 😅
   * …
- [X] [Tests](https://github.com/web-platform-tests/wpt) are written and can be reviewed and commented upon at:
   * https://github.com/web-platform-tests/wpt/pull/27720
   * Chrome: …
   * Firefox: …
   * Safari: …

***
<a href="https://whatpr.org/url/582.html" title="Last updated on Jun 17, 2021, 8:12 AM UTC (405c917)">Preview</a> | <a href="https://whatpr.org/url/582/557567c...405c917.html" title="Last updated on Jun 17, 2021, 8:12 AM UTC (405c917)">Diff</a>

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -2298,6 +2298,10 @@ these steps:
 
        <li><p>If <a>c</a> is not U+002F (/), then decrease <var>pointer</var> by 1.
       </ol>
+
+     <li><p>Otherwise, if <var>state override</var> is given and <var>url</var>'s
+      <a for=url>host</a> is null, <a for=list>append</a> the empty string to <var>url</var>'s
+      <a for=url>path</a>
     </ol>
 
    <dt><dfn>path state</dfn>
@@ -3439,6 +3443,7 @@ Jeffrey Yasskin,
 Joe Duarte,
 Joshua Bell,
 Jxck,
+Karl Wagner,
 田村健人 (Kent TAMURA),
 Kevin Grandon,
 Kornel Lesiński,
--- a/url.bs
+++ b/url.bs
@@ -2300,8 +2300,8 @@ these steps:
       </ol>
 
      <li><p>Otherwise, if <var>state override</var> is given and <var>url</var>'s
-      <a for=url>host</a> is null, <a for=list>append</a> the empty string to <var>url</var>'s
-      <a for=url>path</a>
+     <a for=url>host</a> is null, <a for=list>append</a> the empty string to <var>url</var>'s
+     <a for=url>path</a>.
     </ol>
 
    <dt><dfn>path state</dfn>


url.bs
```diff
@@ -2298,6 +2298,10 @@ these steps:
 
        <li><p>If <a>c</a> is not U+002F (/), then decrease <var>pointer</var> by 1.
       </ol>
+
+     <li><p>Otherwise, if <var>state override</var> is given and <var>url</var>'s
+      <a for=url>host</a> is null, <a for=list>append</a> the empty string to <var>url</var>'s
+      <a for=url>path</a>
```
annevk, line 2304: Why only when host is null? It seems this needs to generally happen.

annevk (commented):
Thanks for spotting this, this does look like a somewhat serious oversight.

url.bs
karwa: Because non-special URLs with a host (e.g. `foo://somehost`) can have an empty string as a path. The issue is limited to path-only URLs.

domenic:
Editorially LGTM and matches the tests and jsdom/whatwg-url implementation.

In terms of interop and implementation alignment, of the three tests added in https://github.com/web-platform-tests/wpt/pull/27720:

- Safari passes test 1 and 3, but not test 2
- Firefox and Chrome pass test 1, but not test 2 or test 3
- jsdom/whatwg-url, and thus the current spec, passes tests 1 and 2 but not test 3

Should we consider changing the spec for test 2? I don't have a strong intuition on the model here though so maybe there's a reason that would be bad.

annevk:
Chrome and Firefox still don't have non-special URLs implemented correctly so I rather not count them.

Safari doesn't seem to allow for removal of the path if there is a host for non-special URLs. I tend to side with OP that it should be possible. That is, if `foo://host` doesn't create a path and `foo://host/` does, it seems nice if you can get from the latter to the former by setting `pathname` to the empty string.

cc @achristensen07

karwa:
I think it is valid (or, well, _not disallowed_) for some particular schemes/application to treat an empty path differently from a "/" path.

For instance if you're connecting to a remote server: for some hypothetical scheme, `myproto://user@host` could mean to connect to the user's home directory, and `myproto://user@host/` could mean to connect to the remote server's root directory. It may/may not be a good design choice, but if it's allowed, somebody could be relying on it.

rmisev (approved):
LGTM

annevk:
@achristensen07 any thoughts on this?

TimothyGu:
It looks like this PR only changes the behavior of test 3 in https://github.com/web-platform-tests/wpt/pull/27720, and Safari already implements the new behavior for test 3. @annevk, would this be enough to count as implementer interest?

achristensen07:
I haven't gone through the setters like I have the constructor.   I think test 2 shows a bug in WebKit that I'm willing to try to change.  NSURLComponents already has the behavior that would make test 2 pass, and parsing foo://somehost doesn't add a slash to the path, so I think the path setter shouldn't either, unless called with something that isn't empty and doesn't start with a slash.

annevk:
Thanks @achristensen07! @karwa can you file a bug against WebKit for that? I don't think we need bugs for Chrome/Firefox as they haven't yet caught up here.

karwa:
Filed https://bugs.webkit.org/show_bug.cgi?id=227073

annevk:
I filed https://github.com/nodejs/node/issues/39059 on Node.js.

annevk:
Thanks @karwa for finding and fixing this! 🎉

Issue #319, opened 2017-06-03
"Forbidden host code point" should be defined without U+0025


GPHemsley:
The term "forbidden host code point" is used exactly twice within this spec, and in one case it excludes U+0025 (%):

> If asciiDomain contains a forbidden host code point, validation error, return failure.

> If input contains a forbidden host code point excluding U+0025 (%), validation error, return failure.

I think it is easier for implementations to add to a definition rather than take away from it, so I propose removing U+0025 from the definition and special-casing it where it is _included_:

> If asciiDomain contains a forbidden host code point or U+0025 (%), validation error, return failure.

> If input contains a forbidden host code point, validation error, return failure.

Or else creating a second definition that relates to the existing one.

annevk:
It's also used in https://url.spec.whatwg.org/#valid-opaque-host-string now and I think that indeed argues for changing this as it's a bit ambiguous (to me).

alwinb:
Something similar happens with [URL-path-segment strings] which uses the phrase:

> zero or more URL units excluding U+002F (/) and U+003F (?), that together are not a single-dot path segment or a double-dot path segment.

What do you think about rephrasing things to use sets of _allowed_ codepoints for each of username, password, opaque-host, path-segment, query, and fragment instead?

[URL-path-segment strings]: https://url.spec.whatwg.org/#url-path-segment-string

annevk:
That might be reasonable, though I'd rather approach it in small incremental steps as each of those has the potential for regressions.

PR #584, opened 2021-02-24
[Editorial] Guarantee that strings encoded with the component encode-set are valid opaque host strings.

karwa:
Users require guidance about how to encode data as an opaque host string - for instance, it is common to want to encode a filesystem path as a hostname to create an `http+unix:` URL.

Moreover, the `host`/`hostname` setter is unique in that it will fail if the new value contains forbidden host code units, unlike other setters which automatically percent-encode any invalid characters, so it is important that users correctly encode the new value in order for the operation to succeed.

I would like to advise users to encode their hostnames with the component encode-set, and it would be nice if there was an explicit mention in the standard to back this advice up - as there is for the path, query and fragment.

***
<a href="https://whatpr.org/url/584.html" title="Last updated on Mar 3, 2021, 10:44 AM UTC (73b10cf)">Preview</a> | <a href="https://whatpr.org/url/584/557567c...73b10cf.html" title="Last updated on Mar 3, 2021, 10:44 AM UTC (73b10cf)">Diff</a>

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -205,7 +205,7 @@ U+0024 ($) to U+0026 (&amp;), inclusive, U+002B (+), and U+002C (,).
 <p class=note>This is used by <cite>HTML</cite> for
 {{NavigatorContentUtils/registerProtocolHandler()}}, and could also be used by other standards to
 percent-encode data that can then be embedded in a <a for=/>URL</a>'s <a for=url>path</a>,
-<a for=url>query</a>, or <a for=url>fragment</a>. Using it with
+<a for=url>query</a>, or <a for=url>fragment</a>; or in an <a for=/>opaque host</a>. Using it with
 <a for=string>UTF-8 percent-encode</a> gives identical results to JavaScript's
 <a method><code>encodeURIComponent()</code> [sic]</a>. [[HTML]] [[ECMA-262]]
 
@@ -3439,6 +3439,7 @@ Jeffrey Yasskin,
 Joe Duarte,
 Joshua Bell,
 Jxck,
+Karl Wagner,
 田村健人 (Kent TAMURA),
 Kevin Grandon,
 Kornel Lesiński,


karwa:
Also, the existing note:

> This is used by  HTML  for  registerProtocolHandler() , and could also be used by other standards to percent-encode data that can then be embedded in a  URL  ’s  path , query , or  fragment.

Does not seem to be entirely true (I think?). Since "component" is not a superset of the "special query" encode set, wouldn't these strings need additional encoding when embedding in a special URL's query? I realise that the current definition aligns with implementations, but it would be nice if there was a way to add U+0027 ('), so that it becomes a true superset.

annevk:
Yeah, that seems right. It's still valid to put a `'` there, but it will be encoded. I suspect extending the note to cover that is the easiest solution here.

I don't think we can change implementations here by the way. I guess I would prefer to not encode `'` if that were possible, but...

karwa:
I see. Thinking about it some more, the note only says that a string encoded with the component set _can_ be embedded in a query string - and it still can. I don't think having an unencoded `'` in the query of a special URL leads to any ambiguities. And in general strings encoded via the component set and embedded will look different to what the parser would give, because the component set conservatively encodes more than the parser does. Maybe it isn't worth calling that out.

But I'd still like to add the mention about opaque host strings if that's agreeable (which just makes a part of the status quo explicit - NFC), because these definitions of allowed/disallowed code points have changed in the past.

url.bs
```diff
@@ -205,7 +205,7 @@ U+0024 ($) to U+0026 (&amp;), inclusive, U+002B (+), and U+002C (,).
 <p class=note>This is used by <cite>HTML</cite> for
 {{NavigatorContentUtils/registerProtocolHandler()}}, and could also be used by other standards to
 percent-encode data that can then be embedded in a <a for=/>URL</a>'s <a for=url>path</a>,
-<a for=url>query</a>, or <a for=url>fragment</a>. Using it with
+<a for=url>query</a>, or <a for=url>fragment</a>, and is also a <a>valid opaque-host string</a>. Using it with
```
annevk, line 208:
```suggestion
<a for=url>query</a>, or <a for=url>fragment</a>, or <a for=/>opaque host</a>. Using it with
```
would be more consistent I think. I did notice while looking at this that valid opaque-host string doesn't say that `%` is fine. Perhaps the forbidden host code points should be two sets after all.

karwa: Sure, that's good - I like how it now says "embedded ... in an opaque host" and not just "is a valid opaque host".

As for `%`, it says a valid opaque-host string is made of URL Units, which are URL code points (doesn't include `%`) and percent-encoded bytes (which do of course include `%`), so I think it's implied that `%` is fine, but _only_ when used for percent-encoding, and not just generally as part of the host's name.

annevk: It doesn't read that well to me, but it's a separate issue.

annevk (approved):

url.bs
annevk: Ah: https://github.com/whatwg/url/issues/319.

annevk:
Thanks @karwa!

luismedicina23 (commented):
@karwa

luismedicina23 (commented):
@karwa

PR #604, opened 2021-05-17
Make hostname state return early if a port is given

### LINKED ISSUE #601: Should hostname setter no-op on colon (:)?
The spec currently provides for the following:

```js
u = new URL("http://example.net/path");
u.hostname = "example.com:8080";
console.assert(u.hostname === "example.com");  // FF & WK: "example.net"
console.assert(u.port === "");

u = new URL("http://example.net:8080/path");
u.hostname = "example.com:";
console.assert(u.hostname === "example.com");  // FF & WK: "example.net"
console.assert(u.port === "8080");
```

(The test cases came from [WPT](https://github.com/web-platform-tests/wpt/blob/e75d075b7c27fb29e83ca9f5ab785995fd09736f/url/resources/setters_tests.json#L1155-L1176).)

However, no one actually implements this. Firefox and Safari both reject the given hostname value if it contains a port part, and leave the hostname unchanged.

Shall we align with Firefox and Safari?

<details><summary>(OTOH, Chrome does something that makes no sense, so I'd like to exclude it from the discussion for now.)</summary>

```js
u = new URL("http://example.net/path");
u.hostname = "example.com:8080";
console.log(u.hostname);   // prints ""
console.log(u.href);       // prints "http://example.com:8080/path"
```

</details>

TimothyGu:
Fixes #601.

   * Firefox (already implemented)
   * Safari (already implemented)
   * https://github.com/web-platform-tests/wpt/pull/29021
   * Chrome: https://crbug.com/1018721
   * Node.js: https://github.com/nodejs/node/issues/38710
   * whatwg-url: https://github.com/jsdom/whatwg-url/pull/183

***
<a href="https://whatpr.org/url/604.html" title="Last updated on May 19, 2021, 4:17 PM UTC (974b094)">Preview</a> | <a href="https://whatpr.org/url/604/8c24e54...974b094.html" title="Last updated on May 19, 2021, 4:17 PM UTC (974b094)">Diff</a>

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -2024,6 +2024,9 @@ these steps:
        <li><p>If <var>buffer</var> is the empty string, <a>validation error</a>, return failure.
        <!-- No URLs with port, but without host. -->
 
+       <li><p>If <var>state override</var> is given and <var>state override</var> is
+       <a>hostname state</a>, then return.
+
        <li><p>Let <var>host</var> be the result of <a>host parsing</a> <var>buffer</var> with
        <var>url</var> <a>is not special</a>.
 
@@ -2032,9 +2035,6 @@ these steps:
        <li><p>Set <var>url</var>'s <a for=url>host</a> to
        <var>host</var>, <var>buffer</var> to the empty string,
        and <var>state</var> to <a>port state</a>.
-
-       <li><p>If <var>state override</var> is given and <var>state override</var> is
-       <a>hostname state</a>, then return.
       </ol>
 
      <li>


annevk (approved):
Thanks!

(I wondered if this should be a validation error, but that doesn't seem to make sense for API-specific conditions.)

annevk:
Could you also add a test for `location.hostname` throwing for the same input?

TimothyGu:
@annevk What do you mean? The [setter steps](https://html.spec.whatwg.org/multipage/history.html#dom-location-hostname) don't seem to have any provision for throwing on URL parsing failure.

annevk:
Oh, sorry, I somehow assumed Firefox's behavior was correct. It seems that Safari does navigate to the current URL. And Chrome ignores the `:` as per the current specification. So it does seem like something that ought to be tested, but the pass condition should probably match Safari?

TimothyGu:
@annevk I agree that more location tests would be useful, though to be fair, Chrome doesn't exactly follow the current spec:

Screenshot from 2021-05-19 04-58-59

And if the current URL doesn't have a port while the `hostname` has a port, the hostname's port is used.

In fact, all the Location properties could probably use some more setter tests.

----

By the way, I don't have write permissions to this repo, so you'll have to merge this PR.

annevk:
I see, I had not tested Chrome when setting from a URL with a port. Alright, I guess we can merge this and consider the `Location` issue somewhat distinct. Could you file an issue to track that? Either against WPT or HTML?

annevk:
Actually, I don't see why this needs to return failure. I think we only return failure at the moment if it's relevant for the caller. Otherwise we just return. (That's also why I thought the `Location` setter would throw.)

TimothyGu:
> I guess we can merge this and consider the `Location` issue somewhat distinct. Could you file an issue to track that? Either against WPT or HTML?

There's already https://github.com/web-platform-tests/wpt/issues/4103 to track Location stuff.

> Actually, I don't see why this needs to return failure. I think we only return failure at the moment if it's relevant for the caller. Otherwise we just return.

Fixed.

annevk:
Thanks!

rmisev:
It seems a bit late now, but I think if host with port is given, then 2.2. step (added by this PR) must emit validation error.
Such practices are used in other steps (for example in 3.2.) when given value is not accepted.
If you agree, I can prepare PR.

annevk:
@rmisev I considered that, but I thought we didn't do it for override cases?

rmisev:
Yeah, it seems the only case, when override emits validation error, is 3.2. step in host & hostname state.

Issue #560, opened 2020-11-20
IPv4 host parser + site definition seems potentially dangerous.


MattMenke2:
Per the IPv4 parser, http://127.0.0.127.1 is a valid, non-IPv4 host.  Similarly, http://foo.0.127.1 is a valid, non-IPv4 host.  Per the Public Suffix List spec, the prevailing rule that should apply to them (since they have no matches in the list) is "*", so their eTLD is just "1", and so their site (defined in the fetch spec) is "http", "127."1".

This is strange for two reasons:
1)  "http://127.1" is, per the IPv4 parsing spec, equivalent to "http://0.0.127.1", which seems unsafe.  I believe it's a reserved IP, anyways, but still not great.  FireFox maps this to "http://127.0.0.1", at least when passed to the Javascript URL constructor, which seems even less safe.  So trying to create a URL or origin from a such a site seems like it could potentially lead to security bugs.  Admittedly, this is not something the web platform currently does, though can happen, e.g., when displaying the "URL" a cookie is set on to the user.
2)  "http://127.0.0.0.1" looks a lot like "http://127.0.0.1", and we always tell users to look at the end of a hostname, so doesn't in general seem very safe to display to the user.

Now, admittedly, trying to resolve "127.0.0.0.1" at the DNS layer should generally fail, but the situation here seems a bit troubling.

I ran into this because Chromium is trying to store the Network Partition Key's site | origin as an origin object, but a fuzzer recently discovered an assert failure when creating an origin with a host of "0.1" (which is recognized as not being in the canonical form of "0.0.0.1").  Obviously, we could just make it hold an origin | (scheme + hostname), and we can adhere to the spec, the described behavior just doesn't seem that safe to me.

I'd rather just reject an host that ends in a number but isn't a valid IPv4 IP, so filing this to see if anyone else has any thoughts on this.

annevk:
cc @whatwg/security

mozfreddyb:
A bit on a tangent, but re 1), a missing dot is filled in so far that the last number is treated as an integer to fill the bytes that are still required. I.e., 127.1 ought to be "127.0.0.1".
However this may be resolved, it needs to take hexadecimal, octal, (dotted and undotted) into account.

mikewest:
> I'd rather just reject an host that ends in a number but isn't a valid IPv4 IP, so filing this to see if anyone else has any thoughts on this.

This sounds reasonable to me.

> a missing dot is filled in so far that the last number is treated as an integer

[...]

PR #619, opened 2021-07-08
Reject non-IPv4 hostnames that end in numbers.

MattMenke2:
This is aimed at addressing issue #560.  If the last component of a URL's hostname is numeric, it's parsed as an IPv4 hostname, and if that fails, the URL's host is rejected.  e.g., "foo.0", "bar.0.09", "a.1.2.0x.", "1.2.3.4.5" were all previously considered valid non-IPv4 hostnames, but are now all rejected.  See #560 for more discussion on why allowing these are potentially concerning.

   * Chrome
   * FireFox
   * https://github.com/web-platform-tests/wpt/pull/29666
   * Chrome: https://crbug.com/1237032
   * Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1723456
   * Safari: https://bugs.webkit.org/show_bug.cgi?id=228826

***
<a href="https://whatpr.org/url/619.html" title="Last updated on Aug 5, 2021, 2:47 PM UTC (c09bcfa)">Preview</a> | <a href="https://whatpr.org/url/619/0672f2e...c09bcfa.html" title="Last updated on Aug 5, 2021, 2:47 PM UTC (c09bcfa)">Diff</a>

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -655,16 +655,39 @@ runs these steps:
  <li><p>If <var>asciiDomain</var> contains a <a>forbidden host code point</a>,
  <a>validation error</a>, return failure.
 
- <li><p>Let <var>ipv4Host</var> be the result of <a lt="IPv4 parser">IPv4 parsing</a>
- <var>asciiDomain</var>.
-
- <li><p>If <var>ipv4Host</var> is an <a>IPv4 address</a> or failure, return
- <var>ipv4Host</var>.
+ <li><p>If <var>asciiDomain</var> <a lt="ends in a number"> ends in a numer, return
+ the result of <a lt="IPv4 parser">IPv4 parsing</a> <var>asciiDomain</var>.
 
  <li><p>Return <var>asciiDomain</var>.
 </ol>
 
 <hr>
+
+<p>The <dfn id=ends-in-a-number>ends in a number checker</dfn> takes a string <var>input</var> and
+then runs these steps:
+
+<ol>
+ <li><p>Let <var>parts</var> be the result of <a>strictly splitting</a> <var>input</var> on
+ U+002E (.).
+
+ <li>
+  <p>If the last <a for=list>item</a> in <var>parts</var> is the empty string, then:
+
+  <ol>
+   <li><p>If <var>parts</var>'s <a for=list>size</a> is 1, return false.
+
+   <li>Otherwise, <a for=list>remove</a> the last <a for=list>item</a> from <var>parts</var>.
+  </ol>
+
+ <li><p>Let <var>last</var> be the last <a for=list>item</a> from <var>parts</var>.
+
+ <li>If parsing <var>last</var> as an <a lt="IPv4 number parser">IPv4 number</a> does not
+ return failure, return true.
+
+ <li><p>If <var>last</var> is non-empty and contains only <a>ASCII digits</a>, return true.
+
+ <li><p>Return false.
+</ol>
 
 <p>The <dfn id=concept-ipv4-parser>IPv4 parser</dfn> takes a string <var>input</var> and then runs
 these steps:
@@ -692,7 +715,8 @@ these steps:
         but if it somehow is this conditional makes sure we can keep going. -->
   </ol>
 
- <li><p>If <var>parts</var>'s <a for=list>size</a> is greater than 4, then return <var>input</var>.
+ <li><p>If <var>parts</var>'s <a for=list>size</a> is greater than 4, then <a>validation error</a>,
+ return failure.
 
  <li><p>Let <var>numbers</var> be an empty <a for=/>list</a>.
 
@@ -700,16 +724,10 @@ these steps:
   <p><a for=list>For each</a> <var>part</var> of <var>parts</var>:
 
   <ol>
-   <li>
-    <p>If <var>part</var> is the empty string, then return <var>input</var>.
-
-    <p class="example no-backref" id=example-c2afe535><code>0..0x300</code> is a
-    <a>domain</a>, not an <a>IPv4 address</a>.
-
    <li><p>Let <var>result</var> be the result of <a lt="IPv4 number parser">parsing</a>
    <var>part</var>.
 
-   <li><p>If <var>result</var> is failure, then return <var>input</var>.
+   <li><p>If <var>result</var> is failure, then <a>validation error</a>, return failure.
 
    <li><p>If <var>result</var>[1] is true, then set <var>validationError</var> to true.
 
@@ -754,7 +772,9 @@ these steps:
 <p>The <dfn>IPv4 number parser</dfn> takes a string <var>input</var> and then runs these steps:
 
 <ol>
- <li><p>Let <var>validationError</var> be false.
+ <li><p>If <var>input</var> is the empty string, then return failure.
+
+<li><p>Let <var>validationError</var> be false.
 
  <li><p>Let <var>R</var> be 10.
 
--- a/url.bs
+++ b/url.bs
@@ -686,6 +686,10 @@ then runs these steps:
 
  <li><p>If <var>last</var> is non-empty and contains only <a>ASCII digits</a>, return true.
 
+ <p class="note no-backref">This can happen if <var>last</var> starts with "0" so the
+ <a lt="IPv4 number parser">IPv4 number parser</a> tries to parse it as octal, but it is not
+ a valid octal number, as is the case with, for example, "09".
+
  <li><p>Return false.
 </ol>
 
--- a/url.bs
+++ b/url.bs
@@ -655,7 +655,7 @@ runs these steps:
  <li><p>If <var>asciiDomain</var> contains a <a>forbidden host code point</a>,
  <a>validation error</a>, return failure.
 
- <li><p>If <var>asciiDomain</var> <a lt="ends in a number"> ends in a numer, return
+ <li><p>If <var>asciiDomain</var> <a lt="ends in a number checker"> ends in a numer, return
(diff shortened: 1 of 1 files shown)


url.bs
```diff
@@ -655,17 +655,40 @@ runs these steps:
...
+
+   <li>Otherwise, <a for=list>remove</a> the last <a for=list>item</a> from <var>parts</var>.
+  </ol>
+
+ <li><p>Let <var>last</var> be the last <a for=list>item</a> from <var>parts</var>.
+
+ <li>If parsing <var>last</var> as an <a lt="IPv4 number parser">IPv4 number</a> does not
+ return failure, return true.
+
+ <li><p>If <var>last</var> is non-empty and contains only <a>ASCII digits</a>, return true.
```
TimothyGu, line 687: Maybe mention in a note we could get here for _last_ = "`09`", because IPv4 number parser would treat it as octal?

MattMenke2: Good idea, done

url.bs
```diff
@@ -655,17 +655,44 @@ runs these steps:
  <li><p>If <var>asciiDomain</var> contains a <a>forbidden host code point</a>,
  <a>validation error</a>, return failure.
 
- <li><p>Let <var>ipv4Host</var> be the result of <a lt="IPv4 parser">IPv4 parsing</a>
- <var>asciiDomain</var>.
-
- <li><p>If <var>ipv4Host</var> is an <a>IPv4 address</a> or failure, return
- <var>ipv4Host</var>.
+ <li><p>If <var>asciiDomain</var> <a lt="ends in a number checker"> ends in a numer, return
```
domenic, line 658:
```suggestion
 <li><p>If <var>asciiDomain</var> <a lt="ends in a number checker">ends in a number</a>, return
```

url.bs
```diff
@@ -655,17 +655,44 @@ runs these steps:
...
+ <li><p>Let <var>parts</var> be the result of <a>strictly splitting</a> <var>input</var> on
+ U+002E (.).
+
+ <li>
+  <p>If the last <a for=list>item</a> in <var>parts</var> is the empty string, then:
+
+  <ol>
+   <li><p>If <var>parts</var>'s <a for=list>size</a> is 1, return false.
+
+   <li>Otherwise, <a for=list>remove</a> the last <a for=list>item</a> from <var>parts</var>.
```
domenic, line 679:
```suggestion
   <li><p><a for=list>remove</a> the last <a for=list>item</a> from <var>parts</var>.
```

url.bs
```diff
@@ -754,7 +776,9 @@ these steps:
 <p>The <dfn>IPv4 number parser</dfn> takes a string <var>input</var> and then runs these steps:
 
 <ol>
- <li><p>Let <var>validationError</var> be false.
+ <li><p>If <var>input</var> is the empty string, then return failure.
+
+<li><p>Let <var>validationError</var> be false.
```
domenic, line 781:
```suggestion
 <li><p>Let <var>validationError</var> be false.
```

domenic (approved):
Non-editor review based on implementing jsdom/whatwg-url: LGTM with nits.

url.bs
```diff
@@ -655,17 +655,44 @@ runs these steps:
...
+  <ol>
+   <li><p>If <var>parts</var>'s <a for=list>size</a> is 1, return false.
+
+   <li><p><a for=list>remove</a> the last <a for=list>item</a> from <var>parts</var>.
+  </ol>
+
+ <li><p>Let <var>last</var> be the last <a for=list>item</a> from <var>parts</var>.
+
+ <li>If parsing <var>last</var> as an <a lt="IPv4 number parser">IPv4 number</a> does not
+ return failure, return true.
```
TimothyGu, line 685:
```suggestion
 <li><p>If parsing <var>last</var> as an <a lt="IPv4 number parser">IPv4 number</a> does not
 return failure, return true.
```

url.bs
```diff
@@ -655,17 +655,44 @@ runs these steps:
...
+
+<ol>
+ <li><p>Let <var>parts</var> be the result of <a>strictly splitting</a> <var>input</var> on
+ U+002E (.).
+
+ <li>
+  <p>If the last <a for=list>item</a> in <var>parts</var> is the empty string, then:
+
+  <ol>
+   <li><p>If <var>parts</var>'s <a for=list>size</a> is 1, return false.
```
annevk, line 677:
```suggestion
   <li><p>If <var>parts</var>'s <a for=list>size</a> is 1, then return false.
```

url.bs
```diff
@@ -655,17 +655,44 @@ runs these steps:
...
+
+   <li><p><a for=list>remove</a> the last <a for=list>item</a> from <var>parts</var>.
```
annevk, line 679:
```suggestion
   <li><p><a for=list>Remove</a> the last <a for=list>item</a> from <var>parts</var>.
```

url.bs
```diff
@@ -655,17 +655,44 @@ runs these steps:
...
+  </ol>
+
+ <li><p>Let <var>last</var> be the last <a for=list>item</a> from <var>parts</var>.
```
annevk, line 682:
```suggestion
 <li><p>Let <var>last</var> be the last <a for=list>item</a> in <var>parts</var>.
```
(for consistency with earlier)

url.bs
```diff
@@ -655,17 +655,44 @@ runs these steps:
...
+
+ <li><p>If parsing <var>last</var> as an <a lt="IPv4 number parser">IPv4 number</a> does not
+ return failure, return true.
```
annevk, line 685:
```suggestion
 return failure, then return true.
```

url.bs
annevk: The `<p>` needs to be on a newline as the `<li>` has two children now. Also "then return".

annevk (commented):
This seems like something we should be doing in name of security. Copying @valenting to be sure, but I think I can say Mozilla is supportive of this effort.

annevk:
The tests for this have merged. I'll push a fixup commit for the outstanding comments and merge this Wednesday in case folks have some late comments/concerns.

annevk:
To be sure, @MattMenke2 do you agree with my changes here? Can you file browser bugs?

valenting:
> This seems like something we should be doing in name of security. Copying @valenting to be sure, but I think I can say Mozilla is supportive of this effort.

This seems like the right thing to do.
I have filed https://bugzilla.mozilla.org/show_bug.cgi?id=1723456 for the Firefox work.
Since it seems the WPT tests have already landed, I'll try to implement this as soon as possible.

annevk:
I filed a WebKit bug, but cannot find a Chromium bug. @MattMenke2 @domenic did I miss something?

MattMenke2:
> I filed a WebKit bug, but cannot find a Chromium bug. @MattMenke2 @domenic did I miss something?

https://crbug.com/1149194 is assigned to me and is technically due to this issue (and why I filed this issue), but it's not exactly obvious that it's the same issue.  I've file another bug about this, and assigned it to me.

achristensen07:
Do we really want to make https://example.4294967295/ invalid but https://example.4294967296/ valid?

MattMenke2:
I'm not following - the spec doesn't seem to do that?  If the last value parses as a number, we use the IPv4 parser, and if it fails, we reject the hostname.  The number parser parses both 4294967295 and 4294967296.  There's also the all ASCII digits fallback, but that's not needed here, anyways, since the IPv4 number parser doesn't have a limit of 2^32-1.

achristensen07:
Oh, right.  Sorry for the noise.

blattersturm:
Why'd you folks do this? This breaks compatibility _irreparably_ for our software, see [a forum post](https://forum.cfx.re/t/nui-files-fail-to-load/4934989/10) on our forums.

MattMenke2:
> Why'd you folks do this without offering any means to disable it?
> 
> This breaks compatibility _irreparably_ for our software, see [a forum post](https://forum.cfx.re/t/nui-files-fail-to-load/4934989/10) in this regard.
> 
> I genuinely keep being amazed by how web browser vendors love to break compatibility at random. From the _issue_ you linked earlier, metrics from 'Chrome' don't contain use cases from downstream users of Chromium, nor do we have a way to report such metrics to 'Chrome' in the first place or manpower to track every single little change you people are making.

See issue #560, linked in the first post for more details.  The specs around the host parser, the public suffix list, and the definition of site, are not very compatible in this case.  This was a fix for potential security issues, due to https://foo.127.1 having an eTLD+1 of 127.1, and thus a schemeful site of https://127.1, which as a URL, maps to https://127.0.0.1/.

blattersturm:
> The specs around the host parser, the public suffix list, and the definition of site, are not very compatible in this case. 

Whatever this may mean, sure, but you should be aware web technologies are also often _embedded_ as _libraries_ in third-party applications, where these use cases _may not necessarily apply_. Breaking compatibility based only on 'web technology used in a web browser' is shortsighted.

MattMenke2:
I'm sorry that this hit you, but end user security is our highest priority.  We don't like breaking things for developers - I mean, we are human, and we don't like making folks (justifiably) unhappy.  Breakages are sometimes inevitable, unfortunately, due to the cobbled-together nature of web standards, which often didn't take security into consideration, or overlook how features interact with each other.

In terms of modifying Chromium: Note that the commit that changed behavior had to completely refactor a method, because the new behavior required extra checking not present in the old code.  So you could perhaps come up with a smaller change in terms of production code to make the new code behave like the old code (I think it's the "if (family != CanonHostInfo::IPV4)" block?  It's pretty ugly code, unfortunately).  The rest of the CL was tests, which, yes, you'd need to revert, if you run tests and want them to pass.

PR #633, opened 2021-08-24
[Editorial] Simplify the ends-in-a-number checker

karwa:
I believe this should be functionally identical to what exists already. 

IMO it's more intuitive, because it means the "ends in a number" checker first checks to see if the label is all digits (which obviously means it is a number), and then checks to see if it's a number containing non-digit characters (i.e. just hex numbers). I think it's clearer and more obvious than the current wording, which first invokes the more expensive IPv4 number parser and then adds a second check because of particular failure conditions.

Writing it this way in the spec gives implementors assurance that this is a safe optimisation.

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -681,15 +681,14 @@ steps:
 
  <li><p>Let <var>last</var> be the last <a for=list>item</a> in <var>parts</var>.
 
- <li><p>If parsing <var>last</var> as an <a lt="IPv4 number parser">IPv4 number</a> does not
- return failure, then return true.
-
  <li>
   <p>If <var>last</var> is non-empty and contains only <a>ASCII digits</a>, then return true.
 
-  <p class=note>This can happen if <var>last</var> starts with "<code>0</code>" so the
-  <a lt="IPv4 number parser">IPv4 number parser</a> tries to parse it as octal, but it is not a
-  valid octal number, as is the case with, for example, "<code>09</code>".
+ <li><p>If parsing <var>last</var> as an <a lt="IPv4 number parser">IPv4 number</a> does not
+ return failure, then return true.
+
+  <p class=note>This is equivalent to checking that <var>last</var> starts with "<code>0X</code>"
+  or "<code>0x</code>", followed only by <a>ASCII hex digits</a>.
 
  <li><p>Return false.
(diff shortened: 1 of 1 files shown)


karwa:
@annevk Do you think this is reasonable?

achristensen07:
FWIW, this is how I implemented it.  Calling the IPv4 parser is too slow.  I just checked for all digits and 0x or 0X and hex digits.

annevk:
Yeah this seems fine, modulo some formatting nits. I would like some way to preserve the 09 example, but perhaps the way to do that is to have an example table for hosts similar to what we have for URLs at https://url.spec.whatwg.org/#urls, at https://url.spec.whatwg.org/#hosts-(domains-and-ip-addresses).

url.bs
```diff
@@ -681,15 +681,14 @@ steps:
 
  <li><p>Let <var>last</var> be the last <a for=list>item</a> in <var>parts</var>.
 
- <li><p>If parsing <var>last</var> as an <a lt="IPv4 number parser">IPv4 number</a> does not
- return failure, then return true.
-
  <li>
   <p>If <var>last</var> is non-empty and contains only <a>ASCII digits</a>, then return true.
```
annevk, line 685: Since this no longer has a note, the `<p>` needs to be on the same line as `<li>`

url.bs
```diff
@@ -681,15 +681,14 @@ steps:
...
 
-  <p class=note>This can happen if <var>last</var> starts with "<code>0</code>" so the
-  <a lt="IPv4 number parser">IPv4 number parser</a> tries to parse it as octal, but it is not a
-  valid octal number, as is the case with, for example, "<code>09</code>".
+ <li><p>If parsing <var>last</var> as an <a lt="IPv4 number parser">IPv4 number</a> does not
```
annevk, line 687: And here it should be on its own line as the `<li>` has multiple children.

annevk:
> Looking at this again this basically means deferring when "09" is handled.

FWIW, I realized this is wrong as the former text does the same as the new text (the first step would get failure and then the second step would return true). Fortunately the new notes are correct despite this misunderstanding with regards to the implications.

PR #649, opened 2021-09-09
Editorial: use Infra substring operations

domenic:
Depends on https://github.com/whatwg/infra/pull/401

***
<a href="https://whatpr.org/url/649.html" title="Last updated on Sep 9, 2021, 10:28 PM UTC (0aa39b7)">Preview</a> | <a href="https://whatpr.org/url/649/7071c49...0aa39b7.html" title="Last updated on Sep 9, 2021, 10:28 PM UTC (0aa39b7)">Diff</a>

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -101,9 +101,10 @@ code point stream.
 <a>pointer</a> points to as long as it does not point nowhere. When the <a>pointer</a> points to
 nowhere <a>c</a> cannot be used.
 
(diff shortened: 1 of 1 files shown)


annevk (approved):
Should we point out that technically code unit substring also works for all these cases? I guess nobody implements this as-is anyway so it's probably fine.

domenic:
I think if I were to point that out I'd do it more generally, noting that past a certain point in the parser we're in the ASCII range and so code points vs. code units does not matter.

PR #655, opened 2021-10-07
Editorial: fold cannot-be-base-URL into path

### LINKED ISSUE #634: [Editorial] Replace the term 'cannot-be-a-base' with hierarchical/non-hierarchical
It is important that users understand the difference between hierarchical/non-hierarchical URLs, as many operations will fail if a URL is non-hierarchical.

However, the term currently used in the spec, 'cannot-be-a-base', is not particularly intuitive. Consider trying to document something like the `hostname` setter:

> This operation fails if the URL cannot be a base

> This operation is only valid if the URL's cannot-be-a-base flag is false

It doesn't read well, and it poses an obvious question: which URLs cannot be a base? Does this affect me? How can I know? Slashes after the scheme delimiter isn't really helpful at a semantic level - it describes how the parser decides whether a URL is hierarchical, but doesn't help users predict whether they should expect to deal with non-hierarchical URLs in their applications.

Or something like a `pathComponents` view:

> cannot-be-a-base URLs do not have path components

> Only base URLs have path components

It's all quite awkward and opaque, IMO.

However, if we rephrased these in terms of hierarchical/non-hierarchical URLs, it all becomes a lot clearer:

`hostname` setter:

> This operation fails if the URL is non-hierarchical

> Only hierarchical URLs may have hostnames

`pathComponents`:

> Non-hierarchical URLs do not have path components

> Only hierarchical URLs have path components

It also makes it a lot easier for libraries to expose the cannot-be-a-base property to users, and explain what it means: a hierarchical URL has an authority component or hierarchical path (i.e. a path beginning with a "/"), also known as a [`hier-part`](https://datatracker.ietf.org/doc/html/rfc3986#section-3) in RFC-3986 terminology. Suddenly it makes sense why the setters of authority components fail on a non-hierarchical URL.

Personally, I think it's better when implementations don't invent their own concepts and are able to refer directly to ideas described in the standard. So, since I think this is something implementations may want to expose to users, I'd like to incorporate this terminology in to the standard. We can keep the term 'cannot-be-a-base' around and simply define it as a synonym for a non-hierarchical URL.

What do people think? This will likely involve a bunch of churn without changes to functionality, so I'd like to get some feedback before working on a PR.

annevk:
Fetch PR: https://github.com/whatwg/fetch/pull/1337.

HTML PR: https://github.com/whatwg/html/pull/7240.

URLPattern PR: https://github.com/WICG/urlpattern/pull/143.

Fixes #634.

***
<a href="https://whatpr.org/url/655.html" title="Last updated on Oct 19, 2021, 10:07 AM UTC (ace32d5)">Preview</a> | <a href="https://whatpr.org/url/655/4edcc9f...ace32d5.html" title="Last updated on Oct 19, 2021, 10:07 AM UTC (ace32d5)">Diff</a>

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -2415,7 +2415,7 @@ these steps:
       </ol>
     </ol>
 
-   <dt><dfn id=cannot-be-a-base-url-path-state>opaque path state</dfn>
+   <dt><dfn export for="basic URL parser" id=cannot-be-a-base-url-path-state>opaque path state</dfn>
    <dd>
     <ol>
      <li><p>If <a>c</a> is U+003F (?), then set <var>url</var>'s <a for=url>query</a> to the empty


annevk:
Thanks Ben! This took a bit more effort than expected, but I think it does look rather nice and having the URL path serializer will also allow for some nice cleanup elsewhere, in particular in HTML. This will need to be validated in jsdom's whatwg-url (ideally by someone that's not me) and I'd like @karwa @TimothyGu @alwinb @domenic to be okay with it.

I'm not a big fan of "string path state". I considered renaming path in the other path-related states to hierarchical path, but I think it would be confusing if the lone path state ended up being the exceptional case.

alwinb:
I just spotted that the id attribute on the opaque path state is unchanged as "cannot-be-a-base-url-path-state". I don't know if that is intentional.

annevk:
Yeah that is intentional. I prefer keeping old IDs around so old links keep working. I also like that it shows the document has some history. We can of course add a new ID as well, but it's not always clear to me that is worth the additional bytes and nodes.

PR #662, opened 2021-10-20
Correct valid IPv4-address string

### LINKED ISSUE #583: Valid IPv4 address string
There is a minor issue with the definition of a **valid IPv4-address string** in the [Host Writing][1] section. 

> A **valid IPv4-address string** must be four sequences of up to three ASCII digits per sequence, each representing a decimal number no greater than 255, and separated from each other by U+002E (.).

From this is is not clear if e.g. `127.0.0.001` is a valid IPv4-address string.
The parsing section suggests that it is not; the [IPv4 number parser][2] records a validationError for e.g. `001` in Step 4. 

[1]: https://url.spec.whatwg.org/#valid-ipv4-address-string
[2]: https://url.spec.whatwg.org/#ipv4-number-parser

annevk:
Fixes #583.

cc @alwinb

***
<a href="https://whatpr.org/url/662.html" title="Last updated on Oct 20, 2021, 11:55 AM UTC (6ffebc5)">Preview</a> | <a href="https://whatpr.org/url/662/43c2713...6ffebc5.html" title="Last updated on Oct 20, 2021, 11:55 AM UTC (6ffebc5)">Diff</a>

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -638,9 +638,9 @@ decisions made, i.e. whether to use the <a>same site</a> or <a>schemelessly same
 <p>A <dfn export oldids=syntax-host-domain>valid domain string</dfn> must be a string that is a
 <a>valid domain</a>.
 
-<p>A <dfn export oldids=syntax-host-ipv4>valid IPv4-address string</dfn> must be four sequences of
-up to three <a>ASCII digits</a> per sequence, each representing a decimal number no greater than
-255, and separated from each other by U+002E (.).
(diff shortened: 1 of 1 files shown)


annevk:
A more ambitious take might be trying to share some infrastructure with HTML's valid non-negative integer...

annevk:
Right, we would still need to constrain it with "shortest possible", but that's also used within HTML to constrain things. Perhaps there is some future point where Infra defines more of this and we could refactor this.

Issue #627, opened 2021-08-10
Forbidden host code points should include all C0 controls & U+007F


TimothyGu:
Right now, [forbidden host code points](https://url.spec.whatwg.org/#forbidden-host-code-point) only includes U+0000 NULL and ASCII whitespaces from the C0 controls range. However, there seems to be no good reason to allow the other unprintable characters. Right now according to spec, `new URL('https://a\u0002b.com')` parses with `u.hostname === 'a\u0002b.com'`, which deviates from what a user might expect from a URL parser.

Most implementations agree with this assessment:

* Chrome and Go forbid C0 controls and `\u007F`, whether encoded or not
* Firefox forbids C0 controls verbatim, as well as `%00`–`%1F` and `%7F`. It allows `\u007F`.
* curl forbids C0 controls and `\u007F`, as well as `%00`–`%1F`. It allows `%7F`.
  * This is the dual of Firefox's behavior.
* Safari follows current spec.
* Python and Ruby accepts any percent-encoded characters in hosts
* Node.js legacy parser breaks when it sees any control character (percent-encoded or not) in hosts (it tries to add a `/` before the character or `%`)

I propose Chrome's behavior as it seems more consistent than Firefox or curl.

valenting:
> Interestingly, in Firefox `new URL('https://a\u0007f.com')` 

That's because that is equivalent to:
`new URL("https://a" + String.fromCodePoint(0x07) + "f.com")` 🙂

achristensen07:
🙂
I implemented this with tests at https://bugs.webkit.org/show_bug.cgi?id=232034

TimothyGu:
@achristensen07 `new URL('https://a\u0007f.com')` is `a` U+0007 `f`, while `'a' + String.fromCodePoint(0x7F)` is `a` U+007F.

I guess I was wrong in my initial post (which I've just updated): Firefox forbids `%00`–`%1F`, `%7F`, and `\u0000`–`\u001F`. It does not forbid `\u007F`. Chrome forbids all of these.

I propose that we follow Chrome here as it's more consistent.

annevk:
As pointed out by @achristensen07 in the WebKit bug one problem here might be non-special URLs.

I suspect we might not want to have things result in failure there as that would end up with things throwing that currently do not throw in either Chrome or Firefox.

That would be a further argument to fix #319 and probably have two separate constants. Thoughts? Anyone up for fixing this in the spec?

alwinb:
[...]

PR #673, opened 2021-11-22
Disallow all C0 control and DEL codepoints in hosts

alwinb:
A first pass at fixing #627 and #319

- [X] At least two implementers are interested (and none opposed):
   * Safari
   * Chrome already implement this behaviour
   * …
   * Chrome: …
   * Firefox: …
   * Safari: …

* * *
This adds the C0 controls and delete to the forbidden host codepoints, as discussed in #627.
It also adapts the opaque-host-parser to percent-encode rather than fail on these code points. 

For this I added an **opaque-host percent encode set** that refers to the forbidden host codepoints. 
Please review.

***
<a href="https://whatpr.org/url/673.html" title="Last updated on Nov 22, 2021, 1:03 PM UTC (993aa2a)">Preview</a> | <a href="https://whatpr.org/url/673/f787850...993aa2a.html" title="Last updated on Nov 22, 2021, 1:03 PM UTC (993aa2a)">Diff</a>

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -196,6 +196,9 @@ U+0027 (').
 <p>The <dfn oldids=default-encode-set>path percent-encode set</dfn> is the
 <a>query percent-encode set</a> and U+003F (?), U+0060 (`), U+007B ({), and U+007D (}).
 
+<p>The <dfn>opaque-host percent-encode set</dfn> is the <a>C0 control percent-encode set</a> and
+the <a>forbidden host code points</a>.
+
 <p>The <dfn oldids=userinfo-encode-set>userinfo percent-encode set</dfn> is the
 <a>path percent-encode set</a> and U+002F (/), U+003A (:), U+003B (;), U+003D (=), U+0040 (@),
 U+005B ([) to U+005E (^), inclusive, and U+007C (|).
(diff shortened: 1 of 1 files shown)


annevk (commented):
I would prefer not adding percent-encoding of additional code points to opaque hosts at the same time. I.e., restrict this change to additional failure cases of domains.

What I was envisioning is that we have two constants:

* Forbidden host code points (this can be renamed to forbidden domain code points). We would modify this as you did.
* Forbidden opaque-host code points (this would be the old forbidden host code points).

karwa:
One thing I'd like to add (before I'm too full of "festive cheer" and forget it):

Failing on certain forbidden host code-points can unlock optimisation opportunities. For example, currently when parsing the hostname portion of a URL string, we have to precisely track whether we're inside or outside of a square bracket (in order to determine whether a colon is part of an IPv6 address or is a hostname/port delimiter). This slows down all URL strings. I've been experimenting with an approach that simply splits on the first colon after a closing square bracket.

So for the hostname `"][::][:80"`, the split is:
| Method | Host | Port
| ---- | --- | --- |
| Std | `"][::]["` | `"80"` |
| Mine | `"]["` | `":][:80"` |

Currently that's fine (I think?). Since the hostname doesn't begin with an opening bracket, it won't be parsed as IPv6, and ultimately the opaque host parser will reject them both because square brackets are forbidden. If we changed to percent-encode forbidden code-points, this input would potentially be accepted.

Perhaps this could be limited to the setter function?

alwinb:
Ok. So it makes sense to investigate this more thoroughly; how percent coding opaque hosts could/ should interact with the ipv6 address syntax, and make it a separate issue (as was suggested). 

Maybe we can take care of that first? I’m fine with handling this one first, but I myself just can’t really get behind having two distinct forbidden host codepoints sets… it may just be a minor thing, I realise. But still…

Is failing on all controls in opaque hosts really not an option? Maybe there are there solutions?

annevk:
I expect we'll take #685 at which point we should probably close this.

alwinb:
Of course. I’m happy @karwa picked this up.

karwa:
Sorry, this was just blocking some interop stuff that I was working on, and since there seemed to be agreement on the idea of blocking C0s and delete, I figured I'd just get it done and take the opportunity to make the tests more comprehensive.

alwinb:
Oh no it’s excellent. I felt like I was blocking the progress here, and now this issue is taken care of and the opaque hosts can maybe be looked at later. I’ll close the PR.

Issue #531, opened 2020-07-17
Support relative URLs


sholladay:
The `new URL()` constructor currently requires an absolute URL as input, or an absolute base URL, which is painful when there is no obvious base URL to use, such as in the context of a command line app or other non-browser environment.

In many cases, it's useful to parse and resolve relative URLs, such as URL paths or origin-relative URLs, without an absolute base URL.

```js
new URL('/to', '/from');
new URL('/to', '//from.com/');
```

The lack of support for this use case is causing me to remove WHATWG URL from Ky, a popular HTTP request library, in favor of our own string replacement. See: https://github.com/sindresorhus/ky/pull/271

annevk:
Well, its purpose is to create a URL and those are by definition not relative. I could see wanting something specialized for path/query/fragment manipulation though. Are there any popular libraries that handle that we could draw inspiration from?

sholladay:
Where is it defined that a URL must contain a scheme and a host in order to be a valid URL?

Even if such a definition exists, `new URL()` is the first API in the web ecosystem that I have encountered that has this limitation, making it quite surprising.

Beyond that, the WHATWG URL spec itself defines relative URLs...

https://url.spec.whatwg.org/#relative-url-string

As for existing implementations, see Node's `url.parse()` and `url.resolve()`, among others. I've used these extensively to manipulate URLs where the scheme and/or host is not known ahead of time and will be determined later by the end-user or browser, depending on where the URL is ultimately used.

annevk:
It defines them as input (though only in the context of a base URL, which at least browsers always use), it doesn't define them as data structures. The data structure is defined at https://url.spec.whatwg.org/#url-representation (though it's fair to say that does make it seem like more is optional than in reality is optional; something to improve).

sholladay:
I get that browsers need an absolute base URL to actually perform a request. And thus it makes sense for the URL specification to define what an absolute base URL is and discuss resolving relative URLs in the context of an absolute base URL, etc.

What doesn't make sense to me is why `new URL()` imposes this limitation. I cannot think of anything else on the web platform that does this. Even HTML's `<base>` tag supports relative URLs, despite the fact that it is specifically meant for defining the base URL.

annevk:
[...]

Issue #479, opened 2020-04-24
Provide a succinct grammar for valid URL strings


alercah:
As an occasional standards-user, the lack of a succinct expression of the grammar for valid URL strings is rather frustrating. It makes it rather difficult to follow what's going on and, in particular, to work out whether a given thing is a valid URL. A grammar in EBNF or a similar form would be greatly appreciated and make this spec significantly easier to understand.

domenic:
Just to be sure, you're saying you prefer something like

```
url-query-string = url-unit{0,}
```

(or whatever grammar format) to the spec's current

> A URL-query string must be zero or more URL units. 

?

masinter:
I think this is #24 #416

alercah:
Apologies for the slow reply.

I think that @masinter is right, and that my concern matches the ones discussed there. I skimmed those threads and a few statements concerned me, such as the assertion that a full Turing machine is required to parse URLs: this would very much surprise me; my instinct on reading the grammar is that, once you separate out the different paths for `file`, special schemes, and non-special schemes in relative URLs, the result is almost certainly context-free. It might even be regular. The fact that the given algorithm mostly does a single pass is a strong indication that complex parsing is not required.

I discussed with a colleague of mine, @tabatkins, and they said that the CSS syntax parser was much improved when it was rewritten from a state-machine based parser into a recursive descent parser. Doing this would effectively require writing the URL grammar out as a context-free grammar, which would make providing a BNF-like specification, even if it's only informative, very easy.

Separately, but not entirely, splitting out the parsing from the semantic functions (checking some validity rules, creating the result URL when parsing a relative URL string) would likely improve the readability of the spec and the simplicity of implementing it. I think this might be better suited for a separate thread, though, as there are some other thoughts I have in this vein as well.

alwinb:
This might be a more complicated problem than you think. I have tried several times, but the scheme dependent behaviour causes a lot of duplicate rules, so you end up with a grammar that is not very concise nor easy to read. And there is a tricky problem with repeated slashes before the host, the handling of which is base URL dependent. 

I have some notes on it [here][1]. (I eventually went with a hybrid approach of a couple of very simple grammars and some logic rules in between). This ties into a model of URLs that I describe [here][2]. 

What's the status of this? It really does work. I developed the theory when I tried to write a library that supports relative URLs. I am quite confident that it matches the standard (but not everything is described in the notes); as the library now passes all of the parsing tests. 

[...]

PR #678, opened 2021-12-22
[Editorial] Correct the note about the IDNA fast-path for ASCII domains

karwa:
The full IDNA algorithm lowercases before checking the `"xn--"` prefix, so any implementors fast-pathing ASCII domains should make sure that they check for it case-insensitively.

Added tests:
   * https://github.com/web-platform-tests/wpt/pull/32177

***
<a href="https://whatpr.org/url/678.html" title="Last updated on Jan 14, 2022, 12:48 PM UTC (4f43025)">Preview</a> | <a href="https://whatpr.org/url/678/f787850...4f43025.html" title="Last updated on Jan 14, 2022, 12:48 PM UTC (4f43025)">Diff</a>

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -583,8 +583,8 @@ decisions made, i.e. whether to use the <a>same site</a> or <a>schemelessly same
 
   <p class=note>If <var>beStrict</var> is false, <var>domain</var> is an <a>ASCII string</a>, and
   <a>strictly splitting</a> <var>domain</var> on U+002E (.) does not produce any
-  <a for=list>item</a> that <a for=string>starts with</a> "<code>xn--</code>", this step is
-  equivalent to <a>ASCII lowercasing</a> <var>domain</var>.
+  <a for=list>item</a> that <a for=string>starts with</a> an <a>ASCII case-insensitive</a> match for
+  "<code>xn--</code>", this step is equivalent to <a>ASCII lowercasing</a> <var>domain</var>.
 
  <li><p>If <var>result</var> is a failure value, <a>validation error</a>, return failure.
 


TimothyGu (approved):
Ah, great catch! Shows how hard it is to get this stuff right.

url.bs
```diff
@@ -583,7 +583,7 @@ decisions made, i.e. whether to use the <a>same site</a> or <a>schemelessly same
 
   <p class=note>If <var>beStrict</var> is false, <var>domain</var> is an <a>ASCII string</a>, and
   <a>strictly splitting</a> <var>domain</var> on U+002E (.) does not produce any
-  <a for=list>item</a> that <a for=string>starts with</a> "<code>xn--</code>", this step is
+  <a for=list>item</a> that <a for=string>starts with</a> an <a>ASCII case-insensitive</a> match for "<code>xn--</code>", this step is
   equivalent to <a>ASCII lowercasing</a> <var>domain</var>.
```
annevk, line 587: Needs a wrapping fix:
```suggestion
  <a for=list>item</a> that <a for=string>starts with</a> an <a>ASCII case-insensitive</a> match for 
  "<code>xn--</code>", this step is equivalent to <a>ASCII lowercasing</a> <var>domain</var>.
```
Although this is still not entirely technically correct given the signature of "starts with", I think it's good enough?

karwa: @annevk  Yeah the "starts with" thing is a bit awkward, but I couldn't find anything better in infra, nor could I think of any way it could be ambiguous as written, so... 🤷‍♂️ .

annevk (approved):

annevk:
Thanks @karwa! Unfortunately our tooling does not help with formatting. I think it's something we ought to tackle, but it's not really my area of expertise so I haven't looked into it myself as of yet.

masstez (commented):
Submit general feedback without explicit approval

asaneh:
#678 #479 #687 #531 #686

PR #685, opened 2022-01-25
Split forbidden host/domain code points, and add all C0 controls and add U+007F to the latter

karwa:
Attempt at resolving #627 and #319

   * Chrome (Presumably - See linked issue)
   * WebKit (Presumably - See linked issue)
   * https://github.com/web-platform-tests/wpt/pull/32524
   * Chrome: https://bugs.chromium.org/p/chromium/issues/detail?id=1291564
   * Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1752320
   * Safari: https://bugs.webkit.org/show_bug.cgi?id=235709
   * Deno: https://github.com/denoland/deno/issues/13509
   * Node.js: https://github.com/nodejs/node/issues/41717

***
<a href="https://whatpr.org/url/685.html" title="Last updated on Jan 28, 2022, 8:59 AM UTC (3229929)">Preview</a> | <a href="https://whatpr.org/url/685/68dc61c...3229929.html" title="Last updated on Jan 28, 2022, 8:59 AM UTC (3229929)">Diff</a>

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -471,11 +471,11 @@ processing.
 <h3 id=host-miscellaneous>Host miscellaneous</h3>
 
 <p>A <dfn export>forbidden host code point</dfn> is U+0000 NULL, U+0009 TAB, U+000A LF, U+000D CR,
-U+0020 SPACE, U+0023 (#), U+0025 (%), U+002F (/), U+003A (:), U+003C (&lt;), U+003E (>), U+003F (?),
-U+0040 (@), U+005B ([), U+005C (\), U+005D (]), U+005E (^), or U+007C (|).
+U+0020 SPACE, U+0023 (#), U+002F (/), U+003A (:), U+003C (&lt;), U+003E (>), U+003F (?), U+0040 (@),
+U+005B ([), U+005C (\), U+005D (]), U+005E (^), or U+007C (|).
 
 <p>A <dfn export>forbidden domain code point</dfn> is a <a>forbidden host code point</a>,
-a <a>C0 control</a>, or U+007F DELETE.
+a <a>C0 control</a>, U+0025 (%), or U+007F DELETE.
 
 <p>A <a for=/>host</a>'s <dfn for=host export>public suffix</dfn> is the portion of a
 <a for=/>host</a> which is included on the <cite>Public Suffix List</cite>. To obtain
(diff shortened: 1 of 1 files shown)


annevk (approved):
This looks good to me. I don't think the conformance requirements end up needing adjusting (and it seems like this fixes a small problem for "valid opaque-host string"), although for "valid domain string" it's a bit unclear due to the imperative nature of its definition.

I'd appreciate a second review.

domenic:
> Chrome (Presumably - See linked issue)

Which part of which linked issue are you referring to?

domenic:
Hmm. But https://wpt.fyi/results/url?diff&filter=ADC&run_id=5656521351364608&run_id=5648068402741248 shows many new failing tests in Chrome. Are all of those tests that would already fail with the current spec? Maybe it'd be best to separate out the tests PR into one that tests the current spec, and one that tests the change here, so that we can actually see if the change here causes any deltas from the current spec?

karwa:
There are more failures because there are more tests, but for the part which is changed by this spec change - domains containing (raw or encoded) code points in the range `\u0000-\u0001F` or `\u007F` - Chrome passes them all.

If I changed the tests to the current spec, there would be even more failures for _all browsers_. None of them implement the current spec. WebKit was the last one, but @achristensen07 recently changed made the change to reject those code-points (for all hosts, not just domains, which is why WebKit still has one failure, but the consensus from the issue seemed to be to leave those alone for now):

> I think there is a consensus that the forbidden host code points should include all C0 controls and U+007F, and that now matches the behavior of all browsers. Let's make that change to the specification and then make a separate issue about what we want to do with opaque hosts, where there is less consensus and less spec compliance.

https://github.com/whatwg/url/issues/627#issuecomment-981906774

domenic:
> If I changed the tests to the current spec, there would be even more failures for all browsers

Right. That would be helpful. Then we could double-check that the new tests, specific to the changes in this PR, are all passing in Chromium.

karwa:
OK, I've put up https://github.com/web-platform-tests/wpt/pull/32539

Same tests as linked for this change, but with expected values that match the current spec.

domenic:
Editorially LGTM, let's just rebase https://github.com/web-platform-tests/wpt/pull/32524 on WPT master, confirm that the result is more in alignment with browsers, and then we should be good to merge. Thanks @karwa for doing the work, including the testing against whatwg-url.

annevk:
@karwa do you want to file the implementation bugs? https://github.com/whatwg/meta/blob/master/MAINTAINERS.md has pointers.

karwa:
@annevk bugs have been filed.

annevk:
Thanks a lot @karwa for your work on tackling this as well as providing all those great tests!

PR #691, opened 2022-04-25
Add more examples of opaque hosts

### LINKED ISSUE #690: Opaque hosts: realistic examples and origins?
What are some realistic examples of URLs that use [opaque hosts](https://url.spec.whatwg.org/#opaque-host)? I.e., nonempty strings that are not domains or IP addresses, but appear in the host position. The spec doesn't seem to contain any. An artificial example might be something like `my-scheme://opaque-string/path`, but I'm unaware of what kind of protocols might use this in the wild. If any exist, it'd be good to document them in the spec!

We thought they might be useful for cases where we want to define a new kind of resource, where relative URL resolution works. E.g., `isolated-app://app-id/path`, where `app-id` is an opaque host. However, when we went down this route, we discovered a second problem, which is that [origin computation](https://url.spec.whatwg.org/#concept-url-origin) works poorly for such hosts: they always result in an opaque origin. It would be better if it returned something of the form ("isolated-app", "app-id", null, null), so that such "isolated-app" cases were cross-origin from each other, and cross-origin from all HTTP(S) origins, but had same-origin relationships within each other.

This issue in general feels like it falls into the gray area where the URL Standard is trying to provide a general framework, but is primarily focused around the needs of the web and the rest of the web spec ecosystem. That is, nothing in any web specs seems to use opaque-host URLs, so they are getting mixed support in the URL Standard. One resolution would be to say that anything which wants to use opaque-host URLs will need to use its own non-URL-Standard origin computation, but that is a bit subpar, as for example it would lead to `(new URL("isolated-app://app-id/")).origin` differing in value between environments. I guess it's a question about whether origins are a URL concept or a HTML/Fetch concept, which has always been a bit fuzzy.

domenic:
Closes #690.

***
<a href="https://whatpr.org/url/691.html" title="Last updated on Apr 25, 2022, 8:19 PM UTC (52d7702)">Preview</a> | <a href="https://whatpr.org/url/691/9a83e25...52d7702.html" title="Last updated on Apr 25, 2022, 8:19 PM UTC (52d7702)">Diff</a>

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -386,46 +386,58 @@ point <a for=/>URLs</a> from <var>A</var> can come from untrusted sources.
 </ul>
 
 <div class=example id=example-host-parsing>
+ <p>A <a lt="host parser">parse</a>-<a lt="host serializer">serialize</a> roundtrip gives the
+ following results, depending on the <var>isNotSpecial</var> argument to the <a>host parser</a>:
+
  <table>
   <tr>
    <th>Input
-   <th>Output
+   <th>Output (<var ignore>isNotSpecial</var> = false)
+   <th>Output (<var ignore>isNotSpecial</var> = true)
   <tr>
    <td><code>EXAMPLE.COM</code>
    <td rowspan=2><code>example.com</code> (<a for=/>domain</a>)
+   <td><code>EXAMPLE.COM</code> (<a>opaque host</a>)
   <tr>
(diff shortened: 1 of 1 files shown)


annevk (approved):

annevk:
Thanks!

PR #717, opened 2022-11-22
Editorial: adopt Infra for URLSearchParams

annevk:
This could be taken even further, but this seems like a good first step.

***
<a href="https://whatpr.org/url/717.html" title="Last updated on Dec 5, 2022, 9:09 AM UTC (c9c4c72)">Preview</a> | <a href="https://whatpr.org/url/717/48f1e79...c9c4c72.html" title="Last updated on Dec 5, 2022, 9:09 AM UTC (c9c4c72)">Diff</a>

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -3341,14 +3341,15 @@ console.log(url.searchParams.get('b')); // "~"</code></pre>
 
 <ol>
  <li>
-  <p>If <var>init</var> is a <a>sequence</a>, then <a for=list>for each</a> <var>tuple</var> of
-  <var>init</var>:
+  <p>If <var>init</var> is a <a>sequence</a>, then <a for=list>for each</a> <var>innerSequence</var>
+  of <var>init</var>:
 
   <ol>
-   <li><p>If <var>tuple</var>'s <a for=list>size</a> is not 2, then <a>throw</a> a {{TypeError}}.
+   <li><p>If <var>innerSequence</var>'s <a for=list>size</a> is not 2, then <a>throw</a> a
+   {{TypeError}}.
 
-   <li><p><a for=list>Append</a> (<var>tuple</var>[0], <var>tuple</var>[1]) to <var>query</var>'s
-   <a for=URLSearchParams>list</a>.
+   <li><p><a for=list>Append</a> (<var>innerSequence</var>[0], <var>innerSequence</var>[1]) to
+   <var>query</var>'s <a for=URLSearchParams>list</a>.
   </ol>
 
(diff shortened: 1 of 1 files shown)


url.bs
```diff
@@ -2853,7 +2853,7 @@ handled with care to prevent spoofing:
 <h2 id="application/x-www-form-urlencoded"><code>application/x-www-form-urlencoded</code></h2>
 
 <p>The <dfn export id=concept-urlencoded><code>application/x-www-form-urlencoded</code></dfn> format
-provides a way to encode name-value pairs.
+provides a way to encode name-value <a for=/>tuples</a>.
```
andreubotella, line 2856: Since this is a general description, "name-value pair" works, if we take "pair" as the common term rather than the now-removed infra definition. Otherwise, it should probably be "tuples consisting of a name and a value" as below.

url.bs
```diff
@@ -3345,20 +3345,19 @@ console.log(url.searchParams.get('b')); // "~"</code></pre>
 
 <ol>
  <li>
-  <p>If <var>init</var> is a <a>sequence</a>, then <a for=list>for each</a> <var>pair</var> in
+  <p>If <var>init</var> is a <a>sequence</a>, then <a for=list>for each</a> <var>tuple</var> of
```
andreubotella, line 3348: The variable name could still be "pair".

url.bs
```diff
@@ -3411,42 +3410,42 @@ method steps are:
 <p>The <dfn method for=URLSearchParams><code>delete(<var>name</var>)</code></dfn> method steps are:
 
 <ol>
- <li><p>Remove all name-value pairs whose name is <var>name</var> from
- <a for=URLSearchParams>list</a>.
+ <li><p><a for=list>Remove</a> all <a for=/>tuples</a> whose name is <var>name</var> from
```
andreubotella, line 3413: "whose first element"? And similarly below.

annevk: I was trying to strike a balance between going full `<dfn>` and being rather opaque.

url.bs
annevk: While true, I think I prefer sticking to a single term throughout. I think that makes it easier for non-native speakers.

domenic: Honestly I think _subSequence_ or _innerSequence_ would be a better name. It's confusing to use the variable name _tuple_ for something that's actually an Infra list/IDL sequence.

For example, at first I wrote a comment below "Why not just append _tuple_ directly, instead of appending (_tuple_[0], _tuple_[1])?" But now I see that would be a type error. I was misled by the variable name.

domenic (approved):
LGTM modulo the variable naming discussion

PR #722, opened 2022-12-09
Clarify we depend on UTS46 and not IDNA2008

### LINKED ISSUE #498: Explain that we are using UTS #46 for IDNA support rather than IDNA2008 directly
It seems like quite a few people have been thinking that we use IDNA2008 or 2003 directly, but we actually do not. The motivation for using UTS 46 rather than either IDNA versions is pretty clearly stated in the document itself:

> A system was introduced in 2003 for internationalized domain names (IDN). This system is called Internationalizing Domain Names for Applications, or IDNA2003 for short. … A revision of IDNA was approved in 2010 (IDNA2008). This revision has a number of incompatibilities with IDNA2003.
>
> The incompatibilities force implementers of client software, such as browsers and emailers, to face difficult choices during the transition period as registries shift from IDNA2003 to IDNA2008. This document specifies a mechanism that minimizes the impact of this transition for client software, allowing client software to access domains that are valid under either system.

So we probably don't need to explain why UTS 46 is used. However, it might be good to make it clear in the spec that we _are_ using UTS 46, perhaps as an introduction paragraph in https://url.spec.whatwg.org/#idna.

annevk:
Closes #498.

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -621,7 +621,7 @@ concepts.
   with <i>domain_name</i> set to <var>domain</var>, <i>UseSTD3ASCIIRules</i> set to
   <var>beStrict</var>, <i>CheckHyphens</i> set to false, <i>CheckBidi</i> set to true,
   <i>CheckJoiners</i> set to true, <i>Transitional_Processing</i> set to false,
-  and <i>VerifyDnsLength</i> set to <var>beStrict</var>.
+  and <i>VerifyDnsLength</i> set to <var>beStrict</var>. [[!UTS46]]
 
   <p class=note>If <var>beStrict</var> is false, <var>domain</var> is an <a>ASCII string</a>, and
   <a>strictly splitting</a> <var>domain</var> on U+002E (.) does not produce any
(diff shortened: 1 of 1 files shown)


url.bs
```diff
@@ -635,6 +635,10 @@ concepts.
  <li><p>Return <var>result</var>.
 </ol>
 
+<p class=note>This document and the web platform at large use
+<cite>Unicode IDNA Compatibility Processing</cite> and not <cite>IDNA2008</cite>. For instance,
```
domenic, line 639: It looks like IDNA2008 is not a document name, but instead an informal term for a collection of documents? The relevant document name is instead "[Internationalized Domain Names for Applications (IDNA): Definitions and Document Framework](https://www.rfc-editor.org/rfc/rfc5890)".

domenic (approved):
LGTM with cite usage nit

url.bs
annevk: So I want to mention IDNA2008. I guess I could just not use `<cite>`. Would I have to quote it?

domenic: It seems fine to just not use `<cite>`: `and not IDNA2008`.

PR #725, opened 2022-12-12
Editorial: clarify constructor example

### LINKED ISSUE #724: "This throws an exception if the input is not an..."
https://url.spec.whatwg.org/commit-snapshots/bed0538a76ccbf383a63e724d169e18cfa725866/#example-5434421b

> This throws an exception if the input is not an absolute\-URL\-with\-fragment string:​ 

This isn't correct. For example, `https:example.org` is not an absolute-URL-with-frament-string, but `new URL("https:example.org")` does not throw an exception.

One potential fix would be not to try to reference a formally-defined term here, and instead say something vaguer and not linked like "absolute URL". Or you could try to say something about "parseable as a URL without a base URL", maybe.

annevk:
Closes #724.

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -3041,7 +3041,7 @@ var input = "https://example.org/💩",
     url = new URL(input)
 url.pathname // "/%F0%9F%92%A9"</code></pre>
 
- <p>This throws an exception if the input is not an <a>absolute-URL-with-fragment string</a>:
+ <p>This throws an exception if the input is a <a>relative-URL string</a>:
 
  <pre><code class="lang-javascript">
 try {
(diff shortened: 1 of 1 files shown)


url.bs
```diff
@@ -3050,15 +3050,15 @@ try {
...
+ <p>For those cases a <a>base URL</a> is necessary:
 
  <pre><code class="lang-javascript">
 var input = "/🍣🍺",
     url = new URL(input, document.baseURI)
 url.href // "https://url.spec.whatwg.org/%F0%9F%8D%A3%F0%9F%8D%BA"</code></pre>
 
- <p>A {{URL}} object can be used as <a>base URL</a> (while IDL requires a string as argument, a
- {{URL}} object stringifies to its {{URL/href}} getter return value):</p>
+ <p>A {{URL}} object can be used as a <a>base URL</a> (while <cite>Web IDL</cite> requires a string
```
domenic, line 3060: I might say "the IDL requires" instead of stating that the _Web IDL_ spec has something to say about this specific constructor.

domenic (approved):

PR #726, opened 2022-12-16
Editorial: clarify the UTF-8ness of percent-encoded bytes

### LINKED ISSUE #501: Percent-encoded bytes conformance requirement
@domenic mentioned in #500 that

> Sequences of percent-encoded bytes, string percent decoded, should not cause UTF-8 decode without BOM or fail to return failure.

Is confusing. What it is mainly trying to say is that, e.g., "`%FF`"  should not be used as 0xFF would end up as U+FFFD when UTF-8 decoded (or return failure for the above cited algorithm). There might be edge cases however where encoding non-string data in URLs is fine.

But perhaps we should simply stay out of this?

annevk:
Closes #501.

***
<a href="https://whatpr.org/url/726.html" title="Last updated on Dec 20, 2022, 7:59 AM UTC (3856b1f)">Preview</a> | <a href="https://whatpr.org/url/726/a05ee27...3856b1f.html" title="Last updated on Dec 20, 2022, 7:59 AM UTC (3856b1f)">Diff</a>

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -119,9 +119,13 @@ error.
 <h3 id=percent-encoded-bytes>Percent-encoded bytes</h3>
 
 <p>A <dfn>percent-encoded byte</dfn> is U+0025 (%), followed by two <a>ASCII hex digits</a>.
-Sequences of <a lt="percent-encoded byte">percent-encoded bytes</a>,
-<a for=string>percent-decoded</a>, should not cause <a>UTF-8 decode without BOM or fail</a> to
-return failure.
+
+<p class=note>It is generally a good idea for sequences of <a>percent-encoded bytes</a> to be such
+that, when <a for=string>percent-decoded</a> and then passed to
+<a>UTF-8 decode without BOM or fail</a>, they do not end up as failure. How important this is
+depends on where the <a>percent-encoded bytes</a> are used. E.g., for the <a>host parser</a> not
+following this advice is fatal, whereas for <a href="#url-rendering-i18n">URL rendering</a> the
+<a>percent-encoded bytes</a> would not be rendered <a for=string>percent-decoded</a>.
 
 <div algorithm>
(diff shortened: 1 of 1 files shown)


url.bs
```diff
@@ -119,9 +119,12 @@ error.
 <h3 id=percent-encoded-bytes>Percent-encoded bytes</h3>
 
 <p>A <dfn>percent-encoded byte</dfn> is U+0025 (%), followed by two <a>ASCII hex digits</a>.
-Sequences of <a lt="percent-encoded byte">percent-encoded bytes</a>,
-<a for=string>percent-decoded</a>, should not cause <a>UTF-8 decode without BOM or fail</a> to
-return failure.
+
+<p class=note>It is generally a good idea for sequences of <a>percent-encoded bytes</a>, when
```
domenic, line 123: Maybe "It is generally a good idea for sequences of percent-encoded bytes **to be such that**, when ... they do not end up as failure"?

My first reading confused me into thinking this was stating something about the algorithms, like, it's generally a good idea for the algorithms to be constructed such that any input sequence of bytes doesn't fail. But upon re-reading #501 I see the intent now.

url.bs
```diff
@@ -119,9 +119,12 @@ error.
...
+<a for=string>percent-decoded</a> and then passed to <a>UTF-8 decode without BOM or fail</a>, to not
+end up as failure. How important this is depends on where <a>percent-encoded bytes</a> are used,
+e.g., for the <a>host parser</a> it is fatal, whereas for
+<a href="#url-rendering-i18n">URL rendering</a> it is merely beneficial.
```
domenic, line 127: Maybe "for the host parser such sequences will cause failure, whereas for URL rendering [fill in the consequences or lack thereof here]"

domenic (approved):

PR #727, opened 2022-12-16
Clarify that percent-encoding often does not roundtrip

### LINKED ISSUE #523: Percent encoding does not round-trip
Hi!

I'm trying to implement the current version of the spec in Swift (for non-browser applications. The idea is that it's useful to have a URL type that behaves as your browser behaves and accepts/rejects the same things).

One issue that I've noticed is that the current definition of the ["UTF8 percent encode"](https://url.spec.whatwg.org/#percent-encoded-bytes) algorithm doesn't round-trip for strings which contain the percent character. For example, following the algorithm, the string "%100" (UTF8: [37, 49, 48, 48]) is not changed at all when encoding (regardless or the percent-encoding set; none of them contain the "%" character itself). However, decoding that same string using the decoding algorithm in the spec results in the UTF8 sequence [16, 48], or "0" (ASCII 0x10 is the unprintable "data link escape" character).

[RFC-3986](https://tools.ietf.org/html/rfc3986#section-2.4) warns about this:

> Because the percent ("%") character serves as the indicator for
   percent-encoded octets, it must be percent-encoded as "%25" for that
   octet to be used as data within a URI. 

Indeed, the `encodeURIComponent` JS function encodes the percent character:
```
> encodeURIComponent("%100")
"%25100"
```

I was ready to submit a PR to have the spec's algorithm also do this, but it appears there is an explicit test that percent characters are _not_ escaped (in this case, in the URL's username component): https://github.com/web-platform-tests/wpt/blob/master/url/resources/urltestdata.json#L2372

```
    "input": "http://%25DOMAIN:foobar@foodomain.com/",
    "base": "about:blank",
    "href": "http://%25DOMAIN:foobar@foodomain.com/",
    "origin": "http://foodomain.com",
    "protocol": "http:",
    "username": "%25DOMAIN",   // <--- I would expect this to be "%2525DOMAIN"
```
I'm not sure if this is correct. Usernames are required to be escaped (meaning they must be unescaped to recover their original value), as in the following example:

```
> new URL("http://;DOMAIN:foobar@foodomain.com/")
...
username: "%3BDOMAIN" // <--- Semicolon is escaped as %3B
```
However, unescaping the string "%25DOMAIN" as the test expects would not recover the original and result in a different username - "%DOMAIN".

Can anybody confirm the correct behaviour? I think we _should_ be escaping the % character, and that the test is incorrect. If that's not the case (and the current behaviour is correct), this could use a note or warning in the spec.

annevk:
Also add some `<div algorithm>` wrappers to this section.

Closes #523.

---

cc @karwa

***
<a href="https://whatpr.org/url/727.html" title="Last updated on Dec 19, 2022, 3:57 PM UTC (11c6d80)">Preview</a> | <a href="https://whatpr.org/url/727/50f0b09...11c6d80.html" title="Last updated on Dec 19, 2022, 3:57 PM UTC (11c6d80)">Diff</a>

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -283,8 +283,12 @@ U+005F (_).
 
 <p class=note>Of the possible values for the <var>percentEncodeSet</var> argument only two end up
 encoding U+0025 (%) and thus give “roundtripable data”: <a>component percent-encode set</a> and
-<a><code>application/x-www-form-urlencoded</code> percent-encode set</a>. Unfortunately the
-<a>URL parser</a> uses neither.
+<a><code>application/x-www-form-urlencoded</code> percent-encode set</a>. The other values for the
+<var>percentEncodeSet</var> argument — which happen to be used by the <a>URL parser</a> — leave
+U+0025 (%) untouched and as such it needs to be
+<a for="code point" lt="UTF-8 percent-encode">percent-encoded</a> first in order to be properly
+represented.
+
 </div>
 
 <div algorithm>
@@ -1328,6 +1332,16 @@ unified model would be, please file an issue.
    <td>
    <td>❌
    <td><code>https://example.com/[]?[]#[]</code>
+  <tr>
+   <td><code>https://example/%?%#%</code>
+   <td>
+   <td>❌
+   <td><code>https://example/%?%#%</code>
+  <tr>
+   <td><code>https://example/%25?%25#%25</code>
+   <td>
+   <td>✅
+   <td><code>https://example/%25?%25#%25</code>
  </table>
 
  <p>The base and output <a lt="URL record">URL</a> are represented in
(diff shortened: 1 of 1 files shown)


domenic (commented):
I'm not sure I understand the connection to the original issue. (Probably, I don't understand the original issue.)

The implicit statement here seems to be that because of the URL parser using neither percent-encode set, there are strings where the URL serialization of the URL parsing of "`x`"` is not "`x`". If that's true, adding an example would be great.

But #523 seems to be talking fairly specifically about the username component, or about some sort of component-wise operation. I can't seem to produce any such strings using the live URL viewer and any of the examples therein.

domenic (approved):

PR #735, opened 2023-01-16
Add value argument to URLSearchParams's has() and delete()

### LINKED ISSUE #335: URLSearchParams delete all vs delete one
The specification for URLSearchParams currently reads that delete should delete all pairs that have the supplied key. I think the ability to delete a specific key value pair should be added.

For example:
```js
let url = new URLSearchParams("key1=value1&key1=value1a&key2=value2");
url.delete("key1","value1");
url.toString(); // key1=value1a&key2=value2
```

A use case for this is faceted searching. Say the user has selected three keys to facet with. Example url: ```http://example.com/?q=searchTerms&facet.category=selection1&facet.category=selection2&facet.category=selection3```. Now the user wants to clear only selection2 from their search.

While it wouldn't be too difficult to use the current methods to obtain the desired output above (getAll the given key, delete all pairs with the given key, iterate over the getAll result adding back everything but the one pair), I think it goes against the goal of the URLSearchParams class as I understand it. It seems URLSearchParams tries to abstract away the string filtering and concatenations that were required for query params before. The workaround falls back into that extra manipulation approach.

I think that any of the following changes would solve this.

1) Allow a second optional argument to be passed to delete that would specify the value of which key-value to delete if there were multiple key-value pairs with the given key.

Example: See above.

2) Change the current definition of delete to require two arguments, a key and a value. The method would then delete the supplied key-value pair, if it exists. Introduce a new method, deleteAll, that behaves how delete does currently. This would follow the pattern of having both get and getAll methods.

```js
let url = new URLSearchParams("key1=value1&key1=value1a&key2=value2&key2=value2b");
url.delete("key1","value1");
url.toString(); // key1=value1a&key2=value2&key2=value2b
url.deleteAll("key2");
url.toString(); // key1=value1a
```

For both options, in cases where there are repeats of the same key-value pair, all instances would be deleted.
Example:
```js
let url = new URLSearchParams("key1=value1&key1=value1&key2=value2");
url.delete("key1","value1");
url.toString(); // key2=value2
```

While I am more of a fan of option 2 because it lines up closer to the get/getAll methods, it requires a backwards incompatible change. Therefore, it would probably be best to go the route of option 1. (Arguably, a third option could be to keep 1 argument delete the same, introduce deleteAll and two argument delete, but is probably extra and unnecessary.)

What are other people's thoughts?

annevk:
Tests: https://github.com/web-platform-tests/wpt/pull/39865.

Fixes #335.

   * Chromium
   * WebKit
   * https://github.com/web-platform-tests/wpt/pull/39865
   * Chromium: https://bugs.chromium.org/p/chromium/issues/detail?id=1442916
   * Gecko: https://bugzilla.mozilla.org/show_bug.cgi?id=1831587
   * WebKit: https://bugs.webkit.org/show_bug.cgi?id=256369
   * Deno: https://github.com/denoland/deno/issues/19000
   * Node.js: https://github.com/nodejs/node/issues/47883
- [x] [MDN issue](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) is filed: https://github.com/mdn/content/issues/26604

***
<a href="https://whatpr.org/url/735.html" title="Last updated on May 5, 2023, 5:21 PM UTC (81412b3)">Preview</a> | <a href="https://whatpr.org/url/735/ae3c28b...81412b3.html" title="Last updated on May 5, 2023, 5:21 PM UTC (81412b3)">Diff</a>

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -3700,10 +3700,10 @@ interface URLSearchParams {
   readonly attribute unsigned long size;
 
   undefined append(USVString name, USVString value);
-  undefined delete(USVString name);
+  undefined delete(USVString name, optional USVString value);
   USVString? get(USVString name);
   sequence&lt;USVString> getAll(USVString name);
-  boolean has(USVString name);
+  boolean has(USVString name, optional USVString value);
   undefined set(USVString name, USVString value);
 
   undefined sort();
@@ -3761,8 +3761,9 @@ console.log(url.searchParams.get('b')); // "~"</code></pre>
 the potential to change that object's <a for=url>path</a> in a manner equivalent to the {{URL}}
 object's {{URL/search}} and {{URL/hash}} setters.
 
+<div algorithm>
 <p>To <dfn for=URLSearchParams oldids=concept-urlsearchparams-new>initialize</dfn> a
-{{URLSearchParams}} object <var>query</var> with <var>init</var>, run these steps:
+{{URLSearchParams}} object <var>query</var> with <var>init</var>:
 
 <ol>
  <li>
@@ -3791,9 +3792,11 @@ object's {{URL/search}} and {{URL/hash}} setters.
    <a lt="urlencoded string parser">parsing</a> <var>init</var>.
   </ol>
 </ol>
+</div>
 
+<div algorithm>
 <p>To <dfn for=URLSearchParams id=concept-urlsearchparams-update>update</dfn> a {{URLSearchParams}}
-object <var>query</var>, run these steps:
+object <var>query</var>:
 
 <ol>
  <li><p>If <var>query</var>'s <a for=URLSearchParams>URL object</a> is null, then return.
@@ -3811,7 +3814,9 @@ object <var>query</var>, run these steps:
  <a>potentially strip trailing spaces from an opaque path</a> with <var>query</var>'s
  <a for=URLSearchParams>URL object</a>.
 </ol>
+</div>
 
+<div algorithm>
 <p>The
 <dfn constructor for=URLSearchParams lt="URLSearchParams(init)"><code>new URLSearchParams(<var>init</var>)</code></dfn>
 constructor steps are:</p>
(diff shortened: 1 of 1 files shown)


jasnell:
I'm +1 on this for both Node.js and workerd (Cloudflare workers).

zloirock:
It's better at first to check the web compatibility of this. Most likely, someone already used it with extra arguments that were ignored, like passing to `forEach`.
```js
array.forEach(params.delete, params);
```

jasnell:
Ah right, yeah, good point. Separate methods are unfortunate but far less likely to break things.

zloirock:
@jasnell I'm not sure that `(delete|has)Value` is a good name option - one value can be under multiple keys. It should be something like "pair".

jasnell:
How about `hasEntry` and `deleteEntry`?

zloirock:
As an option.

annevk:
I'm not sure I'm ready to accept we need renaming here. Note that `URLSearchParams` has to be the `this` as well. Also https://jakearchibald.com/2021/function-callback-risks/.

zloirock:
Web standard developers have no right to break the web and if you don't want to break it, you'll have to check it - this is a pretty popular pattern - and if it does, you'll have to rename these methods. The post you mentioned has absolutely nothing to do with this issue.

annevk:
I'm pretty sure most JavaScript builtins don't do this-checks so they are indeed more exposed. And yes, obviously we'll have to try it and revisit if it doesn't work.

annevk:
@valenting @ricea is there interest from Gecko and Chromium in this API addition? (Note that #335 is maybe the most popular request.)

ricea (commented):
lgtm

ricea:
Tentative support from Chromium (subject as usual to review by API owners, who will probably ask for evidence it doesn't break the web).

andreubotella:
I added a use counter to Chromium. Let's check back in a couple months.

andreubotella:
> [I added a use counter to Chromium.](https://chromestatus.com/metrics/feature/timeline/popularity/4478) Let's check back in a couple months.

The use counter has now been in stable Chromium for a month, and it seems like the the usage that would be broken by this change [is negligible](https://chromestatus.com/metrics/feature/timeline/popularity/4478).

domenic (approved):

foolip:
I've filed https://github.com/web-platform-tests/interop/issues/338 about the Interop 2023 impact of this.

PR #738, opened 2023-01-17
Editorial: point out API section is not suitable for all

### LINKED ISSUE #535: Clarify scope of API define in the spec or split specs
I believe that some confusion has resulted because of the fact that the spec includes, in addition to the definition and semantics of the URL itself, an IDL API which is implied as an API to all URLs, despite IDL only really being a spec for Web APIs. I have heard someone somewhere, though I forget where, express that because it is IDL it is non-normative when not integrating against the web technologies that use such APIs.

The web-centric design causes some difficulty (e.g. servo/rust-url#577) because specific API decisions are made with web APIs in mind. The issue I just linked runs into this problem where code is relying on being able to exchange a special scheme for a non-special one, which is not really supported (or completely unsupported?) in the API provided, but shouldn't be seen as an inherently invalid operation. URLs are a technology that stretches far beyond the browser-based web, and their general usage shouldn't, in my view, be limited by web-specific decisions. 

There's also an argument that IDL does not translate well to certain languages, such as those with value-based objects for which "set" is not a meaningful concept. There's additionally an issue of backwards-compatibility: many languages have strong backwards-compatibolity guarantees and cannot commit to a spec that implements no such guarantees.

Ideally, I think the spec should make clear that the IDL API is *only* for web, and language libraries should design their own APIs as they feel appropriate around the core concepts. As an alternative, it should go in the direction of insisting that all APIs should match its semantics exactly, and clearly specifying that this is true regardless of its use of IDL. Personally, if it went in this direction, I would be inclined to argue against its widespread adoption outside of the context of the Web and support further development of the canonical specification only through the IETF RFC process.

I thought I had brought this up before on an issue here, but I can't find it, so perhaps not. Apologies if it's a duplicate.

annevk:
This language is copied from Encoding, to which similar considerations applied.

Closes #535.

***
<a href="https://whatpr.org/url/738.html" title="Last updated on Jan 17, 2023, 8:50 AM UTC (18c1b4c)">Preview</a> | <a href="https://whatpr.org/url/738/2e4a19c...18c1b4c.html" title="Last updated on Jan 17, 2023, 8:50 AM UTC (18c1b4c)">Diff</a>

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -55,16 +55,16 @@ might increase in scope somewhat.
 
 <h2 id=infrastructure>Infrastructure</h2>
 
-<p>This specification depends on the Infra Standard. [[!INFRA]]
+<p>This specification depends on <cite>Infra</cite>. [[!INFRA]]
 
 <p>Some terms used in this specification are defined in the following standards and specifications:
 
 <ul class=brief>
- <li>Encoding Standard [[!ENCODING]]
- <li>File API [[!FILEAPI]]
- <li>HTML Standard [[!HTML]]
- <li>Unicode IDNA Compatibility Processing [[!UTS46]]
- <li>Web IDL [[!WEBIDL]]
+ <li><cite>Encoding</cite> [[!ENCODING]]
+ <li><cite>File API</cite> [[!FILEAPI]]
+ <li><cite>HTML</cite> [[!HTML]]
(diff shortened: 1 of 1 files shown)


annevk:
@alercah what do you think?

domenic (approved):

PR #739, opened 2023-01-17
Redo validation errors in the IPv4 parser

### LINKED ISSUE #706: Defered validation error reporting in IPv4 parser
After #619, the [IPv4 parser](https://url.spec.whatwg.org/#concept-ipv4-parser) now only returns an IPv4 address or failure. Previously, it returned it's input instead of failure in many cases. The only call to the IPv4 parser is now guarded by the [ends-in-a-number-checker](https://url.spec.whatwg.org/#ends-in-a-number-checker), which makes the following note outdated:

> This uses validationError to track [validation errors](https://url.spec.whatwg.org/#validation-error) to avoid reporting them before we are confident we want to parse input as an IPv4 address as the [host parser](https://url.spec.whatwg.org/#concept-host-parser) almost always invokes the [IPv4 parser](https://url.spec.whatwg.org/#concept-ipv4-parser).

The IPv4 parser essentially pauses validation error reporting in Step 1 and resumes reporting in Step 7. Step 3.1 sets `validationError` to `true`, but fails to report it if the parser returns failure before Step 7.

Step 3.1 should probably be changed to "validation error". This would leave the loop in Step 6 as the only place where `validationError` is set to `true`.

Is the IPv4 parser designed to be used without first running [ends-in-a-number-checker](https://url.spec.whatwg.org/#ends-in-a-number-checker)? Does the use of `validationError` still make sense?

Removing the last use of `validationError` and directly reporting a validation error in [ipv4-number-parser](https://url.spec.whatwg.org/#ipv4-number-parser) Step 4.1 and 5.1 would report a validation error for each part instead of a single validation error for the entire address. I don't know if that is desirable or not.

annevk:
ab0e820b0b559610b30c731b7f2c1a8094181680 didn't consider the impact on the validation error infrastructure.

This also makes some minor additional changes:
* Modernizes the "Host parsing" section.
* Clarifies that the IPv4 parser cannot be invoked directly.
* Clarifies that the IPv6 parser can, but you should still not do it.

Fixes #706.

***
<a href="https://whatpr.org/url/739.html" title="Last updated on Jan 17, 2023, 12:52 PM UTC (8dfebab)">Preview</a> | <a href="https://whatpr.org/url/739/42ce81b...8dfebab.html" title="Last updated on Jan 17, 2023, 12:52 PM UTC (8dfebab)">Diff</a>

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -413,7 +413,8 @@ point <a for=/>URLs</a> from <var>A</var> can come from untrusted sources.
 
 <div class=example id=example-host-parsing>
  <p>A <a lt="host parser">parse</a>-<a lt="host serializer">serialize</a> roundtrip gives the
- following results, depending on the <var>isNotSpecial</var> argument to the <a>host parser</a>:
+ following results, depending on the <var ignore>isNotSpecial</var> argument to the
+ <a>host parser</a>:
 
  <table>
   <tr>
--- a/url.bs
+++ b/url.bs
@@ -731,9 +731,10 @@ to be distinguished.
 
 <h3 id=host-parsing>Host parsing</h3>
 
+<div algorithm>
 <p>The <dfn export id=concept-host-parser lt="host parser|host parsing">host parser</dfn> takes a
 <a>scalar value string</a> <var>input</var> with an optional boolean <var>isNotSpecial</var>
-(default false), and then runs these steps:
+(default false), and then runs these steps. They return failure or a <a for=/>host</a>.
 
 <ol>
  <li>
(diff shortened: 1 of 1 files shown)


TRowbotham:
This looks sensible to me.

url.bs
```diff
@@ -807,26 +811,24 @@ runs these steps:
...
 
-<ol>
- <li>
-  <p>Let <var>validationError</var> be false.
-
-  <p class=note>This uses <var>validationError</var> to track <a>validation errors</a> to avoid
-  reporting them before we are confident we want to parse <var>input</var> as an IPv4 address as the
-  <a>host parser</a> almost always invokes the <a>IPv4 parser</a>.
+<p class=note>The <a for=/>IPv4 parser</a> is not to be invoked directly. Instead check that the
+return value of the <a for=/>host parser</a> is an <a for=/>IPv4 address</a>.
```
domenic, line 821: I don't understand why IPv4 and IPv6 would be different...

domenic (approved):
LGTM. These action-at-a-distance validation errors are going to be kind of a pain in the implementation (there's no object passed to the host/IPv4/IPv6 parsers to thread them back to), but I think that's best left as an implementation problem.

PR #752, opened 2023-02-08
meta: add Ada to pull request template

anonrig:


PULL_REQUEST_TEMPLATE.md
```diff
@@ -13,6 +13,7 @@ Thank you for contributing to the URL Standard! Please describe the change you a
    * WebKit: …
    * Deno: …
    * Node.js: …
+   * Ada: …
```
TimothyGu, line 16: You can probably remove Node.js now that Node.js uses Ada.

annevk:
Let's resolve the Node.js vs Ada question in https://github.com/whatwg/spec-factory/pull/47 and once that merges we can generate a new PR using `factory.py` that might also pull in other changes that have happened.

PR #785, opened 2023-09-07
Escape trailing spaces in opaque paths

### LINKED ISSUE #784: URL serialiser does not produce idempotent strings if an opaque path has a trailing space and "exclude fragment" is true
- A URL's opaque path can contain trailing spaces if there is a query or fragment after it
- This presents a problem, because if the URL somehow loses its query or fragment, the URL's serialisation will have trailing spaces. Since the parser trims spaces, re-parsing such a URL would not return an equivalent value to one that was serialised.

- In #651 we modified the query and fragment setters to safeguard against this - if you use the API to remove the query/fragment, trailing spaces in the opaque path are trimmed.
- However, the URL serialiser also has an "exclude fragment" parameter (default false). If this parameter were to be set true, it would produce the same issue as #651

Fixing this will be tricky. I think there are only a couple of options:

1. We just document the status quo; the API/parser/serialiser are idempotent except when "exclude fragment" is set to true, in which case they might corrupt the `path`. That doesn't sound good. It's quite common to exclude the fragment.

2. We trim opaque paths when serialising. That's also not good, because it means the API tells me one thing, but somebody (such as a browser) could serialise the URL without fragment, reparse the result, and the same API property would have a different value.

3. We always trim trailing spaces from opaque paths. Technically a breaking change, but overall it's better at ensuring everything stays consistent.

karwa:
Resolves #784 

I'll do the tests/bugs/etc if we're okay with this approach (escaping trailing spaces rather than trimming them). PR is for discussion at this point.

---

   * …
   * …
   * …
   * Chromium: …
   * Gecko: …
   * WebKit: …
   * Deno: …
   * Node.js: …
- [ ] [MDN issue](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) is filed: …

***
<a href="https://whatpr.org/url/785.html" title="Last updated on Sep 7, 2023, 12:49 PM UTC (758cac2)">Preview</a> | <a href="https://whatpr.org/url/785/d2ca75f...758cac2.html" title="Last updated on Sep 7, 2023, 12:49 PM UTC (758cac2)">Diff</a>

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -2846,6 +2846,14 @@ and then runs these steps:
    <dt><dfn export for="basic URL parser" id=cannot-be-a-base-url-path-state>opaque path state</dfn>
    <dd>
     <ol>
+     <li><p>If <a>c</a> is U+003F (?), U+0023 (#), or the <a>EOF code point</a>:
+       <ol>
+         <li><p>Let <var>n</var> be the number of U+0020 SPACE code points at the end of <var>url</var>'s
+         <a for=url>path</a>.
+         <li><p>Replace the last <var>n</var> code points in <var>url</var>'s <a for=url>path</a>
+         with <var>n</var> repetitions of the string "%20".
+       </ol>
+
      <li><p>If <a>c</a> is U+003F (?), then set <var>url</var>'s <a for=url>query</a> to the empty
      string and <var>state</var> to <a>query state</a>.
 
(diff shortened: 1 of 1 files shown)


annevk:
I suggest we close this in favor of #844. Thoughts?

PR #793, opened 2023-09-26
Fix validation error when a URL has credentials but no host

karwa:
Since this branch requires that `atSignSeen` is true, we must have already emitted an `invalid-credentials` validation error and this emission must be a duplicate.

What this check is actually doing is confirming that the host is not empty before it begins parsing it, so `host-missing` is a more appropriate error to emit.

   * …
   * …
   * …
   * Chromium: …
   * Gecko: …
   * WebKit: …
   * Deno: …
   * Node.js: …
- [ ] [MDN issue](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) is filed: …

***
<a href="https://whatpr.org/url/793.html" title="Last updated on Sep 26, 2023, 8:32 AM UTC (6e3c878)">Preview</a> | <a href="https://whatpr.org/url/793/45a483a...6e3c878.html" title="Last updated on Sep 26, 2023, 8:32 AM UTC (6e3c878)">Diff</a>

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -2448,7 +2448,7 @@ and then runs these steps:
 
       <ol>
        <li><p>If <var>atSignSeen</var> is true and <var>buffer</var> is the empty string,
-       <a>invalid-credentials</a> <a>validation error</a>, return failure.
+       <a>host-missing</a> <a>validation error</a>, return failure.
        <!-- No URLs with userinfo, but without host. For special URLs it would also not be
             idempotent:
             https://@/example.org/ -> https:///example.org/ -> https://example.org/ -->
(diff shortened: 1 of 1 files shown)


karwa:
Whoops, sorry - made a mistake. Need to check this again.

karwa:
Sorry just wanted to verify it once again. I think this is correct.

annevk (commented):
This is correct, but the validation error table will need updates as well. As invalid-credentials would no longer be a fatal error and the second example there would be better moved to host-missing.

karwa:
@annevk Ah yes, okay - updated. I moved the example and replaced it with one that demonstrates credentials in a non-special scheme.

annevk (approved):

PR #804, opened 2023-11-06
[editorial] Rephrase encoding note to make the implications clearer.

tabatkins:
I didn't immediately realize the implications of this note and was *very* confused about how to handle the URL properly when doing some CSS work. I think this wording would have led me to the correct solution faster, so hopefully it'll help others.

(Note: I used a 100 character wrap, as that's what the rest of the spec appears to use. I also used Bikeshed'd `[=foo=]` syntax, but the rest of the spec seems to prefer explicit `<a>`s. I can switch if you want.)

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -2038,8 +2038,8 @@ and <a>code points</a> in the range U+00A0 to U+10FFFD, inclusive, excluding <a>
 <!-- IRI also excludes the ranges U+E000 to U+F8FF, U+FFF0 to U+FFFD, and U+E0000 to U+E09FF, all
      inclusive. We don't to align with HTML. -->
 
-<p class=note>For historical reasons, rather than storing codepoints and [=percent-encoding=] to
-ASCII for serialization, URLs instead store their value as ASCII internally, eagerly converting
+<p class=note>For historical reasons, rather than storing codepoints and [=byte/percent-encoding=]
+to ASCII for serialization, URLs instead store their value as ASCII internally, eagerly converting
 code points greater than U+007F DELETE to [=percent-encoded bytes=] during [=URL parser|parsing=].
 
 <p class=note>In HTML, when the document encoding is a legacy encoding, code points in the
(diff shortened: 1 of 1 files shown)


url.bs
```diff
@@ -2038,8 +2038,9 @@ and <a>code points</a> in the range U+00A0 to U+10FFFD, inclusive, excluding <a>
 <!-- IRI also excludes the ranges U+E000 to U+F8FF, U+FFF0 to U+FFFD, and U+E0000 to U+E09FF, all
      inclusive. We don't to align with HTML. -->
 
-<p class=note>Code points greater than U+007F DELETE will be converted to
-<a lt="percent-encoded byte">percent-encoded bytes</a> by the <a>URL parser</a>.
+<p class=note>For historical reasons, rather than storing codepoints and [=byte/percent-encoding=]
+to ASCII for serialization, URLs instead store their value as ASCII internally, eagerly converting
+code points greater than U+007F DELETE to [=percent-encoded bytes=] during [=URL parser|parsing=].
```
annevk, line 2043: Apologies for not responding to this more quickly, but I think I never ended up merging it because I'm not sure this is correct. I suspect one could convert at serialization time instead. It's just not how the specification is written.

tabatkins: Sure, the spec *could* be written another way (potentially), but it's currently *not* written that way, and the specifics of how the data is encoded/represented at this point in the spec are important, so I know that the URL structure only includes ASCII code points. If we changed to a "convert at serialization" model, that would *also* be important to note, so it was clear that the URL structure includes non-ASCII code points.

As I said, the nature of this note actively confused me - the spec talks about "URL code points" including non-ASCII codepoints, but URLs themselves do not contain these code points, and that wasn't clear to me from how the note was written.

annevk: Hmm. 1) It's not for "historical reasons". 2) This section is really about writing URLs, it isn't really about their internal representation at all. That's section 4.1 and that already makes it clear most components are ASCII strings.

tabatkins: The "for historical reasons" was me assuming and editorializing. (It seemed like a weird thing to do! It's not usually good practice to encode into the byte format immediately; usually you hold it in the good data model and only encode at the edges, when you have to hit the wire.) If that's not true, and it really is just a quirk of the model, I can rephrase that bit.

And this section is about writing URLs, sure, but there was *already* a note about how those codepoints you write will be encoded. I was just rewriting the note for (imo) better clarity. If there's a better place to make this note, I can move it there, but this section does seem relatively germane to what the note is saying (since URLs can "contain" high codepoints, but the actual internal representation is ASCII-only).

Issue #321, opened 2017-06-03
"IPv6 serializer" does not define an algorithm for longest sequence of IPv6 pieces that are 0


GPHemsley:
Similar to #318, the "IPv6 serializer" requires determining the longest sequence of 0s for IPv6 compression without specifying an algorithm for doing that:

> Let compress be an index to the first IPv6 piece in the first longest sequences of address’s IPv6 pieces that are 0.
> Example: In 0:f:0:0:f:f:0:0 it would point to the second 0.
> If there is no sequence of address’s IPv6 pieces that are 0 that is longer than 1, then set compress to null.

It would be better for implementors if the procedure for doing so were made explicit, as this is another seemingly non-trivial task.

annevk:
Per https://github.com/jsdom/whatwg-url/pull/150 it does not seem that tricky, but happy to write that out in prose (after confirming my modifications are indeed correct).

PR #838, opened 2024-11-18
Editorial: adopt Infra number definitions (and redo IPv6 pieces)

annevk:
***
<a href="https://whatpr.org/url/838.html" title="Last updated on Nov 25, 2024, 4:43 PM UTC (280ff43)">Preview</a> | <a href="https://whatpr.org/url/838/302c5af...280ff43.html" title="Last updated on Nov 25, 2024, 4:43 PM UTC (280ff43)">Diff</a>

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -756,14 +756,14 @@ not equivalent and typically treated as distinct.
 
 <p>An <dfn export>IP address</dfn> is an <a>IPv4 address</a> or an <a>IPv6 address</a>.
 
-<p>An <dfn export id=concept-ipv4>IPv4 address</dfn> is a 32-bit unsigned integer that identifies a
-network address.
+<p>An <dfn export id=concept-ipv4>IPv4 address</dfn> is a <a for=/>32-bit unsigned integer</a> that
+identifies a network address.
 [[RFC791]]
 
-<p>An <dfn export id=concept-ipv6>IPv6 address</dfn> is a 128-bit unsigned integer that identifies a
-network address. For the purposes of this standard it is represented as a <a for=/>list</a> of eight
-16-bit unsigned integers, also known as
-<dfn export lt="IPv6 piece" id=concept-ipv6-piece>IPv6 pieces</dfn>.
+<p>An <dfn export id=concept-ipv6>IPv6 address</dfn> is a <a for=/>128-bit unsigned integer</a> that
+identifies a network address. For the purposes of this standard it is defined as of a
+<a for=/>list</a> of 8 <a for=/>16-bit unsigned integers</a>, also known as its
+<dfn export for="IPv6 address" id=concept-ipv6-piece>pieces</dfn>.
 [[RFC4291]]
 
 <p class="note">Support for <code>&lt;zone_id></code> is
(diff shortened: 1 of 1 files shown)


url.bs
```diff
@@ -1444,16 +1445,16 @@ actually doing that with the editors of this document first.
...
+  <var>compress</var> to the index of the first item of the first of the longest sequences of
+  <var>address</var>'s <a for="IPv6 address">pieces</a> that are 0.
 
   <p class=example id=example-e2b3492e>In <code>0:f:0:0:f:f:0:0</code> it would point to
   the second 0.
 
- <li><p>If there is no sequence of <var>address</var>'s <a>IPv6 pieces</a> that are 0 that is
- longer than 1, then set <var>compress</var> to null.
-
  <li><p>Let <var>ignore0</var> be false.
```
annevk, line 1458: This is probably the biggest change here.

domenic: I think this changed from excluding 1-length sequences to including them.

FYI here is whatwg-url's implementation. It might be worth writing the whole thing out as it's pretty complex. https://github.com/jsdom/whatwg-url/blob/0fb65e3c3671c2e790ff38aed9332fc7fb2cd12b/lib/url-state-machine.js#L387

annevk: I was hoping that multiple and sequentially would do that work here. I could rewrite this though. #321 also requested that. Hopefully next week.

domenic: Hmm yeah, I think I skipped over the word "multiple". Sounds good to me, pending a rewrite :).

url.bs
```diff
@@ -756,14 +756,14 @@ not equivalent and typically treated as distinct.
...
 [[RFC791]]
 
-<p>An <dfn export id=concept-ipv6>IPv6 address</dfn> is a 128-bit unsigned integer that identifies a
-network address. For the purposes of this standard it is represented as a <a for=/>list</a> of eight
-16-bit unsigned integers, also known as
-<dfn export lt="IPv6 piece" id=concept-ipv6-piece>IPv6 pieces</dfn>.
+<p>An <dfn export id=concept-ipv6>IPv6 address</dfn> is a <a for=/>128-bit unsigned integer</a> that
+identifies a network address. This integer is composed of a <a for=/>list</a> of 8
+<a for=/>16-bit unsigned integers</a>, also known as an <a for=/>IPv6 address</a>'s
+<dfn export for="IPv6 address" id=concept-ipv6-piece>pieces</dfn>.
```
rmisev, line 766: This definition is obscure: it is not clear how to convert a 128-bit integer to a list of pieces and vice versa. Also, all algorithms use an IPv6 address as a list of pieces, not as a 128-bit integer. By the way, RFC4291 defines an IPv6 address slightly differently: "IPv6 addresses are 128-bit identifiers...". So I think it is better to use the RFC4291 definition - replace "unsigned integer" with "identifier":
> An IPv6 address is a 128-bit **identifier** that identifies a network address. This **identifier** is composed of ...

annevk: All algorithms inside URL, yes, but outside? This we would have to change separately though as changing it from integer to identifier would not be an editorial change.

rmisev (approved):

url.bs
domenic: I do agree that this is a bit confusing. I can see how it's "conceptually" a u128, or "corresponds to" a u128, but in practice every part of the spec treats it as a list of 8 u16s.

domenic (approved):
LGTM, matches whatwg-url with some variable names being clearer in this version.

PR #840, opened 2024-11-29
Remove redundant domain to Unicode call from valid domain

### LINKED ISSUE #817: Explain why valid domain needs to run ToUnicode
### What is the issue with the URL Standard?

https://url.spec.whatwg.org/#valid-domain could use an informative note that states the implications of the two-step (both ToASCII and ToUnicode) check. Given the how both use UTS 46 "Processing" and "ToASCII" does more stuff after "Processing", it would be helpful to call out what the second run of "Processing" (as part of "ToUnicode") catches still.

annevk:
I'm not entirely sure why this redundant check existed. Either because there was a difference back when this definition was introduced in 3bec3b89c4deb10842ba6c464c700df47c268f17 or (more likely) I wasn't sure if there was a difference.

Fixes #817.

***
<a href="https://whatpr.org/url/840.html" title="Last updated on Nov 29, 2024, 8:21 AM UTC (d0f8256)">Preview</a> | <a href="https://whatpr.org/url/840/7ff8de0...d0f8256.html" title="Last updated on Nov 29, 2024, 8:21 AM UTC (d0f8256)">Diff</a>

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -952,20 +952,13 @@ concepts.
 <a>valid IPv4-address string</a>, or: U+005B ([), followed by a
 <a>valid IPv6-address string</a>, followed by U+005D (]).
 
-<p>A <var>domain</var> is a <dfn>valid domain</dfn> if these steps return success:
+<p>A <a for=/>string</a> <var>input</var> is a <dfn>valid domain</dfn> if these steps return true:
 
 <ol>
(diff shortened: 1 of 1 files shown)


hsivonen (approved):
LGTM. Thanks!

PR #844, opened 2024-12-02
Ensure opaque paths always roundtrip

### LINKED ISSUE #784: URL serialiser does not produce idempotent strings if an opaque path has a trailing space and "exclude fragment" is true

annevk:
In fdaa0e5a3790693a82f578d7373f216d8fef9ac8 we tackled a problem whereby removing the fragment or query from a URL with an opaque path through the API would not make the URL roundtrip due to the opaque path being able to end in non-percent-encoded spaces.

However, this failed to address other ways of serializing the URL. As such this is a new approach whereby opaque paths simply cannot end with non-percent-encoded spaces. Enforcing this in the URL parser allows us to completely revert the aforementioned commit, greatly simplifying the API implementation.

Fixes #784.

---

   * Chromium (at least during Interop 2024)
   * Gecko
   * WebKit
   * https://github.com/web-platform-tests/wpt/pull/51129
   * Chromium: https://issues.chromium.org/issues/400919461
   * Gecko: https://bugzilla.mozilla.org/show_bug.cgi?id=1951963
   * WebKit: https://bugs.webkit.org/show_bug.cgi?id=289160
   * Deno: https://github.com/denoland/deno/issues/28395
   * Node.js: https://github.com/nodejs/node/issues/57329
- [x] [MDN issue](https://github.com/whatwg/meta/blob/main/MAINTAINERS.md#handling-pull-requests) is filed: Since https://github.com/mdn/content/issues/23320 is not fixed yet we can perhaps reuse that.
- [x] The top of this comment includes a [clear commit message](https://github.com/whatwg/meta/blob/main/COMMITTING.md) to use. 

***
<a href="https://whatpr.org/url/844.html" title="Last updated on Mar 13, 2025, 11:57 AM UTC (5600998)">Preview</a> | <a href="https://whatpr.org/url/844/076afff...5600998.html" title="Last updated on Mar 13, 2025, 11:57 AM UTC (5600998)">Diff</a>

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -2913,24 +2913,31 @@ and then runs these steps:
    <dt><dfn export for="basic URL parser" id=cannot-be-a-base-url-path-state>opaque path state</dfn>
    <dd>
     <ol>
-     <li><p>If <a>c</a> is U+003F (?), then set <var>url</var>'s <a for=url>query</a> to the empty
-     string and <var>state</var> to <a>query state</a>.
+     <li><p>If <a>c</a> is U+003F (?), then set <var>buffer</var> to the empty string,
+     <var>url</var>'s <a for=url>query</a> to the empty string, and <var>state</var> to
+     <a>query state</a>.
 
-     <li><p>Otherwise, if <a>c</a> is U+0023 (#), then set <var>url</var>'s <a for=url>fragment</a>
-     to the empty string and <var>state</var> to <a>fragment state</a>.
+     <li><p>Otherwise, if <a>c</a> is U+0023 (#), then set <var>buffer</var> to the empty string,
+     <var>url</var>'s <a for=url>fragment</a> to the empty string, and <var>state</var> to
+     <a>fragment state</a>.
+
+     <li><p>Otherwise, if <a>c</a> is U+0020 SPACE, then append <a>c</a> to <var>buffer</var>.
 
      <li>
-      <p>Otherwise:
+      <p>Otherwise, if <a>c</a> is not the <a>EOF code point</a>:
 
       <ol>
-       <li><p>If <a>c</a> is not the <a>EOF code point</a>, not a <a>URL code point</a>, and not
-       U+0025 (%), <a>invalid-URL-unit</a> <a>validation error</a>.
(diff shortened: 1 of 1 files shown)


annevk:
@karwa @hayatoito @valenting what do you think? We should be able to make test changes again so would be great to hear what the interest in this change is.

valenting:
I'm supportive of this change. It's a nice simplification, and slightly better than having to strip spaces from the path when clearing the search or hash. I plan to be making these changes in Firefox once WPT tests are available.

karwa:
(thanks for taking this on, and apologies for the long delay!)

+1. This is a lot more predictable for users.

I hope it is web-compatible. I understand that there are use-cases which unfortunately require that unescaped spaces are preserved (e.g. `javascript:` URLs), but I can't imagine that _trailing_ spaces are important to any of them.

valenting (approved):

rmisev (approved):

domenic:
I guess this also obsoletes #785.

PR #845, opened 2024-12-02
Editorial: only care about the empty string when beStrict is false

annevk:
VerifyDnsLength will take care of it otherwise. (See #497 for additional context.)

***
<a href="https://whatpr.org/url/845.html" title="Last updated on Dec 2, 2024, 4:08 PM UTC (d4110e0)">Preview</a> | <a href="https://whatpr.org/url/845/c3d173f...d4110e0.html" title="Last updated on Dec 2, 2024, 4:08 PM UTC (d4110e0)">Diff</a>

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -934,8 +934,12 @@ concepts.
     <a href="https://github.com/whatwg/url/issues/397">issue #397</a>.
   </ol>
 
- <li><p><a for=/>Assert</a>: <var>result</var> is not the empty string and does not contain a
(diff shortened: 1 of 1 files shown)


hsivonen (approved):
Step 4 could use a note to mention that if _beStrict_ is true, the assertion holds for UTS 46 reasons.

r+ with or without the note, though.

rmisev (approved):

Issue #772, opened 2023-05-26
invalid-reverse-solidus error definition has character mixed up.


JamesNJep:
Under the host passing section https://url.spec.whatwg.org/#host-parsing 

What I think it is trying to say is that the URL scheme uses the solidus character U+002F (/) and not the reverse solidus character U+005C (\), but the characters seem to be in the wrong order in the sentence above the example.

annevk:
Are you talking about https://url.spec.whatwg.org/#invalid-reverse-solidus?

The characters after the scheme are correct, but those are not the ones in error. The ones later on are.

JamesNJep:
Yes was talking about [https://url.spec.whatwg.org/#invalid-reverse-solidus](https://url.spec.whatwg.org/#invalid-reverse-solidus)  thank you for looking at this.

The part I had trouble understanding is the description sentence above the example.
The description text reads 

> "The URL has a special scheme and it uses U+005C (\) instead of U+002F (/)."

The way I have read and interpreted this when first reading this is that the sentence is making a statement, about how the URL should be written to avoid this error, but I think the sentence is actually trying to make a positive statement to describe the error condition.

I've re-read this table several times and I think the other items in the table are easier to understand because they use words or prases like, "exceeds", "improper", "too many", "too few", "too big", "invalid" that make it clear that the statement is describing an error state. 

I think if the sentence was written more like  the desciption for [https://url.spec.whatwg.org/#example-host-missing](https://url.spec.whatwg.org/#example-host-missing)  and used "but" instead of "and" that would make it easier to understood the statment was describing the error condition if someone is jumping directly to that description.
For example if it was written like this 

> "The URL has a special scheme, but has used U+005C (\) in the place of U+002F (/)."

annevk:
I'm not sure how to apply that consistently. Many of these are statements describing the error state in exactly that fashion, e.g.,

> An [IPv4 address](https://url.spec.whatwg.org/#concept-ipv4) ends with a U+002E (.).

Although, maybe we should remove "instead of U+002F (/)" or put that between parenthesis? Nothing else has advice like that so maybe removing it would be best and make it clearer overall.

What do you think?

annevk:
@TRowbotham any thoughts from you on this feedback?

TRowbotham:
[...]

PR #861, opened 2025-03-06
Meta: correct dark mode .yesno styling

annevk:
This did not work correctly as pointed out here:
https://github.com/whatwg/url/issues/772#issuecomment-2702776864

---

@JamesNJep this looks pretty reasonable to me, but let me know if you prefer a different color.

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -18,7 +18,7 @@ spec: UTS46; urlPrefix: https://www.unicode.org/reports/tr46/
 </pre>
 
 <style>
-.yesno .yes { background: papayawhip; }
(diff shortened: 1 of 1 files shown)


annevk:
CI is blocked on https://github.com/tc39/ecma426/issues/174. Don't think I should address that here.

domenic (approved):
We should consider upstreaming this from HTML + URL into standard.css: https://github.com/whatwg/html/blob/aa586d0ba966ae97b824f636241dfd0cbfb83ce1/styles.css#L29-L30

annevk:
Good idea: https://github.com/whatwg/whatwg.org/pull/457.

domenic (approved):

PR #872, opened 2025-05-12
Editorial: Use code unit less than algorithm for URLSearchParams sort

shannonbooth:
I noticed this algorithm in the infra spec, and I believe this is what all engines already do for the sort.

Hopefully not mistaken!

***
<a href="https://whatpr.org/url/872.html" title="Last updated on May 12, 2025, 7:54 AM UTC (216797e)">Preview</a> | <a href="https://whatpr.org/url/872/cc8b776...216797e.html" title="Last updated on May 12, 2025, 7:54 AM UTC (216797e)">Diff</a>

### DIFF
--- a/url.bs
+++ b/url.bs
@@ -4061,7 +4061,8 @@ sorted.sort()</code></pre>
 <p>The <dfn method for=URLSearchParams><code>sort()</code></dfn> method steps are:
 
 <ol>
- <li><p><a for=list>Sort in ascending order</a> <a>this</a>'s <a for=URLSearchParams>list</a>,
+ <li><p>Set <a>this</a>'s <a for=URLSearchParams>list</a> to the result of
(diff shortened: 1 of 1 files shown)


annevk (commented):
Thanks for noticing this. We want to reference https://infra.spec.whatwg.org/#list-sort-in-ascending-order here as well and update the wording to follow that (similar to the example).

shannonbooth:
> Thanks for noticing this. We want to reference [infra.spec.whatwg.org#list-sort-in-ascending-order](https://infra.spec.whatwg.org/#list-sort-in-ascending-order) here as well and update the wording to follow that (similar to the example).

Thanks! Made that edit, though I'm not 100% sure I got the tuple part of this completely correct in proper spec wording

annevk (commented):
The wording looks correct, but we need to start with something like:

> Set this's list to the result of sorting in ...

as we are now allocating a new list instead of mutating in place.

annevk (approved):

